PR #4, opened 2018-12-02
Gracefully restart server instead of killing process

olafurpg:
Metals performs important cleanups during the `shutdown` request
so it's best to avoid killing the process.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -109,17 +109,21 @@ function startServer(context: ExtensionContext, javaHome: string) {
   );
 
   context.subscriptions.push(
-    commands.registerCommand("metals.restartServer", async () => {
-      await client.sendRequest(ShutdownRequest.type);
-      client.sendNotification(ExitNotification.type);
-      const showLogsAction = "Show server logs";
-      const selectedAction = await window.showInformationMessage(
-        "Metals Language Server killed, it should restart in a few seconds",
-        showLogsAction
-      );
-      if (selectedAction === showLogsAction) {
-        client.outputChannel.show(true);
-      }
+    commands.registerCommand("metals.restartServer", () => {
+      // First try to gracefully shutdown the server with LSP `shutdown` and `exit`.
+      // If Metals doesn't respond within 4 seconds we kill the process.
+      const killTimeout = setTimeout(() => {
+        window.showWarningMessage(
+          "Metals is unresponsive, killing the process and starting a new server."
+        );
+        const serverPid = client["_serverProcess"].pid;
+        exec(`kill ${serverPid}`);
+      }, 4000);
+      client.sendRequest(ShutdownRequest.type).then(() => {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -108,14 +110,13 @@ function startServer(context: ExtensionContext, javaHome: string) {
 
   context.subscriptions.push(
     commands.registerCommand("metals.restartServer", async () => {
-      const serverPid = client["_serverProcess"].pid;
-      await exec(`kill ${serverPid}`);
+      await client.sendRequest(ShutdownRequest.type);
+      client.sendNotification(ExitNotification.type);
```
gabro: I think we should always fallback to killing the pid after a timeout, to account for completely unresponsive servers.

Also, have you tested this? This previously relied on `language-client` to restore the server when killed, I'm not sure it will restart it if terminated gracefully.

src/extension.ts
```diff
@@ -108,14 +110,13 @@ function startServer(context: ExtensionContext, javaHome: string) {
...
       const showLogsAction = "Show server logs";
       const selectedAction = await window.showInformationMessage(
         "Metals Language Server killed, it should restart in a few seconds",
```
gabro: While we're at this, I suggest we use something slightly less violent than "killed" 🗡 
How about "Terminated"?

olafurpg: Good point, I kept "killed" when the server is truly `kill`ed. For the happy path the message is "Metals is restarting"

src/extension.ts
olafurpg: Good catch. It now tries to gracefully shutdown but if there's no response in 4s then we kill the process. I manually tested the `kill` path with a local version of Metals that sleeps during `shutdown`. Would be good to get a second eye on the timeout logic though, easy to mess this up.

src/extension.ts
```diff
@@ -107,18 +109,21 @@ function startServer(context: ExtensionContext, javaHome: string) {
...
+          "Metals is unresponsive, killing the process and starting a new server."
+        );
+        const serverPid = client["_serverProcess"].pid;
+        exec(`kill ${serverPid}`);
+      }, 4000);
+      client.sendRequest(ShutdownRequest.type).then(() => {
+        clearTimeout(killTimeout);
+        client.sendNotification(ExitNotification.type);
+        window.showInformationMessage("Metals is restarting");
+      });
```
gabro: This look ok, I'll give it a try later.
Alternatively you can model this as follows:
- promise A, that fails after a timeout
- promise B, that tries to terminate gracefully
- Promise.race(a, b).catch(killPid)

But only if you think it's clearer

olafurpg: Nice, that reads more cleanly. Merged your branch

gabro:
Merging as I've manually tested both scenarios (using ad-hoc timeouts to simulate server unresponsiveness)

PR #8, opened 2018-12-05
Show progress while downloading Metals.

olafurpg:
Previously, the plugin didn't show any activity while downloading
Metals which made it seem like Metals wasn't working the first
time users run it. Now, we show a progress bar and open the output
channel while downloading Metals for the first time.


olafurpg:
I also validated that the progress bar does not appear when all artifacts are cached locally, this only happens when there is some output to the console.

This PR also updates the coursier launcher so that `[2k` is no longer emitted in the logs thanks to https://github.com/coursier/coursier/pull/966

src/extension.ts
```diff
@@ -30,63 +32,95 @@ import {
...
+    classpath => {
+      launchMetals(
+        outputChannel,
+        context,
+        javaPath,
+        classpath,
+        serverProperties
+      );
+    },
+    _ => {
```
gabro: `_` still binds a variable, you can simply use `() => {`

For reference, TypeScript suppresses the warning for unused variables if they start with `_`, to allow for uses cases such as `(_dontcare, a) => { /* use a */ }`, so that's why you don't get a diagnostic on this one.

src/extension.ts
```diff
@@ -30,63 +32,95 @@ import {
...
+      let msg = `Failed to download Metals, make sure you have an internet connection and that the Metals version '${serverVersion}' is correct`;
```
gabro: minor, but I prefer `const` as a default

src/extension.ts
```diff
@@ -275,6 +310,32 @@ function startServer(context: ExtensionContext, javaHome: string) {
...
+  output: OutputChannel,
+  download: ChildProcessPromise
+): Promise<string> {
+  const progress = new LazyProgress();
+  let stdout = "";
+  download.stdout.on("data", (out: Buffer) => {
+    stdout += out.toString().trim();
+  });
+  download.stderr.on("data", (err: Buffer) => {
+    let msg = err.toString().trim();
```
gabro: ditto, `const`

src/lazy-progress.ts
```diff
@@ -0,0 +1,35 @@
...
+            function complete() {
+              p.report({ increment: 100 });
+              setTimeout(() => {
+                resolve();
+              }, 2000);
+            }
+            download.then(() => {
+              output.hide();
+              complete();
+            }, complete);
```
gabro: Minor, but here you're using `.then` and in other places you're using `await`.
I would prefer to pick a style and be consistent.

My personal preference is to avoid `async/await` as they compose poorly, but I care more about consistency.

gabro: In this specific example, I usually go with

```typescript
download.then(output.hide).then(complete, complete)
```

which makes more obvious the "finally-like" intention.

By the way https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally should be supported by modern browser engines, so you can also try that. Not sure it works on Electron, but it seems like it should.

```typescript
download.then(output.hide).finally(complete)
```

src/extension.ts
olafurpg: 👍 Fixed

src/extension.ts
olafurpg: Done

src/extension.ts
olafurpg: Done

src/lazy-progress.ts
olafurpg: I refactored to `download.then(complete, complete)` but `.then(output.hide)` did not work. New version is still nicer.

Also refactored `await` to use `then` instead, my head needs more time to understand async/await

PR #9, opened 2018-12-05
Don't start server if Dotty is enabled in the workspace

olafurpg:
This makes us play nicely with the Dotty extension, allowing the to
co-exist.


src/extension.ts
```diff
@@ -48,8 +49,16 @@ export async function activate(context: ExtensionContext) {
 
 function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   const outputChannel = window.createOutputChannel("Metals");
-  outputChannel.appendLine(`Java home: ${javaHome}`);
 
+  if (fs.existsSync(dottyIdeArtifact())) {
+    outputChannel.appendLine(
+      `Metals will not start since Dotty is enabled for this workspace. ` +
+        `To enable Metals, remove the file ${dottyIdeArtifact()} and run 'Reload window'`
```
gabro: super tiny pet-peeve, but I wouldn't prefer binding `dottyIdeArtifact()` and avoid calling it twice. It's irrelevant now, but it's "safer" in case the detection gets more complicated.

olafurpg: I thought the same but figured we can refactor this logic if it gets trickier.

PR #10, opened 2018-12-05
Doctor reload

olafurpg:


src/extension.ts
```diff
@@ -212,7 +212,10 @@ function launchMetals(
 
     // Handle the metals/executeClientCommand extension notification.
     client.onNotification(ExecuteClientCommand.type, params => {
-      if (params.command == "metals-doctor-run") {
+      if (
+        params.command === "metals-doctor-run" ||
+        (doctor !== undefined && params.command === "metals-doctor-reload")
```
gabro: The "golden standard" for guarding against `undefined` is `typeof doctor !== "undefined"` (it's really a pedantic thing, it just guards against `undefined` being redefined, since it's a value!)

That said, the "day-to-day-JS-way" would simply be

```js
doctor && params.command === "metals-doctor-reload"
```

TS will guarantee that types sort of makes sense and we don't actually have other potentially falsy values assigned to `doctor` :P

olafurpg: I like `doctor && params.command === "metals-doctor-reload"` 👍

PR #17, opened 2018-12-07
Detect when javaHome and serverVersion change and prompt to reload

gabro:
`javaHome` and `serverVersion` require re-launching the server with the new configuration, but we read them when the extension activates so we were not picking them up when using the `metals.restartServer` command.

I've experimented a bit with restarting the server by re-initializing the language client, and making the configuration dynamic, but it turned out to be **a lot** of added complexity for not much value.

This PR implements a super naive solution that simply prompts to reload the entire window, causing all extensions to start from scratch and so picking up the new configuration.

Considering that this should not be a common use case and that reload is pretty quick, I think this is an ok behavior, especially weighed against the simplicity of its implementation.

Here's a quick demo:

kapture 2018-12-07 at 13 32 34

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -36,6 +36,8 @@ import { LazyProgress } from "./lazy-progress";
 import * as fs from "fs";
 
 export async function activate(context: ExtensionContext) {
+  detectLaunchConfigurationChanges();
+
   const userJavaHome = workspace.getConfiguration("metals").get("javaHome");
   if (typeof userJavaHome === "string" && userJavaHome !== "") {
     fetchAndLaunchMetals(context, userJavaHome);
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -36,6 +36,8 @@ import { LazyProgress } from "./lazy-progress";
 import * as fs from "fs";
 
 export async function activate(context: ExtensionContext) {
+  detectLaunchConfigurationChanges();
```
gabro: Note: this is intentionally invoked early on.

If the launch goes wrong it's likely related to configuration, so you will be fiddling with those values.

olafurpg: 👍

PR #18, opened 2018-12-07
Use markdown for configuration description

gabro:
Super tiny improvements to the configuration section, after I found out about `markdownDescription`.

Before:

image

After:

image


olafurpg:
This is awesome, thanks @gabro !

PR #21, opened 2018-12-09
Add new configuration options

olafurpg:
Follow up from https://github.com/scalameta/metals/pull/387

The markdown descriptions are copy-pasted from the generated markdown, we might want to auto-generate the JSON on the website somehow to make it easier to update in the future.


olafurpg:
Seems camel case settings have nicer titles in the UI
screenshot 2018-12-09 at 19 35 10

gabro:
Ah wow, that's nice 👍

olafurpg:
I was able to autocomplete a `title` key but it wasn't picked up in the UI, so let's keep this it camel for now

gabro:
Yeah, I've tried `name` and `label`, but those are generic JSON Schema keys and it seems the UI ignored most of them

PR #26, opened 2018-12-09
Warn on out-of-date server version

### LINKED ISSUE #23: Warn users if serverVersion is older than default
Reporting an idea from @olafurpg 

Given:

- a `metals.serverVersion` value containing a semver-compliant version
- a `metals.serverVersion` set to an older release than the default value

we should warn the user upon extension activation.

I would probably exclude SNAPSHOT versions from this check.

gabro:
Closes #23

Tested also with SNAPSHOT versions:

If the current version is 0.3.1, this warns on:

- `0.3.0`
- `0.3.0+10-dba8c0aa-SNAPSHOT
- `0.3.1-SNAPSHOT`

but __does not__ warn on

- `0.3.1`
- `0.3.2`
- `0.3.1+15-dba8c0aa-SNAPSHOT`

kapture 2018-12-09 at 21 16 16

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,12 +118,14 @@
   "devDependencies": {
     "@types/mocha": "^2.2.42",
     "@types/node": "^8.10.25",
+    "@types/semver": "^5.5.0",
     "typescript": "^2.6.1",
     "vsce": "^1.53.2"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


laughedelic:
@gabro thanks! I saw #26 and thought the same 🙂

PR #27, opened 2018-12-09
Fix #24, print stdout in case download failed

### LINKED ISSUE #24: Log stdout on download error
Reported by @jvican

jvican

The error message is likely hidden in stdout but we don't log it because stdout should contain the fetched coursier classpath.

olafurpg:


olafurpg:
For https://github.com/scalameta/metals/issues/383 it wouldn't make a difference because the download succeeded, it only has an impact if `coursier fetch` fails.

gabro:
Ah, right. I got confused.

PR #29, opened 2018-12-11
Use infinite ttl for coursier fetch

olafurpg:
Here's the output I got this morning while using a SNAPSHOT version
screenshot 2018-12-11 at 11 38 58

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -86,6 +86,11 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
       coursierPath,
       "fetch",
       "-p",
+      "--ttl",
+      // Use infinite ttl to avoid redunant "Checking..." logs when using SNAPSHOT
+      // versions. Metals SNAPSHOT releases are effectively immutable since we
+      // never publish the same version twice.
+      "Inf",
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -86,6 +86,11 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
       coursierPath,
       "fetch",
       "-p",
+      "--ttl",
+      // Use infinite ttl to avoid redunant "Checking..." logs when using SNAPSHOT
+      // versions. Metals SNAPSHOT releases are effectively immutable since we
+      // never publish the same version twice.
+      "Inf",
```
gabro: I think this breaks the `publishLocal` workflow though

gabro: I mean, I can live with that, it's just a matter of deleting the artifact locally before re-launching the server, but I would document this.

olafurpg: I don't think it breaks the publish local flow since `publishLocal` overrides the jars themselves. The default ttl is 24hr so you would have already noticed if it breaks `publishLocal`

gabro: Got it. LGTM then

PR #33, opened 2018-12-12
Java Home detection improvements

### LINKED ISSUE #32: missing JAVA_HOME when using jenv results in unclear "make sure you have an internet connection" error
Running vscode with the java version set by jenv, but without setting "JAVA_HOME" gives an unclear error message:
```Failed to download Metals, make sure you have an internet connection and that the Metals version '0.3.1' is correct```

running it like this solves the problem: `jenv exec vscode .`. Maybe the error message could be more detailed, since jenv is pretty common in the wild.

Thanks!

### PR DESCRIPTION
**Pull request by gabro**
Fixes #32 

The main change is replacing `find-java-home` with [`locate-java-home`](https://github.com/jvilk/locate-java-home).

The latter detects all versions of Java installed on the system, so we can cherry-pick 1.8 if installed and error otherwise.

I've tested this on my machine and it works even when using jenv (because we directly check installed versions, instead of trying to guess which version is the user's default)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -146,7 +146,7 @@
     "vsce": "^1.53.2"
   },
   "dependencies": {
-    "find-java-home": "0.2.0",
+    "locate-java-home": "^1.1.2",
```
gabro: this is the main change. https://github.com/jvilk/locate-java-home can successfully detect all versions installed on the system, allowing us to cherry-pick java 8 if available.

On the other hand `find-java-home` returned a spurious value `~/.jenv` when using jenv and launching VSCode with `code /path/to/project`. A workaround was to launch VSCode with `jenv exect code /path/to/project`, but it wasn't really acceptable for end users.

src/getJavaHome.ts
```diff
@@ -0,0 +1,19 @@
+import locateJavaHome from "locate-java-home";
+import { workspace } from "vscode";
+
+export function getJavaHome(): Promise<string> {
+  const userJavaHome = workspace.getConfiguration("metals").get("javaHome");
+  if (typeof userJavaHome === "string" && userJavaHome.trim() !== "") {
+    return Promise.resolve(userJavaHome);
+  } else {
+    return new Promise((resolve, reject) => {
+      locateJavaHome({ version: "1.8" }, (err, javaHomes) => {
```
gabro: `locateJavaHome` will fail if it can't find Java 1.8 installed on the system

src/extension.ts
```diff
@@ -36,25 +36,25 @@ import {
...
-  } else {
-    require("find-java-home")((err, javaHome) => {
-      if (err) window.showErrorMessage("Unable to find Java home.");
-      fetchAndLaunchMetals(context, javaHome);
+  getJavaHome()
+    .then(javaHome => fetchAndLaunchMetals(context, javaHome))
+    .catch(err => {
+      const message = "Unable to detect Java home";
+      outputChannel.appendLine(message);
+      outputChannel.appendLine(err);
```
gabro: This should make for higher quality user reports in the future.

package.json
olafurpg: cc/ @laughedelic might be relevant for Atom

src/extension.ts
```diff
@@ -36,25 +36,25 @@ import {
...
-  if (typeof userJavaHome === "string" && userJavaHome !== "") {
-    fetchAndLaunchMetals(context, userJavaHome);
-  } else {
-    require("find-java-home")((err, javaHome) => {
-      if (err) window.showErrorMessage("Unable to find Java home.");
-      fetchAndLaunchMetals(context, javaHome);
+  getJavaHome()
+    .then(javaHome => fetchAndLaunchMetals(context, javaHome))
+    .catch(err => {
+      const message = "Unable to detect Java home";
```
olafurpg: Update message?

> Unable to find Java 8 home. To fix this problem, update the 'Java Home' setting in Metals to point to a Java 8 home directory.

gabro: image

Here it is

olafurpg: 😍

PR #36, opened 2018-12-14
Release v1.1.2

olafurpg:


olafurpg:
I pushed a tag v1.1.1 but it didn't trigger a release so I ended up publishing this from my local machine. v1.1.2 should be up on the marketplace soon.

olafurpg:
Pushed new tag to update the extension overview readme

PR #44, opened 2019-01-10
Implement metals/inputBox extension

olafurpg:
This endpoint will soon be used by Metals to allow users to write a
custom Scalafmt version to add to `.scalafmt.conf`.


olafurpg:
Implemented in server here https://github.com/scalameta/metals/pull/452

PR #50, opened 2019-01-12
Synchronize client commands with latest Metals.

olafurpg:
It seems we forgot to update these in the shift to kebab-case.
Also, "run doctor" was missing.

These are commands that metals expects the client to support, for
example "focus diagnostics" or "run doctor". Status bar messages
can have attached client commands so that for example if we update the
status bar with "Build mis-configuration" the user can click on it to
open the doctor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "@types/mocha": "^2.2.42",
     "@types/node": "^8.10.25",
     "@types/semver": "^5.5.0",
-    "typescript": "^2.6.1",
+    "typescript": "^3.2.2",
     "vsce": "^1.54.0"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


olafurpg:
Context: https://github.com/scalameta/metals/pull/466

gabro:
I've proposed a refactor in https://github.com/olafurpg/metals-vscode/pull/1, that should make future refactors slightly safer

olafurpg:
The CI job failed 🤔 I restarted it...

gabro:
Ah, the error is legitimate. I'm using `unknown` which was only recently introduced in TS, but we use an old version. I'll fix it on this PR

gabro:
(I didn't catch this locally, because I'm using the latest TS in VSCode)

package.json
```diff
@@ -138,7 +138,7 @@
     "@types/mocha": "^2.2.42",
     "@types/node": "^8.10.25",
     "@types/semver": "^5.5.0",
-    "typescript": "^2.6.1",
+    "typescript": "^3.2.2",
```
olafurpg: That was a smooth upgrade 😆

gabro: I'm surprised as well, the trick is that we use very little dependencies and mostly stable ones :D

PR #51, opened 2019-01-12
Add "cascade compile" and "cancel compile" tasks.

olafurpg:
Followup of https://github.com/scalameta/metals/pull/467

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,16 @@
         "title": "Rescan sources"
       },
       {
+        "command": "metals.compile-cascade",
+        "category": "Metals",
+        "title": "Cascade compile"
+      },
+      {
+        "command": "metals.compile-cancel",
+        "category": "Metals",
+        "title": "Cancel compilation"
+      },
+      {
         "command": "metals.doctor-run",
         "category": "Metals",
         "title": "Run doctor"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -93,6 +93,16 @@
...
+      {
+        "command": "metals.compile-cascade",
+        "category": "Metals",
+        "title": "Cascade compile"
+      },
+      {
+        "command": "metals.compile-cancel",
+        "category": "Metals",
+        "title": "Cancel compilation"
+      },
```
gabro: We should also add these to the `menus.commandPalette` section of the manifest, so that they only appear when the selected buffer is a Scala file

src/extension.ts
```diff
@@ -219,14 +219,19 @@ function launchMetals(
...
-      }
-    );
+    [
+      "build-import",
+      "build-connect",
+      "sources-scan",
+      "doctor-run",
+      "compile-cascade",
+      "compile-cancel"
+    ].forEach(command => {
```
gabro: Not strictly related to this PR, but I suspect something like this would work:

```typescript
JSON.parse(fs.readFileSync('package.json', 'utf8')).contributes.commands.forEach(({ command }) => {
 const [, commandName] = command.split('.');
 const cancel = commands.registerCommand(command, async () =>
   client.sendRequest(ExecuteCommandRequest.type, { command: commandName })
 );
 context.subscriptions.push(cancel);
})
```

olafurpg: This is a good idea but it requires automatically converting from camel case in package.json to kebab case here. I think it's better left to a separate PR, I'm fine with the manual process.

PR #54, opened 2019-01-24
Add -Dmetals.input-box property, fixes #53.

### LINKED ISSUE #53: Add -Dmetals.input-box=on by default
Currently, the Metals master enables this option by default for vscode clients but that breaks when using older vscode clients. It's better to let the extension declare what features it supports.

olafurpg:
This property is needed for the Scalafmt version dialogue.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -148,6 +148,7 @@ function launchMetals(
   enableScaladocIndentation();
 
   const baseProperties = [
+    `-Dmetals.input-box=on`,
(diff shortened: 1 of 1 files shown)


olafurpg:
It's already removed.

olafurpg:
Lesson learned for future addition of server properties, it's better to declare them in the extension than server. The `-Dmetals.client=sublime` property is only useful for editors where people need to run bootstrap themselves.

PR #58, opened 2019-01-24
Select latest JVM version when possible.

### LINKED ISSUE #47: Pick the most recent Java 8 version
If the user has Java 8 `_11` and `_191` installed then we should pick `_191`. See #46 for related discussion. The relevant line to fix this issue is here https://github.com/scalameta/metals-vscode/blob/c88649324c73817d49a20a4c030181ec74fc4554/src/getJavaHome.ts#L20

olafurpg:
Previously, we picked the first Java 1.8 home, which could cause Metals
to pick old security versions like _11 instead of _191. Fixes #47.

### DIFF
--- a/src/getJavaHome.ts
+++ b/src/getJavaHome.ts
@@ -13,6 +13,10 @@ export function getJavaHome(): Promise<string> {
         } else if (javaHomes.length === 0) {
           reject(new Error("No suitable Java version found"));
         } else {
+          // Sort by reverse security number so the highest number comes first.
+          javaHomes.sort((a, b) => {
+            return b.security - a.security;
+          });
           const jdkHome = javaHomes.find(j => j.isJDK);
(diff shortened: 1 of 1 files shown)


src/getJavaHome.ts
```diff
@@ -13,6 +13,10 @@ export function getJavaHome(): Promise<string> {
         } else if (javaHomes.length === 0) {
           reject(new Error("No suitable Java version found"));
         } else {
+          // Sort by reverse security number so the highest number comes first.
+          javaHomes.sort((a, b) => {
+            return b.security - a.security;
+          });
```
gabro: A bit more concise:

```js
javaHomes.sort((a, b) => b.security - a.security)
```

olafurpg: I love them curly braces :3

PR #64, opened 2019-01-26
Add issue templates

gabro:
99% copy-pasted from Metals main repo


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,41 @@
+---
+name: 🐛 Bug report
+about: Create a report to help us improve
+title: ""
+labels: "bug"
```
olafurpg: Can we leave out this label? I prefer to manually label issues once they've been validated as bugs.

PR #73, opened 2019-02-01
Upgrade to v0.4.2.

olafurpg:
It shouldn't affect VS Code since it always sends `workspaceFolders`.


gabro:
Still worth releasing for keeping versions in sync

PR #74, opened 2019-02-02
Automate CHANGELOG generation

gabro:
This should automatically re-generate `CHANGELOG.md` at every tag and commit it back to master.


gabro:
Only one way to know whether it's working...

PR #78, opened 2019-02-04
Fix shell quote types

### LINKED ISSUE #77: TypeError: line.startsWith is not a function
**Describe the bug**

With version 1.2.6 metals fails to start, leaving the following messages in the Output window:

```
Java home: /usr/lib/jvm/java-8-openjdk-amd64/
Using JAVA options set in JAVA_OPTS
Unable to find Java 8 home. To fix this problem, update the 'Java Home' setting to point to a Java 8 home directory
TypeError: line.startsWith is not a function
```

I have set the `Java Home` setting, but no avail.  I presume it's some other error swallowed by this [overly broad `catch`](https://github.com/scalameta/metals-vscode/blob/8ddd181a38a82dd36c36e4918601e8b000ed7456/src/extension.ts#L54).

Load some Scala workspace.  

**Expected behavior**

metals loads.

**Installation:**

- Operating system: Linux
- VSCode version: 1.30.2
- VSCode extension version: 1.2.6
- Metals version: (found in VSCode settings, under `metals.serverVersion`): 0.4.4

**Additional context**

I searched for `startsWith`, and it only appears twice, and [this code](https://github.com/scalameta/metals-vscode/blob/8c338f12c55588e7337803a87360dd482733e972/src/getJavaOptions.ts#L33) looks relevant, since it changed in 1.2.6.  

Guessing, I'd that `shell-quote.parse` return an [array with some non-strings inside](https://www.npmjs.com/package/shell-quote#parsing-shell-operators).  It's [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/shell-quote/index.d.ts) also imply that it can return non-strings in some cases.  I know that I suggested the package, but I wasn't aware of this behaviour, it just looked like the best package to use for this job.

I wonder why typescript compiled nonetheless?  You didn't add the typings as far as I can see (I guess you should), but without types tsc should have warned about inferring `any`, shouldn't it?

swiesner-dlr:
Use typings from DefinitelyTyped and ignore non-string results.

Fixes #77 

/cc @gabro


src/getJavaOptions.ts
```diff
@@ -15,11 +15,18 @@ function jvmOpts(outputChannel: OutputChannel): string[] {
 
 function javaOpts(outputChannel: OutputChannel): string[] {
   const javaOpts = process.env.JAVA_OPTS;
+  const options = [];
   if (javaOpts) {
     outputChannel.appendLine("Using JAVA options set in JAVA_OPTS");
-    return parse(javaOpts);
+    for (const option of javaOpts) {
+      if (typeof option === 'string') {
+        options.push(option)
```
gabro: I'm probably being extra picky, but I don't really like loops and mutability (even in JS :D)

For consistency with the rest of the code, you would mind refactoring this to use `.reduce` or other combinators?

swiesner-dlr: I don't either, and I tried, but combinators don't work well with type checks, i.e. `.filter(typeof _ === string)` doesn't propagate the type constraint to the result.

I'll try again, and harder :blush:

gabro: yeah, filter is particularly stupid about it.
You can use a custom guard to make it smarter I think

```
function isString(x: unknown): x is string {
  return typeof x === 'string'
}
```

or using `reduce` (which you might need anyway to log the errors) should fix it

swiesner-dlr: @swiesner-dlr Thanks.  My typescript foo isn't the best :blush:

PR #80, opened 2019-02-04
Bail when no directory has been opened and enable strict mode in TypeScript

### LINKED ISSUE #79: Metals fails to launch, when not in a workspace
When opening a file, without opening a workspace/folder beforehand, Metals doesn't launch, but instead reports the following error:

> Unable to find Java 8 home. To fix this problem, update the 'Java Home' setting to point to a Java 8 home directory
> TypeError: Path must be a string. Received undefined

The problem is that during the configuration of the launch, 
the `workspace.rootPath` is used twice:

1. https://github.com/scalameta/metals-vscode/blob/2e68c2c18b8986bcf9b30825d91fa2d6fd7764a5/src/extension.ts#L467

2. https://github.com/scalameta/metals-vscode/blob/2e68c2c18b8986bcf9b30825d91fa2d6fd7764a5/src/getJavaOptions.ts#L7

path.join will fail in this case, because `workspace.rootPath` is `undefined`.

Two possible solutions:
Test if `workspace.rootPath` is undefined and then
1. ignore the rest of the configuration or
2. Return a meaningful error message, telling that metals only works in a workspace/folder.

gabro:
This PR enables the `strict: true` compiler option, which finally highlighted a lot of potentially unsafe accesses to `any` or `undefined` structures.

This also turns the bug reported in #79 into a type error 🎉 

The PR fixes #79 and all other potentially unsafe accesses.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -65,10 +65,17 @@ export async function activate(context: ExtensionContext) {
 }
 
 function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
-  if (fs.existsSync(dottyIdeArtifact())) {
+  if (!workspace.workspaceFolders) {
+    outputChannel.appendLine(
+      `Metals will not start because you've opened a single file and not a project directory.`
+    );
+    return;
+  }
+  const dottyArtifact = dottyIdeArtifact();
+  if (dottyArtifact && fs.existsSync(dottyArtifact)) {
     outputChannel.appendLine(
       `Metals will not start since Dotty is enabled for this workspace. ` +
-        `To enable Metals, remove the file ${dottyIdeArtifact()} and run 'Reload window'`
+        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
     );
     return;
   }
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -65,10 +65,17 @@ export async function activate(context: ExtensionContext) {
 }
 
 function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
-  if (fs.existsSync(dottyIdeArtifact())) {
+  if (!workspace.workspaceFolders) {
+    outputChannel.appendLine(
+      `Metals will not start because you've opened a single file and not a project directory.`
```
olafurpg: 👍 Maybe we can do better in the future but this is already a big improvement over crashing

src/extension.ts
```diff
@@ -463,8 +470,13 @@ function enableScaladocIndentation() {
   });
 }
 
-function dottyIdeArtifact(): string {
-  return path.join(workspace.rootPath, ".dotty-ide-artifact");
+function dottyIdeArtifact(): string | undefined {
+  if (workspace.workspaceFolders) {
```
gabro: I've also started using `workspaceFolders` instead of the deprecated `rootPath`. As per the documentation `rootPath` == `worspaceFolders[0]`

src/extension.ts
gabro: this is the fix for #79

gabro:
Thanks for the review @olafurpg. `!` essentially casts `A | undefined` into `A`.

PR #81, opened 2019-02-04
Improve error messages for failed download of Metals

### LINKED ISSUE #67: Confusing error message: Failed to download Metals
**Describe the bug**

The following error message is confusing: "Failed to download Metals, make sure you have an internet connection, the Metals version '0.4.0' is correct and the Java Home '/usr/lib/jvm/java-1.8.0-openjdk/bin/java' is valid. You can configure the Metals version and Java Home under settings."

Suggestions for improving the message

- test if env variable `FLATPAK_SANDBOX_DIR` is defined, if so may be to blame https://flatpak.org/
- don't warn againt the version '0.4.0' if it's the default version, we know it's correct then

gabro:
Fixes #67 

Changes:

- explicitly detect Flatpak (using `FLATPAK_SANDBOX_DIR`)
- avoid suggesting to check the version if it's the default one

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -86,9 +86,11 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
 
   const config = workspace.getConfiguration("metals");
   const serverVersionConfig: string = config.get<string>("serverVersion")!;
+  const defaultServerVersion = config.inspect<string>("serverVersion")!
+    .defaultValue!;
   const serverVersion = serverVersionConfig
     ? serverVersionConfig
-    : config.inspect("serverVersion")!.defaultValue;
+    : defaultServerVersion;
   const serverProperties: string[] = workspace
     .getConfiguration("metals")
     .get("serverProperties")!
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -98,7 +100,7 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
 
   const javaOptions = getJavaOptions(outputChannel);
 
-  const process = spawn(
+  const fetchProcess = spawn(
```
gabro: `process` was shadowing `global.process`

src/extension.ts
```diff
@@ -134,10 +136,26 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
-      const msg =
-        `Failed to download Metals, make sure you have an internet connection, ` +
-        `the Metals version '${serverVersion}' is correct and the Java Home '${javaPath}' is valid. ` +
-        `You can configure the Metals version and Java Home under settings.`;
+      const msg = (() => {
+        if (process.env.FLATPAK_SANDBOX_DIR) {
+          return (
+            `Failed to download Metals. It seems you are running Visual Studio Code inside the` +
+            `Flatpak sandbox, which is known to interfere with the download of Metals.` +
+            `Please, try running Visual Studio Code without Flatpatk.`
```
olafurpg: > flatpak

gabro: nice catch

PR #88, opened 2019-03-18
Experiment with release drafter

olafurpg:
More information https://github.com/toolmantim/release-drafter/blob/master/README.md

Thank you @gabro for the suggestion!


olafurpg:
Seems to be working 👍 
Screenshot 2019-03-18 at 11 20 35

PR #92, opened 2019-04-02
 Link to HTTP proxy instructions in "failed to download" error.

### LINKED ISSUE #84: Improve error message for failed download due to HTTP proxy
Followup of https://github.com/scalameta/metals/issues/529

We should document how to configure HTTP proxies if the Metals download step fails 

https://github.com/scalameta/metals-vscode/blob/f8a49bae54af41a722beab860495879de7431dd2/src/extension.ts#L148-L155

olafurpg:
Fixes #84. Previously, the download error message only said "make sure
you have an internet connection" when the most common cause for this
problem was HTTP proxies. Now, the error message links to the Metals
website for how to configure HTTP proxies.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -143,7 +143,7 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
       const msg = (() => {
         const proxy =
           `See https://scalameta.org/metals/docs/editors/vscode.html#http-proxy for instructions ` +
-          `if you are using a corporate HTTP proxy.`;
+          `if you are using an HTTP proxy.`;
         if (process.env.FLATPAK_SANDBOX_DIR) {
           return (
             `Failed to download Metals. It seems you are running Visual Studio Code inside the` +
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -139,6 +141,9 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
     },
     () => {
       const msg = (() => {
+        const proxy =
+          `See https://scalameta.org/metals/docs/editors/vscode.html#http-proxy for instructions ` +
```
gabro: 👍 for linking to the website. It's worth to keep in mind that we better not break those urls going forward

src/extension.ts
```diff
@@ -139,6 +141,9 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
+          `if you are using a corporate HTTP proxy.`;
```
```suggestion
          `if you are using an HTTP proxy.`;
```

Not all proxies are corporate (e.g. universities)

olafurpg: Good point 👍

PR #94, opened 2019-04-02
Update default for "editor.suggestSelection" to "first", fixes #86.

### LINKED ISSUE #86: Override default value for editor.suggestSelection setting to be "first"
The default behavior in VS Code is to preselect completion items based on what you have selected before. For example, if you select `toString` once then vscode will preselect that method the next time it appears. However, Metals has better understanding of the context of the source code, for example we use type-aware sorting to highlight the most relevant completion items in pattern matching

Screenshot 2019-03-08 at 13 48 13

Related discussion https://github.com/Microsoft/vscode/issues/54203#issuecomment-404797260

The Metals VS Code extension should either 
- automatically set the value of `editor.suggestSelection` to be `"first"` in Scala sources
- highly recommend that users configure this setting when they use Metals for the first time

olafurpg:
Previously, the default "editor.suggestSelection" value was "last"
causing VS Code to pre-select counter-intuitive completions items.
For example, if you selected `toString()` once, then VS Code would
pre-select `toString()` in the following completion regardless of the
order of completions returned by Metals.

Metals already orders completion items in a "smart" order based on the
local scope and completion position (pattern matching vs. type vs.
term). Now with this change, VS Code always pre-selects the first
completion item returned by Metals.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,11 @@
     "workspaceContains:project/build.properties"
   ],
   "contributes": {
+    "configurationDefaults": {
+      "[scala]": {
+        "editor.suggestSelection": "first"
+      }
+    },
     "configuration": {
       "title": "Metals",
       "properties": {


olafurpg:
Although completions are not part of the latest stable release, this fix will already benefit people using the latest snapshot.

Gama11:
Urm... are you sure thie ever worked? https://github.com/microsoft/vscode/issues/91180#issuecomment-591495127

PR #101, opened 2019-04-24
Add gradle setting to visual-code plugin and enable scala dir detection for non-sbt build tools

tgodzik:


.gitignore
```diff
@@ -2,4 +2,5 @@ out
 node_modules
 .vscode-test/
 *.vsix
+*.lock
```
gabro: why? yarn.lock is meant to be checked out in version control

package.json
```diff
@@ -71,6 +73,10 @@
           "type": "string",
           "markdownDescription": "Optional absolute path to an `sbt` executable to use for running `sbt bloopInstall`.\n\nBy default, Metals uses `java -jar sbt-launch.jar` with an embedded launcher while respecting `.jvmopts` and `.sbtopts`.\n\nUpdate this setting if your `sbt` script requires more customizations like using environment variables."
         },
+        "metals.gradleScript": {
+          "type": "string",
+          "markdownDescription": "Optional absolute path to an `gradle` executable to use for running `gradle bloopInstall`.\n\nBy default, Metals uses embeded gradlew script.\n\nUpdate this setting if your `gradle` script requires more customizations like using environment variables."
```
gabro: typo: embeded

gabro: s/an gradle/a gradle

.gitignore
tgodzik: och, I thought it might just be a mistake. Didn't know lock files were sensible to be included

package.json
```diff
@@ -117,32 +123,25 @@
     "menus": {
       "commandPalette": [
         {
-          "command": "metals.restartServer",
-          "when": "editorLangId == scala"
+          "command": "metals.restartServer"
```
tgodzik: I don't think we need the when clause, since we might be in a txt file and still run the import build. Besides with more and more supported tools this might be harder to maintain.

.gitignore
gabro: yeah, the purpose of yarn.lock is to "lock" the dependency tree to ensure the build is reproducible.

Unfortunate name, but there have been worse ones in this space (google "npm shrinkwrap" for an example 😅)

package.json
gabro: that's a concern that I have as well. I would prefer not to change it in this PR, but I'm open to discuss it further.

The reason why the scoping was introduced was due to confusing errors when running the commands without the extension being active.

I don't love this solution either, so let maybe open an issue for that?

olafurpg: I agree with @gabro, let's move this discussion into a separate issue since it's not strictly related to the Gradle setting.

tgodzik: Reverted, should we add the possibility at least for gradle files?  `resourceExtname == .gradle`

package.json
```diff
@@ -43,7 +43,8 @@
   "activationEvents": [
     "onLanguage:scala",
     "workspaceContains:build.sbt",
-    "workspaceContains:project/build.properties"
+    "workspaceContains:project/build.properties",
+    "workspaceContains:**/scala/**"
```
tgodzik: @gabro @olafurpg what about this? It would work for both gradle and maven, which could normally also not have a scala source.

gabro: I think this would generate many false positives with other builds tools Metals does not support, but it's not the end of the world, since Metals would bail anyway.

Out of curiosity, is there any way we can be more specific to Gradle and Maven?

tgodzik: >> I think this would generate many false positives with other builds tools Metals does not support, but it's not the end of the world, since Metals would bail anyway.

But if someone opens a project and it doesn't have any supported build tool then the message about the import should not pop up I think, so shouldn't be an issue.

>> Out of curiosity, is there any way we can be more specific to Gradle and Maven?

We would need to somehow parse the files or run the build tool, which I don't think is possible here.

src/extension.ts
```diff
@@ -63,6 +63,7 @@ export async function activate(context: ExtensionContext) {
         }
       });
     });
+  commands.executeCommand("setContext", "metals:enabled", true);
```
marek1840: Could it be disabled when the server is restarting? Since we cannot run those commands then anyway

package.json
gabro: 👍 fine by me, then!

src/extension.ts
gabro: 😍 this is a beautiful trick I didn't know about!

tgodzik: The trick doesn't seem to work with deactivating the commands unfortunately

gabro: I don't think it's a big deal. If the user triggers a server restart they're aware it's happening and it shouldn't be a big surprise that commands don't work for a little while.

olafurpg: This is awesome! I've been frustrated by not having "Metals: Import build" when editing non-scala files.

PR #104, opened 2019-05-05
Add metals/windowStateDidChange notification

agajek:
Publish `metals/windowStateDidChange` notification to indicate whether
vscode window is focused or not.

The client-side implementation of the https://github.com/scalameta/metals/issues/668

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -34,7 +34,9 @@ import {
   MetalsStatus,
   MetalsDidFocus,
   ExecuteClientCommand,
-  MetalsInputBox
+  MetalsInputBox,
+  MetalsWindowStateDidChange,
+  MetalsWindowStateDidChangeParams
 } from "./protocol";
 import { LazyProgress } from "./lazy-progress";
 import * as fs from "fs";
@@ -354,6 +356,10 @@ function launchMetals(
           editor.document.uri.toString()
         );
       }
+    });
+
+    window.onDidChangeWindowState(windowState => {
+      client.sendNotification(MetalsWindowStateDidChange.type, { focused: windowState.focused })
     });
 
     client.onRequest(MetalsInputBox.type, (options, requestToken) => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -356,6 +358,10 @@ function launchMetals(
       }
     });
 
+    window.onDidChangeWindowState(windowState => {
```
olafurpg: Should we guard this against the server version number? This would require merging the PR to the main Metals repo first and wait until there is a SNAPSHOT release out. We already depend on a semver library.

gabro: What happens if you send the notification to the current server? Doesn't it just discard it or log it with a warning?

olafurpg: It logs a warning, which would probably confuse users.

gabro: I'm 👍 for adding a guard here, but I would remove it in the future when the server version targeted by the VSCode extension supports it.

olafurpg: Alternatively, the server could register its interest in these notifications similar to how file watchers work. That might be overkill however

gabro: I agree it's overkill. Let's add a guard and remove it when we release the next version of the server.

agajek: Good spot, but how can I do this? And also correct me if I'm wrong but I think that when we merge the server part that guard will be unnecessary, am I right?

gabro: @agajek correct, they idea is to merge this with the guard on, and to remove it when Metals 0.6.0 is released. I would use the semver package to do something like:

```ts
const serverVersion = config.get<string>("serverVersion")!;
if (semver.gte(serverVersion, "0.6.0")) {
  // ...
}
```

(I wrote this directly as a comment, it might be wrong, but it's just to give an idea)

olafurpg: The `serverVersion` variable is already computed here https://github.com/scalameta/metals-vscode/blob/1cc6b239ac7505c35900b65717a0b864cc2fd0aa/src/extension.ts#L94

agajek:
@olafurpg where can I find this version that should be included in the guard?

olafurpg:
https://scalameta.org/metals/docs/editors/vscode.html#using-latest-metals-snapshot

PR #106, opened 2019-05-16
Run code using Debug  Protocol Adapter

marek1840:


src/debugger.ts
```diff
@@ -0,0 +1,55 @@
+import * as vscode from "vscode"
+import {CancellationToken, DebugConfiguration, Disposable, ProviderResult, WorkspaceFolder} from "vscode"
+import {number} from "vscode-languageclient/lib/utils/is";
+
+export const runScala = "metals-scala-run";
+export const startSession = "debug-session-start";
+const configurationType = "scala";
+
+export async function run(buildTarget: BuildTargetIdentifier, mainClass: string): Promise<Boolean> {
```
olafurpg: Is it possible to automatically register a "launch configuration" if it doesn't exist before? https://code.visualstudio.com/docs/editor/debugging#_launch-configurations This would make it possible to trigger the run via keyboard shortcut after clicking on "Run" for the first time.

olafurpg: This can be left for a separate PR.

src/debugger.ts
```diff
@@ -0,0 +1,55 @@
+import * as vscode from "vscode"
+import {CancellationToken, DebugConfiguration, Disposable, ProviderResult, WorkspaceFolder} from "vscode"
+import {number} from "vscode-languageclient/lib/utils/is";
+
+export const runScala = "metals-scala-run";
```
olafurpg: These keys should be defined in `client-commands.ts`

src/debugger.ts
```diff
@@ -0,0 +1,55 @@
+import * as vscode from "vscode"
+import {CancellationToken, DebugConfiguration, Disposable, ProviderResult, WorkspaceFolder} from "vscode"
```
olafurpg: Please format the source code with prettier, that's what me and @gabro currently use even we don't enforce it in CI (but probably should).

src/extension.ts
```diff
@@ -41,6 +41,7 @@ import * as fs from "fs";
 import * as semver from "semver";
 import { getJavaHome } from "./getJavaHome";
 import { getJavaOptions } from "./getJavaOptions";
+import * as scalaDebugger from "./debugger";
```
olafurpg: Can we use the name `debugger` instead? I prefer to avoid new names unless it's necessary (conflict with another `debugger` for example).

package.json
```diff
@@ -145,7 +145,16 @@
           "when": "editorLangId == scala"
         }
       ]
-    }
+    },
+    "debuggers": [
+      {
+        "type": "scala",
+        "label": "scala debugger",
```
olafurpg: Does this label appear in the UI? If so, should it maybe be named "Scala Debugger" or "Metals Debugger"?

src/extension.ts
marek1840: we cannot use the `debugger` name. when trying to do so I get error `Identifier expected.ts(1003)`

package.json
marek1840: done

src/extension.ts
olafurpg: Hmm, it looks like `debugger` is a JavaScript keyword https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/debugger

Can we rename the file to `scalaDebugger` in that case? I think it's good to try and keep the variable name consistent with the imported filename.

src/extension.ts
```diff
@@ -329,12 +334,43 @@ function launchMetals(
...
       }
     };
     Object.entries(clientCommands).forEach(([name, command]) =>
       registerCommand(name, command)
     );
 
+    scalaDebugger
+      .initialize()
+      .forEach(disposable => context.subscriptions.push(disposable));
+    registerCommand(scalaDebugger.startSessionCommand, scalaDebugger.start);
```
olafurpg: Can we optionally enable the debugger based on whether the Metals language server supports debugging? We currently do this for the tree view by looking at the server capabilities during `initialize` https://github.com/scalameta/metals-vscode/blob/7a80af4af34c361ca20424ee6b5d950980121d52/src/MetalsFeatures.ts#L20

src/MetalsFeatures.ts
```diff
@@ -8,6 +8,8 @@ export interface TreeViewProvider {}
 
 export class MetalsFeatures implements StaticFeature {
   treeViewProvider?: TreeViewProvider;
+  supportsDebugging?: Boolean;
```
olafurpg: For consistency with capability providers in LSP
```suggestion
  debuggingProvider?: DebuggingProvider;
```

This can be an empty object for now

```ts
interface DebuggingProvider {}
```

The benefit of this approach is that we can add more fields in the future to configure what kind of debugging functionality is supported.

src/extension.ts
```diff
@@ -329,12 +334,40 @@ function launchMetals(
...
     Object.entries(clientCommands).forEach(([name, command]) =>
       registerCommand(name, command)
     );
 
+    // should be the compilation of a currently opened file
+    // but some race conditions may apply
+    let compilationDoneEmitter = new EventEmitter<void>();
+
+    let codeLensRefresher: CodeLensProvider = {
+      onDidChangeCodeLenses: compilationDoneEmitter.event,
```
olafurpg: Cool! I didn't know about this

src/scalaDebugger.ts
```diff
@@ -0,0 +1,65 @@
...
+  outputChannel.appendLine("Initializing Scala Debugger");
+  return [
+    vscode.debug.registerDebugConfigurationProvider(
+      configurationType,
+      new ScalaConfigProvider()
+    )
+  ];
+}
+
+export async function start(parameters: any): Promise<Boolean> {
```
gabro: nit: no need to make the function `async` if we're using the normal Promise api in the implementation (which I personally prefer, anyway)

```suggestion
export function start(parameters: any): Promise<Boolean> {
```

src/extension.ts
```diff
@@ -329,12 +334,40 @@ function launchMetals(
...
+      provideCodeLenses: (
+        document: VscodeTextDocument,
+        token: CancellationToken
+      ) => undefined,
+      resolveCodeLens: (codeLens: CodeLens, token: CancellationToken) =>
+        undefined
```
gabro: `resolveCodeLens` is optional, it can be removed.

src/scalaDebugger.ts
```diff
@@ -0,0 +1,65 @@
...
+  return vscode.commands
+    .executeCommand<DebugSession>(startAdapterCommand, parameters)
+    .then(response => {
+      if (response === undefined) return false;
+
+      const debugServer = vscode.Uri.parse(response.uri);
+      const segments = debugServer.authority.split(":");
+      const port = parseInt(segments[segments.length - 1]);
```
gabro: potentially, you could use `Url` from node

```ts
import * as Url from "url";

// ...

const port = parseInt(Url.parse(response.uri).port!)
```

PR #111, opened 2019-06-10
Update metals default version to 0.6.0

tgodzik:


package.json
```diff
@@ -57,7 +57,7 @@
       "properties": {
         "metals.serverVersion": {
           "type": "string",
-          "default": "0.5.2",
+          "default": "0.6.0",
```
olafurpg:
```suggestion
          "default": "0.6.1",
```

It's already out on Maven 🎉

PR #119, opened 2019-06-18
Remove "work in progress" from the description

gabro:
Followup of https://github.com/scalameta/metals/pull/782

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,12 @@
 {
   "name": "metals",
   "displayName": "Scala (Metals)",
-  "description": "Language Server for Scala",
+  "description": "Scala language server with rich IDE features",
   "keywords": [
     "scala",
     "lsp",
     "language server",
+    "ide",
     "scalameta"
   ],
   "version": "1.4.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "metals",
   "displayName": "Scala (Metals)",
-  "description": "Work in progress language server for Scala",
+  "description": "Language Server for Scala",
```
olafurpg: I feel like we should come up with a slightly more detailed description first. Something like “Language server for Scala with rich IDE features”. I’m totally sold on removing the WIP title and updating it everywhere (website, marketplace,...)

gabro: Isn't it a bit redundant? Other language servers follow pretty much the same pattern: "X language server", e.g. https://github.com/eclipse/eclipse.jdt.ls

olafurpg: It would be nice to have “IDE” in the title to help match search queries

tgodzik: We could add that to keywords

gabro: With that in mind I'm 👍 for "Scala language server with rich IDE feature" and also adding "ide" to the keywords

gabro:
Better?

PR #122, opened 2019-06-24
Implement Tree View Protocol (TVP).

olafurpg:
This commit implements a new LSP extension called "Tree View Protocol"
which enables the language server to implement tree views such as:

- package explorer for library dependencies
- dynamically updating compilation progress overviews

The protocol is implemented as a thin shim on top of the VS Code tree
view API exposing the following functionality:

- List children: client->server request to go from a tree view node to
  its children nodes.
- Get parent: client->server request go from a tree view node to its
  parent node, if any.
- View visibility did change: client->server notification to notify the
  server that the visibility of a tree view changed.
- Node did change: server->client notification to notify the editor a
  tree view has updated.
- Reveal node: server->client notification to request the editor to
  focus on a tree view node.

Tree view nodes can have the following metadata:

- label: the title
- tooltip: shown on hover
- command: to execute when user clicks on the node
- icon: an svg to display next to the title
- collapse state: whether the node has children and if the node should auto-expand

The tree views are only started if the server supports the tree view
extensions. The server registers TVP support by adding the
`treeViewProvider: {}` field in the `initialize` LSP handshake.

Below are some example demos for tree views that the server can implement with this API.


olafurpg:
I created the svg icons in https://vectr.com/ and I double checked that they look OK in a light theme as well. I made the "scala class" icon a bit brighter since recording the screenshots and GIFs in this PR.

Screenshot 2019-06-24 at 21 38 49

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+      resourceUri:
+        item.nodeUri && item.nodeUri.indexOf(".") > 0
+          ? Uri.parse(item.nodeUri)
+          : undefined,
+      collapsibleState:
+        item.collapseState == "expanded"
+          ? TreeItemCollapsibleState.Expanded
+          : item.collapseState == "collapsed"
+          ? TreeItemCollapsibleState.Collapsed
+          : TreeItemCollapsibleState.None,
```
gabro: can we move this to a utility function?

```ts
function toTreeItemCollapsibleState(s: TreeViewNode["collapseState"]): TreeItemCollapsibleState {
  switch (s) {
    case "expanded": return TreeItemCollapsibleState.Expanded;
    case "collapsed": return TreeItemCollapsibleState.Collapsed;
    case "none": return TreeItemCollapsibleState.None;
  }
}
```

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  // Populate TreeItem based on cached children response from the server.
+  getTreeItem(uri: string): TreeItem {
+    const item = this.items.get(uri);
+    if (!item) return {};
+    const result: TreeItem = {
+      label: item.label,
+      id: item.nodeUri,
+      resourceUri:
+        item.nodeUri && item.nodeUri.indexOf(".") > 0
```
gabro:
```suggestion
        item.nodeUri && item.nodeUri.includes(".")
```

src/MetalsFeatures.ts
```diff
@@ -0,0 +1,18 @@
...
+  StaticFeature,
+  InitializeParams,
+  ServerCapabilities
+} from "vscode-languageclient";
+
+export interface TreeViewProvider {}
+
+export class MetalsFeatures implements StaticFeature {
+  treeViewProvider?: TreeViewProvider;
+  fillInitializeParams?: (params: InitializeParams) => void;
```
gabro: This method can be omitted since it's optional and it's not implemented

```suggestion
```

src/extension.ts
```diff
@@ -36,17 +40,22 @@ import {
...
 import { getJavaHome } from "./getJavaHome";
 import { getJavaOptions } from "./getJavaOptions";
+import { startTreeView } from "./treeview";
+import ProtocolCompletionItem from "vscode-languageclient/lib/protocolCompletionItem";
+import { MetalsFeatures } from "./MetalsFeatures";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
 const openSettingsCommand = "workbench.action.openSettings";
+var treeViews: MetalsTreeViews | undefined = undefined;
```
```suggestion
let treeViews: MetalsTreeViews | undefined = undefined;
```

src/extension.ts
```diff
@@ -331,7 +342,39 @@ function launchMetals(
           const panel = getDoctorPanel(isReload);
           panel.webview.html = html;
         }
+      } else {
+        switch (params.command) {
+          case "metals-goto-location":
+            const location =
+              params.arguments && (params.arguments[0] as Location);
```
gabro: I really don't like that we're casting at use site, but this is not strictly related to this PR.

Ideally, we should validate at the networking layer and use well-typed models after that. We do this at work using io-ts, I can experiment with it sometime in the near future.

It's 👍 for this PR.

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+  MetalsTreeViewVisibilityDidChange,
+  MetalsRevealTreeView,
+  MetalsTreeViewParent
+} from "./protocol";
+
+export function startTreeView(
+  client: LanguageClient,
+  out: OutputChannel,
+  context: ExtensionContext,
+  packageJson: any
```
gabro: same here, it would be good to have some level of validation beforehand, so that we don't pass `any` values around.

src/extension.ts
```diff
@@ -436,6 +486,16 @@ function launchMetals(
         );
       });
     });
+    if (features.treeViewProvider) {
+      const packageJson = JSON.parse(
+        fs.readFileSync(
+          path.join(context.extensionPath, "package.json"),
+          "utf8"
+        )
+      );
```
gabro: `require("./package.json")` should work, in theory.

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+): MetalsTreeViews {
+  let allProviders: Map<string, MetalsTreeDataProvider> = new Map();
```
gabro:
```suggestion
  const allProviders: Map<string, MetalsTreeDataProvider> = new Map();
```

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+  let allViews: Map<string, TreeView<string>> = new Map();
```
gabro:
```suggestion
  const allViews: Map<string, TreeView<string>> = new Map();
```

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+  let viewIds: string[] = packageJson.contributes.views["metals-explorer"].map(
+    (view: { id: string }) => view.id
+  ) as string[];
```
gabro: this cast is redundant (and probably also the type annotation)
```suggestion
  );
```

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export function startTreeView(
+  client: LanguageClient,
+  out: OutputChannel,
+  context: ExtensionContext,
+  packageJson: any
+): MetalsTreeViews {
+  let allProviders: Map<string, MetalsTreeDataProvider> = new Map();
+  let allViews: Map<string, TreeView<string>> = new Map();
+  let viewIds: string[] = packageJson.contributes.views["metals-explorer"].map(
```
gabro:
```suggestion
  const viewIds: string[] = packageJson.contributes.views["metals-explorer"].map(
```

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+    (view: { id: string }) => view.id
+  ) as string[];
+  const disposables = viewIds.map(viewId => {
+    let provider = new MetalsTreeDataProvider(
```
gabro:
```suggestion
    const provider = new MetalsTreeDataProvider(
```

src/treeview.ts
```diff
@@ -0,0 +1,186 @@
...
+      });
+    });
+
+    return [view, onDidChangeVisibility];
+  });
+
+  // Update tree nodes on server notificiations
+  client.onNotification(MetalsTreeViewDidChange.type, params => {
+    params.nodes.forEach(node => {
+      let provider = allProviders.get(node.viewId);
```
gabro:
```suggestion
      const provider = allProviders.get(node.viewId);
```

images/sample.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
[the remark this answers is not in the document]
olafurpg: Good catch, removed!

src/MetalsFeatures.ts
olafurpg: I just implemented this method to add the experimental `treeViewProvider` field.

src/extension.ts
olafurpg: Fixed.

src/extension.ts
olafurpg: Validating the input here would be nice  and would benefit the other client commands that also cast 👍

src/extension.ts
olafurpg: > in theory

I thought so as well but I wasted probably 30 minutes trying to get it working and eventually gave up and went this route.

I did a quick retry  and it seems to be reading a very old version of package.json

Screenshot 2019-06-26 at 19 49 01

Running `yarn install` didn't fix the problem.

I left a comment explaining.

src/treeview.ts
olafurpg: I moved the `const viewIds= ...` logic to `extension.ts` and changed the param to be `viewIds: string[]`. I agree that passing `any` around is bad.

src/treeview.ts
olafurpg: Fixed all of these.

src/treeview.ts
olafurpg: Right, makes sense. A bit shady that ascribing the parameter type in the lambda casts the type 😨

src/treeview.ts
olafurpg: Aha, I tried `.contains` and completions gave me nothing so I tried `indexOf` and was happy 😅

src/treeview.ts
olafurpg: Great idea, much clearer. I first tried to implement this helper method on the interface itself but that didn't work (which makes sense when I think of the JS runtime but feels unsatisfying given my JVM background).

src/extension.ts
```diff
@@ -239,6 +250,8 @@ function launchMetals(
     serverOptions,
     clientOptions
   );
+  const features = new MetalsFeatures();
+  client.registerFeature(features);
 
   function registerCommand(command: string, callback: (...args: any[]) => any) {
```
gabro: I just remembered that we already have `client-commands.ts` where we define the list of client commands, that are then automatically registered.

If possible, it would be nice to centralize the tvp commands as well.

olafurpg: I made a quick attempt to use that but failed somehow. Don't remember the details. I agree it would be good to consolidate the logic for registering commands.

PR #128, opened 2019-07-16
Show warning when using snaps

tgodzik:


readme.md
```diff
@@ -153,6 +155,9 @@ Use the 'Custom Repositories' setting for the Metals VS Code extension to tell
 [Coursier](https://get-coursier.io/docs/other-proxy) to try to download Metals
 artifacts from your private artifact repository.
 
+IMPORTANT : The 'Custom Repositories' setting uses spaces instead of `|` to
```
tgodzik: This is only real change.

gabro: this seems very ad hoc. What makes us think that the user will read this when they didn't read the setting description, which is even more "local"?

Can't we be more resilient and accept also `|` as a separator?

My actual favorite solution would be to make this an array setting, but we deemed it too user unfriendly since VS Code lacks the UI for it.

tgodzik: Och, actually it should work with `|` already :O

I am bad at noticing things today.

PR #135, opened 2019-08-09
Client command registration fix

v66moroz:
Registration is using `ClientCommands` keys instead of names which leads to an error message when clicking on  compilation status (specifically `metals-diagnostics-focus`)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -330,8 +330,9 @@ function launchMetals(
         }
       }
     };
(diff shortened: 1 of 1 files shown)


tgodzik:
@olafurpg should we merge this? Not sure about what we actually fix here, but my knowledge of typescript is a bit limited.

gabro:
@v66moroz thanks for this. It makes a ton of sense and it was indeed a bug!
I've included your fix in #147, which now supersedes this PR.

PR #141, opened 2019-09-06
Use type array for serverProperties and customProperties settings

gabro:
Starting from last July VSCode release, this type of setting finally has a UI 🎉

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -108,9 +108,7 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
     : defaultServerVersion;
   const serverProperties: string[] = workspace
     .getConfiguration("metals")
-    .get<string>("serverProperties")!
-    .split(" ")
-    .filter(e => e.length > 0);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -108,9 +108,7 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
     : defaultServerVersion;
   const serverProperties: string[] = workspace
     .getConfiguration("metals")
-    .get<string>("serverProperties")!
-    .split(" ")
-    .filter(e => e.length > 0);
```
gabro: I've verified that the UI won't let you input a blank string (not even just spaces), so this I've removed this check.

Same thing below

tgodzik:
@gabro I think we need to add some kind of migration - if user had a setting as string, we need to replace it with a lsit

gabro:
@tgodzik you are right, I'll work on a fix for this soon.

olafurpg:
This is awesome :D

PR #142, opened 2019-09-06
Migrate old settings from space-separated values to arrays

gabro:
This is a followup of #141, which migrates the old settings values to the new format

Example:


gabro:
@olafurpg yes, both of them are client-side only

gabro:
@olafurpg see also the relevant discussion happening here, which would make `customRepositories` a server setting https://github.com/scalameta/metals/issues/902

PR #147, opened 2019-10-05
Misc fixes to the extension

gabro:
This PR does a few things, which I've tried to separate commit by commit:

- upgrade dependencies (and remove scaffolded tests). Hopefully this also fixes the GH security warnings
- fix ClientCommands registration. Client commands were not being registered correctly, and this resulted in almost all commands not working (the ones which did, it was because they were manually registered, which should not be necessary)
- fix 'Unknown command' warning for `metals-reload-doctor` (this was a false positive, due to a faulty logic flow)


src/extension.ts
```diff
@@ -370,43 +368,43 @@ function launchMetals(
...
+          compilationDoneEmitter.fire();
+          break;
+        default:
+          const isRun = params.command === "metals-doctor-run";
+          const isReload = params.command === "metals-doctor-reload";
+          if (isRun || (doctor && isReload)) {
+            const html = params.arguments && params.arguments[0];
+            if (typeof html === "string") {
+              const panel = getDoctorPanel(isReload);
+              panel.webview.html = html;
```
gabro: the main change is moving this logic after the switch case.

Previously we would match on `metals-doctor-reload`, but then fall back to the "unknown" in case we didn't want to handle the reload.

src/extension.ts
```diff
@@ -346,7 +346,10 @@ function launchMetals(
       }
     };
     Object.entries(clientCommands).forEach(([name, command]) =>
-      registerCommand(name, command)
+      registerCommand(
+        ClientCommands[name as keyof typeof ClientCommands],
```
gabro: this is the fix for ClientCommands: we were using `name` instead of accessing the id.

For example, we were registering the command as `runDoctor` instead of `metals-doctor-run`

src/extension.ts
```diff
@@ -580,7 +578,6 @@ function launchMetals(
       scalaDebugger
         .initialize(outputChannel)
         .forEach(disposable => context.subscriptions.push(disposable));
-      registerCommand(scalaDebugger.startSessionCommand, scalaDebugger.start);
```
gabro: @marek1840 the command is now automatically registered by `ClientCommands` (see the relative fix above). Can you double check this is ok?

gabro: I've tested that I can still use the extension to run a main, as expected.

marek1840: Thanks for testing!

PR #154, opened 2019-10-17
Refine Java Home detection

olafurpg:
First change: respect JAVA_HOME evironment variable, if defined.

The Metals website says that by default we try to use the `$JAVA_HOME`
environment variable while in reality we don't. This commit changes the
Java Home detection to work like it's explained in the documentation.

For people who start VS Code as a GUI application (not launched via
`code` command in the terminal) this change mostly likely won't have
an impact since environment variables that are defined in
`~/.bash_profile` and `~/.zshrc` are not set in GUI processes.

Second change: fallback to any Java 8 or Java 11 installation that's
available on the computer.  Previously, the VS Code extension failed to
launch on computers that only have Java 11 installed. Now, the VS Code
extension uses the most recent Java installation that's available that's
1.8 or higher and 1.11 or lower.

Sadly, the second fallback doesn't support Jabba installed JDKs but we
can maybe improve on that in the future.

Fixes https://github.com/scalameta/metals/issues/762

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -68,11 +68,12 @@ export async function activate(context: ExtensionContext) {
   detectLaunchConfigurationChanges();
   checkServerVersion();
 
-  getJavaHome()
+  getJavaHome(outputChannel)
     .then(javaHome => fetchAndLaunchMetals(context, javaHome))
     .catch(err => {
       const message =
-        "Unable to find Java 8 home. To fix this problem, update the 'Java Home' setting to point to a Java 8 home directory";
+        "Unable to find a Java 8 or Java 11 installation on this computer. " +
+        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory";
       outputChannel.appendLine(message);
       outputChannel.appendLine(err);
       window.showErrorMessage(message, openSettingsAction).then(choice => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -68,11 +68,12 @@ export async function activate(context: ExtensionContext) {
   detectLaunchConfigurationChanges();
   checkServerVersion();
 
-  getJavaHome()
+  getJavaHome(outputChannel)
```
gabro: `outputChannel` seems unused by `getJavaHome`

```suggestion
  getJavaHome()
```

src/getJavaHome.ts
```diff
@@ -1,30 +1,37 @@
 import locateJavaHome from "locate-java-home";
-import { workspace } from "vscode";
+import { workspace, OutputChannel } from "vscode";
```
gabro:
```suggestion
import { workspace } from "vscode";
```

olafurpg:
Thank you for the review, I removed the unused `OutputChannel`. I had kept it around in case it would come in handy again for the future but we can just add it back then 👍

gabro:
Thanks for the fix. I understand the intent, but in general I'm not a fan of passing the output channel around to utility methods, since it risks to make the output "flow" hard to follow when developing. Whenever possible, I prefer for utilities to return some kind of result, and for the main extension methods to stitch them together as outputs.

Anyway, LGTM 👍

PR #159, opened 2019-11-03
Add support for a new "Decoration Protocol"

olafurpg:
This commit enables Metals to support worksheets.
See https://github.com/scalameta/metals/pull/1041

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -24,7 +24,12 @@ import {
   Selection,
   TextDocument as VscodeTextDocument,
   TextEditorRevealType,
-  TextEditor
+  TextEditor,
+  DecorationRangeBehavior,
+  DecorationOptions,
+  Position,
+  TextEditorDecorationType,
+  WorkspaceEdit
 } from "vscode";
 import {
   LanguageClient,
@@ -58,11 +63,21 @@ import { startTreeView } from "./treeview";
 import { MetalsFeatures } from "./MetalsFeatures";
 import { MetalsTreeViewReveal, MetalsTreeViews } from "./tree-view-protocol";
 import * as scalaDebugger from "./scalaDebugger";
+import {
+  DecorationTypeDidChange,
+  DecorationsRangesDidChange
+} from "./decoration-protocol";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
 const openSettingsCommand = "workbench.action.openSettings";
 let treeViews: MetalsTreeViews | undefined;
+let decorationType: TextEditorDecorationType = window.createTextEditorDecorationType(
+  {
+    isWholeLine: true,
+    rangeBehavior: DecorationRangeBehavior.OpenClosed
+  }
+);
 
 export async function activate(context: ExtensionContext) {
   detectLaunchConfigurationChanges();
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -58,11 +63,21 @@ import { startTreeView } from "./treeview";
...
+import {
+  DecorationTypeDidChange,
+  DecorationsRangesDidChange
+} from "./decoration-protocol";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
 const openSettingsCommand = "workbench.action.openSettings";
 let treeViews: MetalsTreeViews | undefined;
+let decorationType: TextEditorDecorationType = window.createTextEditorDecorationType(
```
gabro: Having a single `decorationType` means that the server will need to manage (and mix) multiple decoration sources, is that intentional? I'm fine with it, but I just wanted to check

src/extension.ts
```diff
@@ -584,6 +602,39 @@ function launchMetals(
...
+        decorationType = window.createTextEditorDecorationType(options);
+      });
+      client.onNotification(DecorationsRangesDidChange.type, params => {
+        outputChannel.appendLine(JSON.stringify(params));
+        const uris = window.visibleTextEditors.map(e => e.document.uri);
+        outputChannel.appendLine(JSON.stringify(uris));
+        const editor = window.activeTextEditor;
+        if (editor) {
+          outputChannel.appendLine(
+            JSON.stringify(editor.document.uri.toString())
```
gabro: why `JSON.stringify`?

src/extension.ts
```diff
@@ -584,6 +602,39 @@ function launchMetals(
...
+                new Position(o.range.start.line, o.range.start.character),
+                new Position(o.range.end.line, o.range.end.character)
+              ),
+              hoverMessage: o.hoverMessage,
+              renderOptions: o.renderOptions
+            };
+          });
+          editor.setDecorations(decorationType, options);
+        } else {
+          outputChannel.appendLine("no editor");
```
gabro: This will print `no editor` also in case `params.uri !== editor.document.uri.toString()`, is that intentional?

src/extension.ts
```diff
@@ -584,6 +602,39 @@ function launchMetals(
...
+        outputChannel.appendLine(JSON.stringify(params));
+        const uris = window.visibleTextEditors.map(e => e.document.uri);
+        outputChannel.appendLine(JSON.stringify(uris));
+        const editor = window.activeTextEditor;
+        if (editor) {
+          outputChannel.appendLine(
+            JSON.stringify(editor.document.uri.toString())
+          );
+        }
+        if (editor && params.uri == editor.document.uri.toString()) {
```
```suggestion
        if (editor && params.uri === editor.document.uri.toString()) {
```

I've never found a valid use-case for `==` in JS

src/extension.ts
olafurpg: So far, the Metals server doesn't update the `decorationType` but we have an endpoint to change it just in case that will be handy for the future. I wanted to minimize the need for updating the VS Code client once we want to experiment with using larger parts of the decoration API. 

If we end up needing to use multiple decoration types then we could maybe give them somehow names. For now, it's enough to support only a single default decoration type.

src/extension.ts
olafurpg: Leftover from debugging, removed.

src/extension.ts
olafurpg: Thanks, fixed.

src/extension.ts
olafurpg: Good point, I updated the message to say that decorations were ignored for a "non-active document".

PR #160, opened 2019-11-04
Add automatic JDK installation via Jabba

sswistun-vl:
Previously, the VS Code extension would fail with an error message when it was unable to find a valid Java installation on the computer. Now, the extension asks the user to automatically install Java version via Jabba (https://github.com/shyiko/jabba) when there is no valid Java installation on the computer or if there's only Java 9.

### DIFF
--- a/package.json
+++ b/package.json
@@ -243,14 +243,17 @@
     "vscode:publish": "vsce publish --yarn"
   },
   "devDependencies": {
+    "@types/mkdirp": "^0.5.2",
     "@types/node": "^8.10.25",
     "@types/semver": "^5.5.0",
+    "@types/request": "^2.48.3",
     "@types/shell-quote": "^1.6.1",
     "typescript": "^3.2.2",
     "vsce": "^1.54.0"
   },
   "dependencies": {
     "metals-languageclient": "^0.1.6",
+    "locate-java-home": "^1.1.2",
     "promisify-child-process": "3.1.0",
     "semver": "^5.6.0",
     "vscode": "^1.1.34",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -39,7 +39,6 @@ import {
   ExecuteCommandRequest,
   ShutdownRequest,
   ExitNotification,
-  CancellationToken,
   Location,
   TextDocumentPositionParams,
   TextDocument
@@ -66,6 +65,7 @@ import {
   DecorationTypeDidChange,
   DecorationsRangesDidChange
 } from "./decoration-protocol";
+import { JdkInstaller } from "./jdkInstaller";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
@@ -78,6 +78,16 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
   }
 );
 
+const jdk8Jabba = "adopt@1.8";
+const jdk8Action = "JDK8";
+
+const jdk11Jabba = "adopt@1.11";
+const jdk11Action = "JDK11";
+
+const defaultJabbaVer = "0.11.2";
+
+const config = workspace.getConfiguration("metals");
+
 export async function activate(context: ExtensionContext) {
   detectLaunchConfigurationChanges();
   checkServerVersion();
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
     .catch(err => {
       const message =
         "Unable to find a Java 8 or Java 11 installation on this computer. " +
-        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory";
+        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
+        "or select a version to install automatically (window will need to be reloaded afterwards)";
+
       outputChannel.appendLine(message);
       outputChannel.appendLine(err);
-      window.showErrorMessage(message, openSettingsAction).then(choice => {
-        if (choice === openSettingsAction) {
-          commands.executeCommand("workbench.action.openSettings");
-        }
-      });
+
+      const jabba_ver = defaultJabbaVer;
+      showJabbaMessage(message, jabba_ver);
     });
   commands.executeCommand("setContext", "metals:enabled", true);
+}
+
+function showJabbaMessage(
+  message: string,
+  jabba_ver: string = defaultJabbaVer
+) {
+  return window
+    .showErrorMessage(message, openSettingsAction, jdk8Action, jdk11Action)
+    .then(choice => {
+      const jdkinstaller = new JdkInstaller(outputChannel);
+      switch (choice) {
+        case openSettingsAction: {
+          commands.executeCommand("workbench.action.openSettings");
+          break;
+        }
+        case jdk8Action: {
+          window.withProgress({
+            location: ProgressLocation.Notification,
+            title: `Installing JDK 8, please wait...`,
+            cancellable: true
+          }, (progress, _) => {
+            return jdkinstaller.installJava(jdk8Jabba, jabba_ver).then(updateJavaConfig);
+          })
+          break;
+        }
+        case jdk11Action: {
+          window.withProgress({
+            location: ProgressLocation.Notification,
+            title: `Installing JDK 11, please wait...`,
+            cancellable: true
+          }, (progress, _) => {
+            return jdkinstaller.installJava(jdk11Jabba, jabba_ver).then(updateJavaConfig);
+          })
+          break;
+        }
+      }
+    });
 }
 
 function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
@@ -110,7 +157,15 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   if (dottyArtifact && fs.existsSync(dottyArtifact)) {
     outputChannel.appendLine(
       `Metals will not start since Dotty is enabled for this workspace. ` +
-        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+      `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+    );
+    return;
+  }
+
+  if (!javaHome || javaHome.trim().length == 0 || !fs.existsSync(javaHome)) {
+    showJabbaMessage(
+      "Incorrect 'Java Home' setting. To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
+      "or select a version to install automatically (window will need to be reloaded afterwards)."
     );
     return;
   }
@@ -119,7 +174,6 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   const javaPath = path.join(javaHome, "bin", "java");
   const coursierPath = path.join(context.extensionPath, "./coursier");
 
-  const config = workspace.getConfiguration("metals");
   const serverVersionConfig: string = config.get<string>("serverVersion")!;
   const defaultServerVersion = config.inspect<string>("serverVersion")!
     .defaultValue!;
@@ -226,6 +280,10 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
       });
     }
   );
+}
+
+function updateJavaConfig(javaHome: string, global: boolean = true) {
+  config.update("javaHome", javaHome, global);
 }
 
 function launchMetals(
--- a/src/jdkInstaller.ts
+++ b/src/jdkInstaller.ts
@@ -1,0 +1,97 @@
+import * as path from "path";
+import * as request from "request";
+import * as fs from "fs";
+import * as mkdirp from "mkdirp";
+import * as pcp from "promisify-child-process";
+import { OutputChannel } from "vscode";
+
+export class JdkInstaller {
+  out: OutputChannel;
+
+  homedir = require("os").homedir();
+  bin = path.join(this.homedir, "bin");
+
+  constructor(out: OutputChannel) {
+    this.out = out;
+  }
+
+  installJava(javaVersion: string, jabbaVersion: string) {
+    const jabbaUrl = `https://github.com/shyiko/jabba/releases/download/${jabbaVersion}/jabba-${jabbaVersion}-${this.jabbaUrlSuffix()}`;
+    const jabba = path.join(this.bin, this.jabbaName());
+    return new Promise<void>((res, rej) => {
+      mkdirp(this.bin, (err, made) => {
+        if (err) {
+          const msg = `${err} ${made}`;
+          console.log(msg);
+          this.out.appendLine(msg);
+          rej(err);
+        } else res();
+      });
+    }).then(() =>
+      this.curl(jabbaUrl, jabba, true)
+        .then(pth => pcp.exec(`${jabba} ls-remote`))
+        .then(out =>
+          this.outputToString(out.stdout)
+            .split("\n")
+            .filter((str, i, arr) => str.includes(javaVersion))[0]
+            .trim()
+        )
+        .then(java => {
+          this.out.appendLine(`Installing ${java}`);
+          const jabbaSpawn = pcp.spawn(`${jabba}`, ["install", java], {});
+          jabbaSpawn.stdout.on("data", this.out.append);
+          jabbaSpawn.stderr.on("data", this.out.append);
+
+          return jabbaSpawn
+            .then(() => this.out.appendLine(`${java} installed`))
+            .then(() => pcp.exec(`${jabba} which --home ${java}`))
+            .then((e: pcp.Output) => this.outputToString(e.stdout).trim());
+        })
+    );
+  }
+
+  private jabbaUrlSuffix(): string {
+    const runnerOs = process.platform;
+    switch (runnerOs.toLowerCase()) {
+      case "linux":
+        return "linux-amd64";
+      case "darwin":
+        return "darwin-amd64";
+      case "win32":
+        return "windows-amd64.exe";
+      default:
+        throw new Error(
+          `unknown runner OS: ${runnerOs}, expected one of Linux, macOS or Windows.`
+        );
+    }
+  }
+
+  private isWindows(): boolean {
+    return process.platform === "win32";
+  }
+
+  private jabbaName(): string {
+    if (this.isWindows()) return "jabba.exe";
+    else return "jabba";
+  }
+
+  private curl(url: string, outputFile: string, exec?: boolean) {
+    return new Promise<string>((res, rej) => {
+      request
+        .get(url)
+        .pipe(fs.createWriteStream(outputFile))
+        .on("close", () => {
+          if (exec) fs.chmodSync(outputFile, 755);
+          res(outputFile);
+        })
+        .on("error", err => rej(err));
+    });
+  }
+
+  private outputToString(
+    out: Buffer | string | null | undefined,
+    enc: string = "utf-8"
+  ) {
+    return out instanceof Buffer ? out.toString(enc) : out ? <string>out : "";
+  }
+}
--- a/yarn.lock
+++ b/yarn.lock
@@ -9,6 +9,18 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
+"@types/caseless@*":
+  version "0.12.2"
+  resolved "https://registry.yarnpkg.com/@types/caseless/-/caseless-0.12.2.tgz#f65d3d6389e01eeb458bd54dc8f52b95a9463bc8"
+  integrity sha512-6ckxMjBBD8URvjB6J3NcnuAn5Pkl7t3TizAg+xdlzzQGSPSmBcXf8KoIH0ua/i+tio+ZRUHEXp0HEmvaR4kt0w==
+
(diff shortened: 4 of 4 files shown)


src/extension.ts
```diff
@@ -96,11 +122,14 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
     outputChannel.appendLine(
       `Metals will not start since Dotty is enabled for this workspace. ` +
-        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+      `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
     );
     return;
   }
 
+  if (!javaHome || javaHome.length == 0 || !fs.existsSync(javaHome))
+    throw new Error("Invalid Java Home path");
```
olafurpg: Should we report an error message to the user instead of throwing an exception?

src/jdkInstaller.ts
```diff
@@ -0,0 +1,73 @@
...
+            }
+            else res();
+        });
+    }).then(() => curl(jabbaUrl, jabba, true)
+        .then(pth => pcp.exec(`${jabba} ls-remote`))
+        .then(out => outputToString(out.stdout).split("\n")
+            .filter((str, i, arr) => str.includes(javaVersion))[0].trim())
+        .then(async java => {
+            if (out) out(`Installing ${java}`);
+            await pcp.exec(`${jabba} install ${java}`);
```
olafurpg: Do we display any progress bar to the user while this download happens? It would be great to stream stdout to the Metals `OutputChannel` and make sure that the output channel is visible during the download.

src/extension.ts
```diff
@@ -96,11 +122,14 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   if (dottyArtifact && fs.existsSync(dottyArtifact)) {
     outputChannel.appendLine(
       `Metals will not start since Dotty is enabled for this workspace. ` +
-        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+      `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
```
olafurpg: Was this change intentional? Make sure to use Prettier for formatting the files

src/extension.ts
```diff
@@ -58,12 +58,21 @@ import { startTreeView } from "./treeview";
 import { MetalsFeatures } from "./MetalsFeatures";
 import { MetalsTreeViewReveal, MetalsTreeViews } from "./tree-view-protocol";
 import * as scalaDebugger from "./scalaDebugger";
+import * as jdki from "./jdkInstaller";
```
olafurpg: Can we try to use consistent names between the file and imported variable?
```suggestion
import * as jdkInstaller from "./jdkInstaller";
```

olafurpg:
I took the liberty to update the PR description and title according to https://github.com/scalameta/metals-vscode/pull/161#issuecomment-550147732 Please try to use this format in the future.

src/extension.ts
sswistun-vl: It gets reported to user when it's caught by catch at line 82

sswistun-vl:
There is one issue that I've noticed - adopt JDK uses different version scheme than others - for example: adopt@1.8.0-232 and amazon-corretto@1.8.232-09.1.

src/jdkInstaller.ts
sswistun-vl: Did you meant something like what I've implemented?

src/extension.ts
sswistun-vl: Fixed.

tgodzik:
@olafurpg what do you think about the change? We also added the Java 9 use case - which should help with https://github.com/scalameta/metals/issues/1037

tgodzik:
@sswistun-vl Let's rebase this PR

sswistun-vl:
Rebased.

src/extension.ts
```diff
@@ -58,29 +55,72 @@ import { startTreeView } from "./treeview";
...
+        .then(choice => {
+          const jdkinstaller = new JdkInstaller(outputChannel);
+          switch (choice) {
+            case openSettingsAction: {
+              commands.executeCommand("workbench.action.openSettings");
+              break;
+            }
+            case jdk8Action: {
+              window.showInformationMessage(
+                `Installing ${jdk8}, please wait...`
```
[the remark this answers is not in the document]
sswistun-vl: It's handled by Jabba.

src/jdkInstaller.ts
```diff
@@ -0,0 +1,97 @@
...
+  isWindows(): boolean {
+    return process.platform === "win32";
+  }
+
+  jabbaName(): string {
+    if (this.isWindows()) return "jabba.exe";
+    else return "jabba";
+  }
+
+  installJava(javaVersion: string, jabbaVersion: string) {
```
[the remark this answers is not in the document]
sswistun-vl: Done.

package.json
```diff
@@ -98,6 +98,11 @@
           "type": "string",
           "markdownDescription": "Optional path to the Java home directory. Requires reloading the window.\n\nDefaults to the most recent Java 8 version computed by the `locate-java-home` npm package."
         },
+        "metals.jabbaVersion": {
```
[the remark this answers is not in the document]
sswistun-vl: Removed.

src/extension.ts
```diff
@@ -58,29 +55,72 @@ import { startTreeView } from "./treeview";
...
   checkServerVersion();
 
   getJavaHome()
     .then(javaHome => fetchAndLaunchMetals(context, javaHome))
     .catch(err => {
-      const message =
-        "Unable to find a Java 8 or Java 11 installation on this computer. " +
-        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory";
+      var message = "";
+      if (err.message === "Java9") {
```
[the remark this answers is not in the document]
sswistun-vl: I've extracted "Java9" to variable. This check is used to determine if problem is with Java version or lack of it.

src/extension.ts
```diff
@@ -101,11 +141,13 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
     return;
   }
 
+  if (!javaHome || javaHome.trim().length == 0 || !fs.existsSync(javaHome))
+    throw new Error("Invalid JavaHome");
```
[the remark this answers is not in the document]
sswistun-vl: This exception is handled at line 119. User gets error message when it's thrown.

src/extension.ts
```diff
@@ -79,23 +79,68 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
   checkServerVersion();
 
   getJavaHome()
     .then(javaHome => fetchAndLaunchMetals(context, javaHome))
     .catch(err => {
-      const message =
-        "Unable to find a Java 8 or Java 11 installation on this computer. " +
-        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory";
+      var message = "";
+      if (err.message === java9) {
```
olafurpg: nit: can we extract this if/else chain into a method so we can do `const message = getMessage(err) + "To fix this problem ..."`

src/extension.ts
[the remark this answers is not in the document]
olafurpg: Can we use something more robust than the stringly typed logic `err.message.includes("JavaHome")`?

src/extension.ts
```diff
@@ -228,6 +275,11 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   );
 }
 
+function updateJavaConfig(javaHome: string, global: boolean = true) {
+  workspace.getConfiguration("metals").update("javaHome", javaHome, global);
+  commands.executeCommand("workbench.action.reloadWindow");
```
olafurpg: Let's not reload automatically, Metals will already prompt the user to reload the window when settings change. Some users may have state in their terminal that they don't want to loose

src/extension.ts
```diff
@@ -841,22 +893,18 @@ function migrateStringSettingToArray(id: string): void {
...
 
   if (typeof setting.globalValue === "string") {
-    workspace
-      .getConfiguration("metals")
-      .update(
-        id,
-        setting.globalValue.split(" ").filter(e => e.length > 0),
-        ConfigurationTarget.Global
-      );
+    workspace.getConfiguration("metals").update(
```
olafurpg: Is this formatting change caused by a Prettier version mis-match? @gabro is there a way to enforce a stable version of Prettier in CI?

gabro: Yes, it's sufficient to `yarn add --dev prettier` and then run `yarn prettier --check` on the CI.

src/extension.ts
gabro: please note that there's a generalized way to do this in response to configuration changes. See https://github.com/scalameta/metals-vscode/blob/136ef625062b690e9ac22a9464c8ac8d281a7d1a/src/extension.ts#L729-L754

gabro: (actually, `javaHome` is already in there, so the user should be prompted automatically already after the `update`)

src/extension.ts
```diff
@@ -79,23 +79,68 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
   }
 );
 
+const jdk8 = "1.8.232";
+const jdk8Action = "JDK8";
+
+const jdk11 = "1.11";
+const jdk11Action = "JDK11";
+
+const default_jabba_ver = "0.11.2";
```
gabro: nit: please use camel case for identifier names.

```suggestion
const defaultJabbaVer = "0.11.2";
```

src/extension.ts
gabro: I agree, this would also allow avoiding mutation.

By the way, @sswistun-vl even if we need a mutable variable, I recommend using `let` instead of `var`, since `let` has way more intuitive and sensible scoping rules (`let` is block-scoped whereas `var` is function-scoped)

src/extension.ts
```diff
@@ -79,23 +79,68 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
+
+const config = workspace.getConfiguration("metals");
+
+const java9 = "Java9";
```
gabro: Suggestion: I would name this constant something more telling, such as `unsupportedJavaVersion9` or something like that.

Also, I would use the same constant where you raise the error, so that if you change it in the future you only have one source of truth for the underlying string.

gabro: as an aside, did I mention that I really really dislike the `Promise` API because it doesn't carry the error type? Nothing related to this PR, feel free to ignore my rants :D

src/extension.ts
sswistun-vl: Changed, good to know about var and let difference, I don't know too much about js/ts.

src/extension.ts
sswistun-vl: I've removed auto-reload.

src/extension.ts
sswistun-vl: Changed.

src/extension.ts
sswistun-vl: Renamed. I understand why you dislike it, but on the other hand it makes chaining futures much clearer to write and read :D

gabro: Oh, I'm all for the chaining API, I'm just bothered by the specific implementation :P 

I may propose a rewrite using `TaskEither` from fp-ts in the future ;)

src/extension.ts
tgodzik: Added an issue for this.

src/extension.ts
```diff
@@ -87,17 +97,58 @@ export async function activate(context: ExtensionContext) {
...
+    .then(choice => {
+      const jdkinstaller = new JdkInstaller(outputChannel);
+      switch (choice) {
+        case openSettingsAction: {
           commands.executeCommand("workbench.action.openSettings");
+          break;
         }
-      });
+        case jdk8Action: {
+          window.showInformationMessage(`Installing ${jdk8}, please wait...`);
```
tgodzik: Can we use `window.withProgress` here? Tested it out and it looks really cool, the only issue I have is the window just shows up and it's hard to say what is going on, it would be nice to have:
1. `Installing ${jdk8}, please wait...` showing progress (or just 100% when it's done)
2. Then the window disappears and reload suggestion appears.

tgodzik: Ok, that's fixed and it shows progress nicely

package.json
```diff
@@ -251,6 +252,8 @@
   },
   "dependencies": {
     "metals-languageclient": "^0.1.6",
+    "@types/request": "^2.48.3",
```
gabro: this should be a devDependency

src/extension.ts
```diff
@@ -78,6 +78,16 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
   }
 );
 
+const jdk8 = "1.8.232";
+const jdk8Action = "JDK8";
+
+const jdk11 = "1.11";
```
gabro: are you sure about this? We recently fixed a bug due to Java 11 being on version “11” rather than “1.11”

tgodzik: It's coming from https://github.com/shyiko/jabba/blob/master/index.json

Renamed it to `jdk11Jabba` to pronounce it better.

package.json
tgodzik: Moved it, thanks!

src/extension.ts
```diff
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
     .catch(err => {
       const message =
         "Unable to find a Java 8 or Java 11 installation on this computer. " +
-        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory";
+        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
+        "or select version to install automatically (window will need to be reloaded afterwards)";
```
ckipp01:
```suggestion
        "or select a version to install automatically (window will need to be reloaded afterwards)";
```

src/extension.ts
```diff
@@ -110,7 +157,15 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
-        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+      `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+    );
+    return;
+  }
+
+  if (!javaHome || javaHome.trim().length == 0 || !fs.existsSync(javaHome)) {
+    showJabbaMessage(
+      "Incorrect 'Java Home' setting. To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
+      "or select version to install automatically (window will need to be reloaded afterwards)."
```
ckipp01:
```suggestion
      "or select a version to install automatically (window will need to be reloaded afterwards)."
```

tgodzik: Thanks! Fixed both!

tgodzik:
@gabro Are you ok with merging this?

src/extension.ts
```diff
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
     .catch(err => {
       const message =
         "Unable to find a Java 8 or Java 11 installation on this computer. " +
-        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory";
+        "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
+        "or select a version to install automatically (window will need to be reloaded afterwards)";
```
gabro: nit: instead of writing it here, can we automatically prompt the user to reload the window after installation?

src/extension.ts
```diff
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
...
+
       outputChannel.appendLine(message);
       outputChannel.appendLine(err);
-      window.showErrorMessage(message, openSettingsAction).then(choice => {
-        if (choice === openSettingsAction) {
+
+      const jabba_ver = defaultJabbaVer;
```
gabro: nit: 

```suggestion
      const jabbaVersion = defaultJabbaVer;
```

gabro: I don't see the point of this const, since it's only used in the next call. I would remove it.

src/extension.ts
```diff
@@ -78,6 +78,16 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
   }
 );
 
+const jdk8Jabba = "adopt@1.8";
+const jdk8Action = "JDK8";
+
+const jdk11Jabba = "adopt@1.11";
+const jdk11Action = "JDK11";
+
+const defaultJabbaVer = "0.11.2";
```
gabro: nit:

```suggestion
const defaultJabbaVersion = "0.11.2";
```

src/extension.ts
```diff
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
...
+
+      const jabba_ver = defaultJabbaVer;
+      showJabbaMessage(message, jabba_ver);
+    });
+  commands.executeCommand("setContext", "metals:enabled", true);
+}
+
+function showJabbaMessage(
+  message: string,
+  jabba_ver: string = defaultJabbaVer
```
gabro: nit: 
```suggestion
  jabbaVersion: string = defaultJabbaVersion
```

src/extension.ts
```diff
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
...
           commands.executeCommand("workbench.action.openSettings");
+          break;
         }
-      });
+        case jdk8Action: {
+          window.withProgress({
+            location: ProgressLocation.Notification,
+            title: `Installing JDK 8, please wait...`,
+            cancellable: true
+          }, (progress, _) => {
```
gabro: nit: `progress` is unused

```suggestion
          }, () => {
```

src/extension.ts
```diff
@@ -87,16 +97,53 @@ export async function activate(context: ExtensionContext) {
...
+            return jdkinstaller.installJava(jdk8Jabba, jabba_ver).then(updateJavaConfig);
+          })
+          break;
+        }
+        case jdk11Action: {
+          window.withProgress({
+            location: ProgressLocation.Notification,
+            title: `Installing JDK 11, please wait...`,
+            cancellable: true
+          }, (progress, _) => {
```
gabro: nit: `progress` is unused

```suggestion
          }, () => {
```

src/extension.ts
```diff
@@ -110,7 +157,15 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
-        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+      `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+    );
+    return;
+  }
+
+  if (!javaHome || javaHome.trim().length == 0 || !fs.existsSync(javaHome)) {
+    showJabbaMessage(
+      "Incorrect 'Java Home' setting. To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
+      "or select a version to install automatically (window will need to be reloaded afterwards)."
```
gabro: nit: same comment as above

src/extension.ts
```diff
@@ -228,6 +282,10 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   );
 }
 
+function updateJavaConfig(javaHome: string, global: boolean = true) {
```
gabro: instead of a boolean, I would prefer this to use `ConfigurationTarget` instead of `boolean`.

src/jdkInstaller.ts
```diff
@@ -0,0 +1,97 @@
+import * as path from "path";
+import * as request from "request";
+import * as fs from "fs";
+import * as mkdirp from "mkdirp";
+import * as pcp from "promisify-child-process";
+import { OutputChannel } from "vscode";
+
+export class JdkInstaller {
```
gabro: I would prefer to refactor this not to be a class.

The only purpose of this class is to close over `out` which can simply be a parameter of `installJava` and then we can have a single function. All other class members can be simple values/functions inside this module.

I can work on this refactor @tgodizk

src/jdkInstaller.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  constructor(out: OutputChannel) {
+    this.out = out;
+  }
+
+  installJava(javaVersion: string, jabbaVersion: string) {
+    const jabbaUrl = `https://github.com/shyiko/jabba/releases/download/${jabbaVersion}/jabba-${jabbaVersion}-${this.jabbaUrlSuffix()}`;
+    const jabba = path.join(this.bin, this.jabbaName());
+    return new Promise<void>((res, rej) => {
+      mkdirp(this.bin, (err, made) => {
```
gabro: minor, but we can probably use Node's `promisify` method here:


```
promisify(mkdirp)(this.bin)
```

I can work on this as well

tgodzik:
> I find the code a bit non-idiomatic and I've left a few comments around. @tgodzik if you're with it I can rework some pieces a bit and directly commit on this PR, wdyt?

Sure, if you have time that would be awesome! I am still struggling a bit with typescript :sweat_smile:

src/extension.ts
tgodzik: The user will be prompted - it's just to warn them about it

gabro:
Superseded by #186

PR #161, opened 2019-11-04
Upgrade Coursier-bootstrap

### LINKED ISSUE #126: Coursier bootstrap downloads from maven central without using custom repositories
**Unable to download coursier-cli_2.12-1.1.0-M13-2-standalone.jar**

1. Go to 'install metals for vs code'
 ```Java home: C:\Program Files (x86)\Java\jdk1.8.0_201```
```Error while downloading https://repo1.maven.org/maven2/io/get-coursier/coursier-cli_2.12/1.1.0-M13-2/coursier-cli_2.12-1.1.0-M13-2-standalone.jar: Connection timed out: connect, ignoring it```

**Expected behavior**
  Should allow to download this coursier.

**Installation:**

- Operating system: macOS/Windows
- VSCode version: 1.36.1
- VSCode extension version:  
- Metals version: (found in VSCode settings, under `metals.serverVersion`)  : 1.5.1

**Additional context**
I tried to install the metals in visual studio code but immediately got the above error. It is trying to download `coursier` from `repo1.maven.org` . But we have separate repository to download this part. not able to setup custom repository to download `coursier`. Also i did the proxy setting but it did not help. So all i want to update the repo to my custom repo. Also tried to do custom repo config option but that also did not help. 

here is the settings i am using for custom repository which i did by adding environment variables : 
COURSIER_REPOSITORIES="ivy2Local|central|sonatype:releases|jitpack|/<my repo url>"

**Search terms**
Unable to download coursier-cli_2.12-1.1.0-M13-2-standalone.jar

### PR DESCRIPTION
**Pull request by sswistun-vl**
Fixes #126. Previously, used version didn't support mirrors. Current does, which makes it possible to download metals from custom repository.

### CODE REVIEW

File: src/extension.ts [hunk 1]
```diff
@@ -96,16 +99,35 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
+      window.showErrorMessage(msg, openSettingsAction).then(resp => {
+        if (resp === openSettingsAction) {
+          commands.executeCommand(openSettingsCommand);
+        }
+      });
+    }
+  }
+  else {
+    coursierPath = path.join(context.extensionPath, "./coursier");
+  }
```
gabro: Can you move this logic to a separate utility function and only leave

```
const coursierPath = getCoursierPath()
```

here?

package.json
```diff
@@ -82,6 +82,10 @@
     "configuration": {
       "title": "Metals",
       "properties": {
+        "metals.coursierLocation": {
```
gabro:
```suggestion
        "metals.coursierPath": {
```

sounds a bit better.

Also, can you add a `markdownDescription` field? This is the documentation that is displayed in the settings UI.

package.json
```diff
@@ -240,6 +244,7 @@
     "vsce": "^1.54.0"
   },
   "dependencies": {
+    "@types/mkdirp": "^0.5.2",
```
gabro: why was this needed?

src/extension.ts
```diff
@@ -174,15 +205,28 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
         javaPath,
         classpath,
         serverProperties,
-        javaOptions
+        javaOptions,
+        customRepositories
       );
     },
-    () => {
+    (reason: any) => {
```
gabro: `any` effectively turns off type-checking, so I would prefer never using this.

Since `trackDownloadProgress` returns a `ChildProcessPromise` we can assume this has type `ExitReason` (from `promisify-child-process`)

```suggestion
    (reason: ExitReason) => {
```

<rant>I really don't like Promises because the error type is not preserved</rant>

src/extension.ts
```diff
@@ -174,15 +205,28 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
       const msg = (() => {
         const proxy =
           `See https://scalameta.org/metals/docs/editors/vscode.html#http-proxy for instructions ` +
           `if you are using an HTTP proxy.`;
-        if (process.env.FLATPAK_SANDBOX_DIR) {
+        if (reason.code == 1) {
```
```suggestion
        if (reason.code === 1) {
```

Minor, but I always use `===` unless there's a good reason not to.

src/extension.ts
```diff
@@ -602,6 +647,7 @@ function trackDownloadProgress(
     if (!msg.startsWith("Downloading")) {
       output.appendLine(msg);
     }
+    console.log(msg);
```
gabro: leftover?

src/extension.ts
```diff
@@ -219,7 +263,8 @@ function launchMetals(
   javaPath: string,
   metalsClasspath: string,
   serverProperties: string[],
-  javaOptions: string[]
+  javaOptions: string[],
+  env?: string
```
gabro: unused?

src/extension.ts
```diff
@@ -136,6 +158,13 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
       ? {}
       : { COURSIER_REPOSITORIES: customRepositories };
 
+  const customRepositoriesParams = new Array<string>();
+  // config
```
tgodzik: I don't think this should be commented out and the same below.

sswistun-vl:
Ok, now it should be fine, sorry for rushed pr.

src/extension.ts
sswistun-vl: If you've meant env, then it's used now :)

olafurpg:
@sswistun-vl I took the liberty to update the PR title and description, please try to use the following format in the future

* present tense instead of past tense: "Add option for custom Coursier binary" instead of "Added option for custom Coursier binary"
* description should explain the state before the PR and how that changes with this PR. It makes it easier to code review a change when the PR description explains the impact of this change on the user experience

src/extension.ts
```diff
@@ -96,16 +99,18 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
   if (dottyArtifact && fs.existsSync(dottyArtifact)) {
     outputChannel.appendLine(
       `Metals will not start since Dotty is enabled for this workspace. ` +
-        `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
+      `To enable Metals, remove the file ${dottyArtifact} and run 'Reload window'`
```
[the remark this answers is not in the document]
sswistun-vl: Not sure to be honest - I use the default one in VS Code.

src/extension.ts
```diff
@@ -133,7 +138,7 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
 
   const customRepositoriesEnv =
     customRepositories.length == 0
-      ? {}
+      ? { COURSIER_REPOSITORIES: "" }
```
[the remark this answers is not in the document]
sswistun-vl: Type-checking was complaining about it

sswistun-vl:
Thanks! I will do better descriptions next time!

src/extension.ts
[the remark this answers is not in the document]
gabro: Cool 👍

src/extension.ts
```diff
@@ -24,7 +24,9 @@ import {
   Selection,
   TextDocument as VscodeTextDocument,
   TextEditorRevealType,
-  TextEditor
+  TextEditor,
+  TextEdit,
+  env
```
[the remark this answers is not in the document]
gabro: I see the conversation is resolves, but the var is still here. Is there a missing commit?

src/extension.ts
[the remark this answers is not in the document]
gabro: Yes, the type of `env` in `launchMetals` needs to be

```diff
-  env?: { COURSIER_REPOSITORIES: string }
+  env: { COURSIER_REPOSITORIES?: string }
```

then you can remove this

```suggestion
      ? {}
```

src/extension.ts
```diff
@@ -213,13 +228,41 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
+
 function launchMetals(
   outputChannel: OutputChannel,
   context: ExtensionContext,
   javaPath: string,
   metalsClasspath: string,
   serverProperties: string[],
-  javaOptions: string[]
+  javaOptions: string[],
+  env?: { COURSIER_REPOSITORIES: string }
```
```suggestion
  env: { COURSIER_REPOSITORIES?: string }
```

src/extension.ts
```diff
@@ -174,15 +179,28 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
+    (reason: ExitReason) => {
       const msg = (() => {
         const proxy =
           `See https://scalameta.org/metals/docs/editors/vscode.html#http-proxy for instructions ` +
           `if you are using an HTTP proxy.`;
-        if (process.env.FLATPAK_SANDBOX_DIR) {
+        if (reason.code === 1) {
+          if (!hasCustomCoursier(config)) {
+            window.showErrorMessage("You'll need to provide external Coursier if you don't have access " +
+              "to external repositories (coursier.jar from provided link).", "Go to page")
```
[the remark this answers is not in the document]
gabro: ☝️ @sswistun-vl have you seen this comment? What do you think about it?

sswistun-vl:
Yes, I'm thinking how to change that. I could add some kind of ping-like solution to check if internet is working - that way, provided that user doesn't use his own Coursier, we can be pretty sure that the problem is related to how Coursier Bootstrap works

gabro:
> Yes, I'm thinking how to change that. I could add some kind of ping-like solution to check if internet is working - that way, provided that user doesn't use his own Coursier, we can be pretty sure that the problem is related to how Coursier Bootstrap works. Tell me what do you think about this approach.

I'm not entirely sure of the benefits. Assuming we're talking about first-time users, I would expect they've just downloaded the extension, so we can expect internet is working.

I would still lean towards a generic hint of checking the Coursier launcher amongst other things (such as the proxy) and point the user to a documentation/troubleshooting page.

By the way, I can't find the original issue that motivated this change (which may provide some context on how often we expect this error to happen). Where was the issue reported?

sswistun-vl:
It's for this issue: https://github.com/scalameta/metals-vscode/issues/126. Coursier Bootstrap used for launching actual Coursier has hardcoded links pointing to Maven, so it fails when Maven is unreachable for any reason. Pinging Google would indicate that there is internet connection and suggest that the problem is caused by bootstrap.

gabro:
> It's for this issue: #126. Coursier Bootstrap used for launching actual Coursier has hardcoded links pointing to Maven, so it fails when Maven is unreachable for any reason.

Thanks! Can you maybe add this to the PR description? If you use the "Fixes #<issueN>" syntax it will automatically close it once we merge the PR.

> Pinging Google would indicate that there is internet connection and suggest that the problem is caused by bootstrap.

Are we 100% the problem is caused by the coursier jar if there's internet connection? Could it be caused by other things, such a proxy misconfiguration?

sswistun-vl:
In most cases fat Coursier jar should help with that, at least as long as custom repositories are set.

src/extension.ts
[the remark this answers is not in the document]
sswistun-vl: Fixed.

yarn.lock
```diff
@@ -9,6 +9,13 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
+"@types/mkdirp@^0.5.2":
+  version "0.5.2"
+  resolved "https://registry.yarnpkg.com/@types/mkdirp/-/mkdirp-0.5.2.tgz#503aacfe5cc2703d5484326b1b27efa67a339c1f"
+  integrity sha512-U5icWpv7YnZYGsN4/cmh3WD2onMY0aJIiTE6+51TwJCttdHvtCYmkBNOobHlXwrJRL0nkH9jH4kD+1FAdMN4Tg==
+  dependencies:
+    "@types/node" "*"
```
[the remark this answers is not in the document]
sswistun-vl: Should be fixed right now.

src/extension.ts
sswistun-vl: Fixed.

src/extension.ts
sswistun-vl: Removed.

package.json
sswistun-vl: Changed.

olafurpg:
@alexarchambault would it be possible for slim coursier bootstrap scripts to respect mirror configuration? The Metals VS Code extension currently embeds a slim `coursier` binary that doesn't work when behind corporate proxies.

alexarchambault:
@olafurpg Yeah, possibly, I've been thinking about it too. I'll try to do that.

tgodzik:
@olafurpg it think it still makes sense to have coursier path as a config option. Do you think we should drop this change?

olafurpg:
@tgodzik I'm not sure what's the best move. I would love to have a single solution for users behind corporate networks. The `coursierPath` setting is quite complicated to explain and use, it assumes the user knows how to build a fat jar of coursier locally and that the user has configured mirrors 🤔

sswistun-vl:
@olafurpg that's why I wanted to keep showing less generic error message - it points you to coursier release on github and instructing which file to download (assuming that github page will be accessible).

PR #165, opened 2019-12-10
Enable setting breakpoints in java and scala files

marek1840:
Previously, metals didn't allow the user to set breakpoints in any file. Now, they can be set in any java or scala file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,6 +214,14 @@
         }
       ]
     },
+    "breakpoints": [
+      {
+        "language": "scala"
+      },
+      {
+        "language": "java"
+      }
+    ],
     "debuggers": [
       {
         "type": "scala",


package.json
```diff
@@ -214,6 +214,14 @@
         }
       ]
     },
+    "breakpoints": [
```
olafurpg: Is there a way to conditionally add this capability based on the server version? If not I guess we'll just have to live with it. The next stable version of Metals should come out soon

gabro: Judging by the docs, I don't think so.

> we use the breakpoints contribution point to list the languages for which setting breakpoints will be enabled. Without this, it would not be possible to set breakpoints

olafurpg: Fair enough

PR #166, opened 2019-12-18
Support StartRunSession command

marek1840:
Previously, there was no way to start a run session - only debug sessions were supported.
Now, StartRunSession (added in https://github.com/scalameta/metals/pull/1200) is also supported.

### DIFF
--- a/src/client-commands.ts
+++ b/src/client-commands.ts
@@ -5,5 +5,6 @@ export const ClientCommands = {
   toggleLogs: "metals-logs-toggle",
   focusDiagnostics: "metals-diagnostics-focus",
   runDoctor: "metals-doctor-run",
-  startDebugSession: "metals-debug-session-start"
+  startDebugSession: "metals-debug-session-start",
+  startRunSession: "metals-run-session-start"
 };
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -353,12 +353,21 @@ function launchMetals(
           channelOpen = true;
         }
       },
-      startDebugSession: args => {
+      startDebugSession: (...args) => {
         if (!features.debuggingProvider) return;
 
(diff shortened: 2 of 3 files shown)


src/client-commands.ts
```diff
@@ -5,5 +5,6 @@ export const ClientCommands = {
   toggleLogs: "metals-logs-toggle",
   focusDiagnostics: "metals-diagnostics-focus",
   runDoctor: "metals-doctor-run",
-  startDebugSession: "metals-debug-session-start"
+  startDebugSession: "metals-debug-session-start",
+  startRunSession: "metals-run-session-start"
```
olafurpg: Is this command only used for testing or also running?

marek1840: for both - a unit (be it class or test) can be either just run or also debugged, hence two commands.

I was thinking about passing additional argument but it would get a bit awkward: there would need to be a separate data structure to hold the `debug: Boolean` field, we would need to extract that on the metals side and separate it from arguments sent to bloop. With this approach we keep it simple: whatever argument this we get is forwarded to bloop as is, without any introspection.

olafurpg: Should we rename the "run" part into something else to make it clear it's just "debugging without breakpoints"? For example `metals-debug-no-breakpoints-session-start` 🤔

marek1840: We could, but it would no longer correspond to `run | debug` code lenses or to IntelliJ Run or Debug actions. That was my reasoning - to make it vaguely familiar to the user.

Although, I agree that debugging is also running therefore the name could be better.

gabro: FWIW, I find "run" and "debug" quite intuitive in their respective meaning of "no breakpoints" / "with breakpoints"

PR #172, opened 2020-01-14
update table in readme

ckipp01:


readme.md
```diff
@@ -17,11 +16,13 @@ following table shows the status of various features.
...
+| Find implementations  |   ✅   |                                                                                                  |
 | Workspace symbol      |   ✅   | Searches workspace sources and library dependencies. All-lowercase queries are case-insensitive. |
 | Highlight             |   ✅   |                                                                                                  |
 | Folding               |   ✅   |                                                                                                  |
-| Rename symbol         |   ❌   |                                                                                                  |
-| Code actions          |   ❌   |                                                                                                  |
+| Rename symbol         |   ✅   |                                                                                                  |
+| Code actions          |   ⚠️    | Limited code actions are currently available                                                     |
```
gabro: I would remove this entry. "Code actions" is extremely generic and it's not very useful to have it here.

tgodzik:
@ckipp01 This should actually be first updated in Metals -  the contents are copied from vscode section on the website.

gabro:
Ah right. We should really automate that process at some point

ckipp01:
> @ckipp01 This should actually be first updated in Metals - the contents are copied from vscode section on the website.

But the VS Code section of the website doesn't have this table. There is a table in the [Overview[(https://scalameta.org/metals/docs/editors/overview.html), however that one is updated.

tgodzik:
> > @ckipp01 This should actually be first updated in Metals - the contents are copied from vscode section on the website.
> 
> But the VS Code section of the website doesn't have this table. There is a table in the [Overview](https://scalameta.org/metals/docs/editors/overview.html), however that one is updated.

You're right, that table is actually the only thing not copied. We should update the table in the editors overview too though.

olafurpg:
iirc the release instructions contain a manual step to update the vscode readme. It might be nice to automate this to reduce the overhead of cutting a release but I don’t expect we’ll have many more releases that add so many new features

PR #174, opened 2020-01-14
Pass user's environment to metals server process

jodersky:
This ensures that a user's environment variables are always passed to the metals server, as used to be the case before the environment was overridden in #161.

Notably, this PR fixes scalameta/metals-vscode#175

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -236,7 +236,7 @@ function launchMetals(
   metalsClasspath: string,
   serverProperties: string[],
   javaOptions: string[],
-  env: { COURSIER_REPOSITORIES?: string }
+  extraEnv: { COURSIER_REPOSITORIES?: string }
 ) {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -254,9 +254,11 @@ function launchMetals(
     .concat(serverProperties)
     .concat(mainArgs);
 
+  const baseEnv = process.env;
+
```
gabro: Super minor nit, you could do

```ts
const env = { ...process.env, ...extraEnv}
```

And leave the rest unchanged

jodersky:
I refactored the changes to include your suggestion, thanks. I also tested it locally and it does solve scalameta/metals#1298.

gabro:
That's great, thank you for fixing this! I'll merge and start a new release as soon as I'm at a computer!

jodersky:
Thanks! Looking forward to it

PR #180, opened 2020-01-18
Use getJavaHome from metals-languageclient

gabro:
This is a first step towards deduplicating logic between `metals-vscode` and `coc-metals`, by centralizing the shared logic in `metals-languageclient`


olafurpg:
This is awesome! So nice we can share launching functionality across editors :)

PR #182, opened 2020-01-22
Add missing bloop restart command and settings

tgodzik:
Recently there was a new command added to restart the build server plus two new options for working with Bloop.


package.json
```diff
@@ -144,6 +152,11 @@
         "category": "Metals",
         "title": "Restart server"
       },
+      {
+        "command": "metals.build-restart",
```
tgodzik: Not sure about the name so that it's not mixed with `Server restart`. Maybe `Bloop restart`?

olafurpg: `bloop-restart` is probably a better name.

tgodzik: Changed to "Restart Bloop server"

Issue #179, opened 2020-01-16
Add explicit way to create new file (a.k.a. wizard for new class/worksheet)


alekseiAlefirov:
The option to create new scala file or worksheet (or, potentially, notebook) should be added to
- most expected, explorer directory context menu,
and, additionally can be added to
- file menu (not sure if there's such an option in VS Code)
- command palette

**Should there be actual wizard?**
Where wizard is a separate window with controls. It suppose something like IDEa's wizard to choose among Scala class/object/trait.
- Should there be such choice at all?
- Additionally, such choice assume auto creating blank class/object/trait with user-suggested name, which can conflict with package provider, triggered by `didOpen`, editor-independent.
- Not sure, there's a way to create wizards in VS Code (quick googling does not give positive answer on that).

As a first iteration my suggestion is to create a simple explorer context menu addition under `New File`:
```
|- New File
|- New Scala >
|- |- File (or Class, not sure which is more accurate here)
|- |- Worksheet
|-...
```

tgodzik:
@gabro  Sorry for not warning you about it!

olafurpg:
I think this is a good idea. I would avoid making this functionality limited to VS Code. It’s vest to implement most of the logic in the Metals server. We already have metals/inputBox that I believe is also supported in Vim.

olafurpg:
One VS Code api we don’t have an extension for right now is “quick open” (I think it’s called that). I’m not sure we need it for this feature however. We could add a server command “New Scala worksheet”. We might be able to expose this command in the context menu in VS Code but it should also be possible to invoke the command from other editors.

ckipp01:
Closed via #183

PR #183, opened 2020-01-22
Added context menu command to create new worksheet

alekseiAlefirov:
Issue: https://github.com/scalameta/metals-vscode/issues/179
(PR for the server: https://github.com/scalameta/metals/pull/1339)

From the explorer context menu:
new-worksheet-explorer

From Command Palette:
new-worksheet-command-palette

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -522,34 +522,57 @@ function launchMetals(
         return undefined;
       })();
 
-      const worksheetPick = { label: 'Worksheet', kind: 'worksheet' }
-      const classPick = { label: 'Class', kind: 'class' }
-      const objectPick = { label: 'Object', kind: 'object' }
-      const traitPick = { label: 'Trait', kind: 'trait' }
+      const classPick = { label: "Class", kind: "class" };
+      const objectPick = { label: "Object", kind: "object" };
+      const traitPick = { label: "Trait", kind: "trait" };
+      const packageObjectPick = {
+        label: "Package Object",
+        kind: "package-object"
+      };
+      const worksheetPick = { label: "Worksheet", kind: "worksheet" };
 
-      return window.showQuickPick(
-        [classPick, objectPick, traitPick, worksheetPick],
-        { placeHolder: 'Select the kind of file to create' }
-      ).then(kindPick => {
-        if (kindPick !== undefined) return window.showInputBox({ prompt: 'Enter name for the new ' + kindPick.kind }).then(name => {
-          if (name !== undefined) {
-            const arg: MetalsNewScalaFileParams =
-            {
-              'directory': directory,
-              'name': name,
-              'kind': kindPick.kind
+      return window
+        .showQuickPick(
+          [classPick, objectPick, traitPick, packageObjectPick, worksheetPick],
+          { placeHolder: "Select the kind of file to create" }
+        )
+        .then(kindPick => {
+          const name = (async () => {
+            switch (kindPick) {
+              case classPick:
+              case objectPick:
+              case traitPick:
+              case worksheetPick:
+                return window.showInputBox({
+                  prompt: "Enter name for the new " + kindPick.kind
+                });
+              case packageObjectPick:
+                return "";
             }
-            client.sendRequest(ExecuteCommandRequest.type, {
-              command: "new-scala-file",
-              arguments: [arg]
-            }).then(result => {
-              workspace
-                .openTextDocument(Uri.parse(result))
-                .then(textDocument => window.showTextDocument(textDocument));
-            });
-          }
+          })();
+
+          name.then(name => {
+            if (kindPick !== undefined && name !== undefined) {
+              const arg: MetalsNewScalaFileParams = {
+                directory: directory,
+                name: name,
+                kind: kindPick.kind
+              };
+              client
+                .sendRequest(ExecuteCommandRequest.type, {
+                  command: "new-scala-file",
+                  arguments: [arg]
+                })
+                .then(result => {
+                  workspace
+                    .openTextDocument(Uri.parse(result))
+                    .then(textDocument =>
+                      window.showTextDocument(textDocument)
+                    );
+                });
+            }
+          });
         });
-      });
     });
 
     window.onDidChangeActiveTextEditor(editor => {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -41,7 +41,8 @@ import {
   MetalsDidFocus,
   ExecuteClientCommand,
   MetalsInputBox,
-  MetalsWindowStateDidChange
+  MetalsWindowStateDidChange,
+  MetalsNewScalaFileParams
 } from "./protocol";
 import { LazyProgress } from "./lazy-progress";
 import * as fs from "fs";
@@ -509,7 +510,7 @@ function launchMetals(
     });
 
     registerCommand("metals.new-scala-file", (dirArg: Uri) => {
-      const dir = (() => {
+      const directory = (() => {
         if (dirArg !== undefined) {
           return dirArg.toString();
         } else if (window.activeTextEditor) {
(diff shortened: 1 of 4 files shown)


tgodzik:
Looks cool! Thanks @alekseiAlefirov !

I am wondering if we should have a separate TAB with also `new class/object` etc.

alekseiAlefirov:
@olafurpg , can you please elaborate on test file, how it should work?
It most likely can be done, though it seems there's no such option for multi-level context menus in VSCode, so we can do like it's in IDEA. There an issue in VSCode project for that: https://github.com/microsoft/vscode/issues/9827
-- nicely, some people want it for... creating new files from templates :)
We can reconstruct submenu functionality with **quick-pick** API of VSCode, probably, though.

olafurpg:
* you have open file src/main/scala/foo/Foo.scala
* Run "Create test file" command
* it creates src/test/scala/foo/FooSuite.scala

olafurpg:
It's just an idea. It would be good to be able to at least handle a growing number of file kinds

tgodzik:
I agree it would be really good to have a couple of possible file templates, but if it's not yet supported sensibly within vs code then I agree we can start with this one.

Although we could prepare a couple of commands within Metals itself.

olafurpg:
My concern is that the "Create Scala worksheet" command will demand too much attention and this approach isn't scalable to a large number of file kinds. I would prefer instead to expose this feature only as a command and a button in the Metals tree view (maybe under "build" explorer 🤔 )

tgodzik:
> My concern is that the "Create Scala worksheet" command will demand too much attention and this approach isn't scalable to a large number of file kinds. I would prefer instead to expose this feature only as a command and a button in the Metals tree view (maybe under "build" explorer)

I think we need to put it in the explorer - people are already super confused how to create that. I fear a button in the TreeView will too hidden and we will not achieve anything of the original purpose.

We can just have the basic file kinds for now: `.worksheet.sc`, `.sc`, `.sbt`,  `.scala`. It's not perfect, but it's not that large of a menu yet.

olafurpg:
How about having a single "New Scala file" button that opens a quick open dialogue for the available options?

olafurpg:
I agree the button should be available in the context menu, it would not be discoverable in the tree view

alekseiAlefirov:
Context menu changed to support creating new scala class:
NewScalaClass
@olafurpg please check this out.

PR to Metals (server) is to be updated soon.

src/extension.ts
```diff
@@ -519,6 +519,72 @@ function launchMetals(
...
+              [classPick, objectPick, traitPick]
+            ).then(classKindPick => {
+              switch (classKindPick) {
+                case classPick:
+                case objectPick:
+                case traitPick:
+                  return withName(name => {
+                    client.sendRequest(ExecuteCommandRequest.type, {
+                      command: "new-scala-class",
+                      arguments: [dir, name, classKindPick.kind]
```
tgodzik: We should be able to have an object as an argument here and we can reduce it all to a single pick. Worksheet/object/trait/class together.

alekseiAlefirov:
All kinds of file (different classes along with worksheet) now in one menu:
new-scala-file

src/extension.ts
alekseiAlefirov: Done

src/extension.ts
```diff
@@ -499,6 +500,72 @@ function launchMetals(
...
+      const objectPick = { label: "Object", kind: "object" };
+      const traitPick = { label: "Trait", kind: "trait" };
+      const packageObjectPick = {
+        label: "Package Object",
+        kind: "package-object"
+      };
+      const worksheetPick = { label: "Worksheet", kind: "worksheet" };
+
+      return window
+        .showQuickPick(
```
olafurpg: Instead of implementing the quick pick here, I think it would be nice to add a new LSP extension so the Metals server can open a quick pick. We can use window/showMessageRequest as a fallback. 

I have tried to follow the principle of avoiding too much custom logic in the vscode client and allowing the server to push behavior instead. The benefit of this strategy is that we 1) we can reuse more code to support multiple editors and 2) next time we need a quick pick we may not have to update the client code.

src/extension.ts
```diff
@@ -499,6 +500,72 @@ function launchMetals(
...
+      const directory = (() => {
+        if (dirArg !== undefined) {
+          return dirArg.toString();
+        } else if (window.activeTextEditor) {
+          const currentPath = window.activeTextEditor.document.uri.fsPath;
+          return Uri.file(path.dirname(currentPath)).toString();
+        }
+        // NOTE(aleksei): I would return workspace folder here, but not sure if there's API for that.
+        // Moving this funcitonality to server.
+        return undefined;
```
gabro: moving stuff to the server makes sense, but I would change this note to something more "final". 

Also, what API are you looking for? `workspace.workspaceFolders[0]` should give you what you want

src/protocol.ts
```diff
@@ -75,3 +75,9 @@ export namespace MetalsWindowStateDidChange {
 export interface MetalsWindowStateDidChangeParams {
   focused: boolean;
 }
+
+export interface MetalsNewScalaFileParams {
+  directory?: string;
+  name: string;
+  kind: string;
```
gabro: can this kind be more specific?

```suggestion
  kind: 'class' | 'object' | 'trait' | 'package-object' | 'worksheet' ;
```

src/extension.ts
gabro: I have some comments about the implementation, but I too think we should move this to the server similarly to how we do for other extensions.

src/extension.ts
```diff
@@ -499,6 +500,72 @@ function launchMetals(
       });
     });
 
+    registerCommand("metals.new-scala-file", (dirArg: Uri) => {
+      const directory = (() => {
+        if (dirArg !== undefined) {
+          return dirArg.toString();
```
gabro: Is there a more specific method than `toString()`? Like `fsPath` or similar

src/extension.ts
```diff
@@ -499,6 +500,72 @@ function launchMetals(
...
+        } else if (window.activeTextEditor) {
+          const currentPath = window.activeTextEditor.document.uri.fsPath;
+          return Uri.file(path.dirname(currentPath)).toString();
```
gabro: same comment about `toString`

src/extension.ts
alekseiAlefirov: fsPath is not URI (it misses `file:///` part).

alekseiAlefirov:
Thanks for the review, @tgodzik , @olafurpg , @gabro . I'll try to create "show quick pick" method in a PR to metals server and return with that soon.

src/extension.ts
gabro: Ok, reading the docs `toString()` is actually the right method 👍

src/extension.ts
alekseiAlefirov: @olafurpg , so this way part of the logic for the dialog about a new file (it's kind and name) is going to be transferred to the server and _inside_ of the new-scala-file command execution. Dialog parts can be cancelled, so I have a question - how this should be handled by the command? Currently it returns a string/URI of the newly created file. Would using null in case of cancellation be alright for this? Or the request should finish with failure? Or instead of just string as the return type, we can use some `Cancelable` interface / generic container with it inside?

olafurpg: If the command “create new scala file” is sent to the server with a directory as the argument (or something like that) then I think it doesn’t matter what Metals returns. Metals already knows how to do the quick pick, do input box, apply workspace edit and also change the focused window (goto locations, it’s used in tree view protocol). Does that make sense?

olafurpg: The directory argument can be optional, for example when triggered via the command palette. Metals already knows the focused document.

alekseiAlefirov:
`metals/pickInput` added, server side PR here: https://github.com/scalameta/metals/pull/1447

PR #189, opened 2020-02-04
Migrate CI to GitHub Actions

gabro:
This is copied almost verbatim from metals-languageclient, so I expected it to work.

I guess we'll find out at the next release 😅


gabro:
🤔 I'm not sure why it's not running for this PR though. Maybe it doesn't work the first time we push from an external repo?

olafurpg:
I would go ahead and merge and then open a PR to fix the errors.

gabro:
Sounds like my typical CI workflow 👍

PR #190, opened 2020-02-04
Some CI improvements

gabro:
- fix prettier check
- remove travis badge from README
- add publishing of VSIX to the latest release (the idea is that when the release workflow runs it was triggered by a release)


.github/workflows/release.yml
```diff
@@ -22,7 +22,8 @@ jobs:
         id: get_version
         run: echo ::set-output name=NEW_VERSION::${GITHUB_REF#/refs/tags/v}
       - run: yarn install
-      - run: yarn version --no-git-tag-version --new-version ${{
+      - run:
```
olafurpg: This is needed to fix the interaction with prettier
```suggestion
      - run: |
```

.github/workflows/release.yml
```diff
@@ -38,3 +39,19 @@ jobs:
...
+        id: get_upload_url
+        run: |
+          URL=curl --silent "https://api.github.com/repos/scalameta/metals-vscode/releases/latest" | jq -r '.upload_url' | cut -d'{' -f 1
+          echo ::set-output name=UPLOAD_URL::$URL
+      - name: Upload VSIX package to release
+        uses: actions/upload-release-asset@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: { { steps.get_upload_url.outputs.UPLOAD_URL } }
```
olafurpg: Woah, this is cool!

gabro: If it works, sure 😅

.github/workflows/release.yml
gabro: Thanks, fixed

gabro:
@olafurpg what do you think about triggering a release to see whether this works?

olafurpg:
@gabro sounds good! I think we have a few fixes in master that would be good to release as well

PR #207, opened 2020-02-10
Add in clientExperimentalCapabilities from old property settings

ckipp01:
This pr will add the correct `clientExperimentalCapabilities` to the extension as we migrate certain properties away from being properties to being capabilities.

The related PR can be found here: https://github.com/scalameta/metals/pull/1414

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,7 +252,7 @@
     "vsce": "^1.54.0"
   },
   "dependencies": {
-    "metals-languageclient": "0.1.17",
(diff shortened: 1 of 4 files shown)


ckipp01:
After seeing @gabro's comment here https://github.com/scalameta/metals-vscode/pull/192, should we wait to merge this until we have a new release of Metals?

gabro:
I think so

PR #212, opened 2020-02-12
Delegate installJava to metals-languageclient

gabro:
The `installJava` utility now lives in `metals-languageclient`

/cc @ckipp01 in case you changed your mind and want to include this in coc-metals it's fairly easy now.


gabro:
Not much to see here, self-merging

ckipp01:
> /cc @ckipp01 in case you changed your mind and want to include this in coc-metals it's fairly easy now.

I think if it's ok with you, I'd rather hold off on adding it. If anyone requests it or mentions that it'd be cool to have, then I'm more than happy to add it, but I really doubt that it will happen.

PR #214, opened 2020-02-17
Add support for metals.pantsTargets setting

murph:
@olafurpg merged basic support for the Pants build tool for metals in https://github.com/scalameta/metals/pull/935

Part of this support is a new user-configurable setting, metals.pantsTargets.

Here's my attempt at adding that setting to the extension settings UI.

I've never touched a vscode extension before, so more than happy for lots of feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,10 @@
             "type": "string"
           },
           "markdownDescription": "Optional list of custom resolvers passed to Coursier when fetching metals dependencies.\n\nFor documentation on accepted values see the [Coursier documentation](https://get-coursier.io/docs/other-repositories).\n\nThe extension will pass these to Coursier using the COURSIER_REPOSITORIES environment variable after joining the custom repositories with a pipe character (|)."
+        },
+        "metals.pantsTargets": {
+          "type": "string",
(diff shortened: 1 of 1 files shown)


olafurpg:
Thank you for this contribution! The pants integration is still evolving and is unsupported at this time. I think it’s best to wait until it’s stable before adding publicly visible configuration options

murph:
I understand your caution. Maybe if I added some text indicating that it's experimental and might change?

From the perspective of a beginner wanting to try out this support - the hardest part in getting it working was setting this setting.

cosmicexplorer:
> the hardest part in getting it working was setting this setting.

Confirmed -- I failed multiple times to do so in demonstrating the pants support to others.

olafurpg:
It's currently unclear if the `pantsTargets` setting will stay and it's also unclear what minimum Pants version will be required. I'm actively working on some changes that I'm hoping to have ready in the next couple weeks after which I'm optimistic I'll know better how the final integration will look like.

In the meantime, I would prefer to wait with advertising the Pants integration by making it visible in the VS Code configuration UI. Does that make sense?

murph:
I can't say I really agree - I don't see the harm in making it easier to configure an early plugin, especially as it only ever matters to users already using Pants for their build.

However, happy to respect your decision here. Thanks for working on this, it has already made me much more productive.

PR #219, opened 2020-02-19
Specify exactly how server properties should look like

tgodzik:
This is a follow up from https://github.com/scalameta/metals/issues/1362#issuecomment-588137143

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
           "items": {
             "type": "string"
           },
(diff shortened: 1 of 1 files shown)


patricker:
@tgodzik Is this not merged into the release version of the VS Code plugin yet?  I just installed the plugin a few days ago and ran into this same bug.  After finding the original issue I split my proxy settings into separate keys and it worked, as described as the original fix.

tgodzik:
@patricker It's merged: 

image

we could probably try to split the items too.

tgodzik:
Or actually this might be more complex and it's better to keep it as is. With splitting we will not be able to do:
-Dlong.description="this is a long description"

Issue #217, opened 2020-01-15
Support launch.json configuration to run and debug Scala code


ColOfAbRiX:
**Is your feature request related to a problem? Please describe.**
I would like to start Run/Debug of scala not from a Code Lens but from the launch configuration so I can add custom configuration and be more flexible. I also find code lenses not nice and I now need to have them enabled.

**Describe the solution you'd like**
Support a "scala" launch type in launch.json and use it through the debug panel of VSCode

**Describe alternatives you've considered**
I'm not aware of any.

tgodzik:
@alekseiAlefirov is starting to work on this as this is a pretty high priority. Without a configuration people are unable to specify any arguments or properties, so run/debug might be problematic to use for them.

ckipp01:
So I'm not super familiar with DAP stuff, but from looking at it, I think so. My initial thought was that it would be great to reuse some of the logic that @alekseiAlefirov did for the `metals.new-scala-file` and do a similar thing like `metals.launch-config` or something and just offload it to the server. Then let it be in charge of creating it correctly with a base template and passing it to the client. _However_, looking at the link you mentioned in [here](https://github.com/scalameta/coc-metals/issues/80#issuecomment-587581170) it looks like a lot of that stuff is already baked into VS Code, including putting the file in the `.vscode` directory, and having built in completion for whats available etc., so I'm not sure if that would actually be a good idea.

With that all being said, the more we can sync on this the better. I'm pinging @danilbykov as well since it looks like he's taking the charge on this on the `coc-metals` side of stuff.

My naive thoughts are, let's offload anything we can to the server to not have to do all the logic both in the VS Code extension and `coc-metals`. If that's not really possible, then let's make sure we agree on the schema (which I believe is already set in DAP world right?) that we are passing in.

tgodzik:
What needs to be sent to Metals is based on https://github.com/build-server-protocol/build-server-protocol/blob/b4a06bdc649cbb03ddcb457bdf0c76a3e296d376/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaMainClass.java - so that would the schema in a sense.

Although, you are right, some improvements on the server side will be useful to both vim and VS Code, but most likely the launch config itself will not.

ctataryn:
[...]

PR #223, opened 2020-02-25
Support 'launch.json'

alekseiAlefirov:
Issue: https://github.com/scalameta/metals-vscode/issues/217
Server side PR: https://github.com/scalameta/metals/pull/1466

launch_debug

### DIFF
--- a/package.json
+++ b/package.json
@@ -257,7 +257,34 @@
         "label": "Scala Debugger",
         "languages": [
           "scala"
-        ]
+        ],
+        "configurationAttributes": {
+          "launch": {
+            "required": ["mainClass"],
+            "properties": {
+              "mainClass": {
+                "type": "string",
+                "description": "Class to run"
+              },
+              "project": {
+                "type": "string",
+                "description": "Location project of the class to run",
+                "default": "root"
+              }
+            }
+          }
+        },
+        "initialConfigurations": [
+          {
+            "type": "scala",
+            "request": "launch",
+            "name": "Run class",
+            "mainClass": "???"
+          }
+        ],
+        "variables": {
+          "startDebugServer": "metals.start-debug-server"
+        }
       }
     ]
   },
--- a/package.json
+++ b/package.json
@@ -260,7 +260,9 @@
         ],
         "configurationAttributes": {
           "launch": {
-            "required": ["mainClass"],
+            "required": [
+              "mainClass"
+            ],
             "properties": {
               "mainClass": {
                 "type": "string",
@@ -270,6 +272,21 @@
                 "type": "string",
                 "description": "Location project of the class to run",
                 "default": "root"
+              },
+              "args": {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                },
+                "description": "Arguments passed to the class to run",
+                "default": []
+              },
+              "jvmOptions": {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                },
+                "default": []
               }
             }
           }
@@ -278,13 +295,28 @@
           {
             "type": "scala",
             "request": "launch",
-            "name": "Run class",
-            "mainClass": "???"
+            "name": "Untitled",
+            "mainClass": "???",
+            "project": "root",
+            "args": [],
+            "jvmOptions": []
           }
         ],
-        "variables": {
-          "startDebugServer": "metals.start-debug-server"
-        }
+        "configurationSnippets": [
+          {
+            "label": "Run main class",
+            "description": "A new configuration for running a debug session of a main class",
+            "body": {
+              "type": "scala",
+              "request": "launch",
+              "name": "Untitled",
+              "mainClass": "???",
+              "project": "root",
+              "args": [],
+              "jvmOptions": []
+            }
+          }
+        ]
       }
     ]
   },
--- a/package.json
+++ b/package.json
@@ -260,18 +260,18 @@
         ],
         "configurationAttributes": {
           "launch": {
-            "required": [
-              "mainClass"
-            ],
             "properties": {
               "mainClass": {
                 "type": "string",
-                "description": "Class to run"
+                "description": "Name of the main class to run"
               },
-              "project": {
+              "testClass": {
                 "type": "string",
-                "description": "Location project of the class to run",
-                "default": "root"
+                "description": "Name of the test class to run"
+              },
+              "buildTarget": {
+                "type": "string",
+                "description": "Name of the build target of the class to run"
               },
               "args": {
                 "type": "array",
@@ -291,29 +291,27 @@
             }
           }
         },
-        "initialConfigurations": [
-          {
-            "type": "scala",
-            "request": "launch",
-            "name": "Untitled",
-            "mainClass": "???",
-            "project": "root",
-            "args": [],
-            "jvmOptions": []
-          }
-        ],
         "configurationSnippets": [
           {
-            "label": "Run main class",
+            "label": "Scala: Run main class",
             "description": "A new configuration for running a debug session of a main class",
             "body": {
               "type": "scala",
               "request": "launch",
               "name": "Untitled",
               "mainClass": "???",
-              "project": "root",
               "args": [],
               "jvmOptions": []
+            }
+          },
+          {
+            "label": "Scala: Run test class",
+            "description": "A new configuration for running a debug session of a test class",
+            "body": {
+              "type": "scala",
+              "request": "launch",
+              "name": "Untitled",
+              "testClass": "???"
             }
           }
         ]
--- a/src/scalaDebugger.ts
+++ b/src/scalaDebugger.ts
@@ -59,15 +59,15 @@ class ScalaConfigProvider implements vscode.DebugConfigurationProvider {
       })
       .then((result) => {
         if (result === mainClassPick) {
-          return this.provideDebugMainClassConfiguration().then((config) => {
-            if (config === undefined) return [];
-            return [config];
-          });
+          return this.provideDebugMainClassConfiguration().then(
+            (config) => [config],
+            (_) => []
+          );
         } else if (result === testClassPick) {
-          return this.provideDebugTestClassConfiguration().then((config) => {
-            if (config === undefined) return [];
-            return [config];
-          });
+          return this.provideDebugTestClassConfiguration().then(
+            (config) => [config],
+            (_) => []
+          );
         }
         return [];
       });
@@ -81,17 +81,11 @@ class ScalaConfigProvider implements vscode.DebugConfigurationProvider {
     return debugConfiguration;
   }
 
-  private provideDebugMainClassConfiguration(): Thenable<
-    DebugConfiguration | undefined
-  > {
-    return this.askForBuildTarget().then((buildTarget) => {
-      if (buildTarget === undefined) return undefined;
-      return this.askForClassName().then((className) => {
-        if (className === undefined) return undefined;
-        return this.askForArguments().then((args) => {
-          if (args === undefined) return undefined;
-          return this.askForConfigurationName(className).then((name) => {
-            if (name === undefined) return undefined;
+  private provideDebugMainClassConfiguration(): Thenable<DebugConfiguration> {
+    return this.askForBuildTarget().then((buildTarget) =>
+      this.askForClassName().then((className) =>
+        this.askForArguments().then((args) =>
+          this.askForConfigurationName(className).then((name) => {
             const result: DebugConfiguration = {
               type: configurationType,
               name: name,
@@ -101,21 +95,16 @@ class ScalaConfigProvider implements vscode.DebugConfigurationProvider {
               args: args,
             };
             return result;
-          });
-        });
-      });
-    });
+          })
+        )
+      )
+    );
   }
 
-  private provideDebugTestClassConfiguration(): Thenable<
-    DebugConfiguration | undefined
-  > {
-    return this.askForBuildTarget().then((buildTarget) => {
-      if (buildTarget === undefined) return undefined;
-      return this.askForClassName().then((className) => {
-        if (className === undefined) return undefined;
-        return this.askForConfigurationName(className).then((name) => {
-          if (name === undefined) return undefined;
+  private provideDebugTestClassConfiguration(): Thenable<DebugConfiguration> {
+    return this.askForBuildTarget().then((buildTarget) =>
+      this.askForClassName().then((className) =>
+        this.askForConfigurationName(className).then((name) => {
           const result: DebugConfiguration = {
             type: configurationType,
             name: name,
@@ -124,12 +113,12 @@ class ScalaConfigProvider implements vscode.DebugConfigurationProvider {
             buildTarget: buildTarget,
           };
           return result;
-        });
-      });
-    });
+        })
+      )
+    );
   }
 
-  private askForBuildTarget(): Thenable<string | null | undefined> {
+  private askForBuildTarget(): Thenable<string | undefined> {
     return vscode.window
       .showInputBox({
         prompt: "Enter name of the build target",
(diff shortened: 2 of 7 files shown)


tgodzik:
And also, is it possible to support test configurations also?

src/scalaDebugger.ts
```diff
@@ -57,7 +61,33 @@ class ScalaConfigProvider implements vscode.DebugConfigurationProvider {
...
+      if(debugParameters === undefined) return null;
+      return vscode.commands.executeCommand<DebugSession>(startAdapterCommand, debugParameters).then(debugSession => {
+        if(debugSession === undefined) return null;
+        return debugServerFromUri(debugSession.uri);
+      });
+    });
+  }
+}
+
+export function debugServerFromUri(uri: string): vscode.DebugAdapterServer {
```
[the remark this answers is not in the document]
alekseiAlefirov: but it _is_ `scalaDebugger.ts`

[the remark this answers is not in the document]
tgodzik: haha, I am so blind

package.json
```diff
@@ -257,6 +257,63 @@
...
+        },
+        "configurationSnippets": [
+          {
+            "label": "Scala: Run main class",
+            "description": "A new configuration for running a debug session of a main class",
+            "body": {
+              "type": "scala",
+              "request": "launch",
+              "name": "Untitled",
+              "mainClass": "???",
```
tgodzik, line 302: We could most likely ask for name as a default line in the docs via:  `${command:AskForProgramName}` - https://code.visualstudio.com/api/extension-guides/debugger-extension

package.json
```diff
@@ -257,6 +257,63 @@
...
+            }
+          },
+          {
+            "label": "Scala: Run test class",
+            "description": "A new configuration for running a debug session of a test class",
+            "body": {
+              "type": "scala",
+              "request": "launch",
+              "name": "Untitled",
+              "testClass": "???"
```
tgodzik, line 314: Same here.

src/scalaDebugger.ts
```diff
@@ -57,6 +60,37 @@ class ScalaConfigProvider implements vscode.DebugConfigurationProvider {
...
+class ScalaDebugServerFactory implements vscode.DebugAdapterDescriptorFactory {
+  createDebugAdapterDescriptor(
+    session: vscode.DebugSession
+  ): ProviderResult<DebugAdapterDescriptor> {
+    if (
+      session.configuration.mainClass !== undefined ||
+      session.configuration.testClass !== undefined
+    ) {
+      return vscode.commands
+        .executeCommand<DebugSession>(
```
tgodzik, line 72: Maybe parallel to executing the command we could create the configuration?

package.json
alekseiAlefirov: Honestly, I spent some time on this, and could not make it work.
*But* I tried it from the extension sample provided in this article, and it does work, though it's not what we want - it ask user for the input when starting actual debug session,  not when the snippet is generated.
But there's an option to program *initial configuration* -- something that's generated when the `launch.json` is generated on user's request -- via ` vscode.DebugConfigurationProvider#provideDebugConfigurations`. So, we probably could ask, what class user wants to debug, there.

src/scalaDebugger.ts
alekseiAlefirov: by this moment configuration is already created and defines what's going on, see `session.configuration`.

package.json
tgodzik: That would be great! Doesn't need to be a part of this PR, but if you have time it would be awesome to try it.

src/scalaDebugger.ts
tgodzik: I was thinking that we should create a configuration when running via code lenses, so that the user can easily retry. Not sure where to put that though.

package.json
alekseiAlefirov: Yeah, should not be big.

src/scalaDebugger.ts
alekseiAlefirov: I think, this is good idea, but quite big to make it in another PR.

alekseiAlefirov:
Added dialog for creating the first (ever) debug configuration.
launch json debug
Not sure, if it's not too massive for the client.
@tgodzik please, check it out.

alekseiAlefirov:
PS, don't see, why it fails on format. On my machine:
```
[aalefirov@localhost metals-vscode]$ yarnpkg format-check
yarn run v1.19.1
warning metals@1.8.5: The engine "vscode" appears to be invalid.
$ prettier --check **/*.{ts,js,json,yml}
Checking formatting...
All matched files use Prettier code style!
Done in 0.81s.
```

gabro:
@alekseiAlefirov i think it’s the same but I’ve fixed today in metals-languageclient. I’ll take a look in the morning

gabro:
@alekseiAlefirov can you try rebasing. I have fixed the issue with prettier in #265

alekseiAlefirov:
@gabro thanks to take care of this. Seems like I had to rerun `yarn compile`. Now all the last item commas are back, as in master :)

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+    return null;
+  }
+}
+
+export function debugServerFromUri(uri: string): vscode.DebugAdapterServer {
+  const debugServer = vscode.Uri.parse(uri);
+  const segments = debugServer.authority.split(":");
+  return new vscode.DebugAdapterServer(
+    parseInt(segments[segments.length - 1]),
+    segments[0]
```
gabro, line 214: can you give meaningful names to the segments?

Something like:

```suggestion
  const segments = debugServer.authority.split(":");
  const protocol = segments[0]
  const port = parseInt(segments[segments.length - 1]);
  return new vscode.DebugAdapterServer(port, protocol);
```

(provided that `port` and `protocol` make sense as names)

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+  }
+
+  private askForArguments(): Thenable<string[] | undefined> {
+    return vscode.window
+      .showInputBox({ prompt: "Enter argument or leave it empty" })
+      .then((argument) => {
+        if (argument === undefined) {
+          return undefined;
+        } else if (argument === "") {
+          return [];
```
gabro, line 172: what's the difference between returning `undefined` or `[]`? Can we instead collapse the two cases and just return `[]`? This would also simplify the return type to `Thenable<string[]>`

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+        } else {
+          return this.askForArguments().then((rest) => {
+            if (rest === undefined) {
+              return [argument];
+            } else {
+              rest.unshift(argument);
+              return rest;
```
gabro, line 179: I generally avoid mutating arrays, you can use this instead:

```suggestion
              return [argument, ...rest];
```

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+  private askForArguments(): Thenable<string[] | undefined> {
+    return vscode.window
+      .showInputBox({ prompt: "Enter argument or leave it empty" })
+      .then((argument) => {
+        if (argument === undefined) {
+          return undefined;
+        } else if (argument === "") {
+          return [];
+        } else {
+          return this.askForArguments().then((rest) => {
```
gabro, line 174: what's the logic here? Asking for new arguments recursively until the user leaves the argument empty?

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+            if (rest === undefined) {
+              return [argument];
```
gabro, line 176: see my comment above about avoiding the `undefined` case

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+      prompt: "Enter name of the class to debug",
+      placeHolder: "<package>.<Class>",
+    });
+  }
+
+  private askForConfigurationName(
+    className: string
+  ): Thenable<string | undefined> {
+    return vscode.window.showInputBox({
+      prompt: "Enter name of the configuration",
```
gabro, line 160:
```suggestion
      prompt: "Enter the name of the configuration",
```

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+          return null;
+        } else {
+          return buildTarget;
+        }
+      });
+  }
+
+  private askForClassName(): Thenable<string | undefined> {
+    return vscode.window.showInputBox({
+      prompt: "Enter name of the class to debug",
```
gabro, line 151:
```suggestion
      prompt: "Enter the name of the class to debug",
```

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+      .then((buildTarget) => {
+        if (buildTarget === undefined) {
+          return undefined;
+        } else if (buildTarget === "") {
+          return null;
+        } else {
+          return buildTarget;
+        }
+      });
+  }
```
gabro, line 147: Can we use just `string | undefined` instead of carrying also `null` as a third option?

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+              mainClass: className,
+              buildTarget: buildTarget,
+              args: args,
+            };
+            return result;
+          });
+        });
+      });
+    });
+  }
```
gabro, line 108: suggestion: instead of using `undefined` to represent the exceptional case, maybe we can use `Promise.reject()` and then catch the `Promise` above (see my other comment).

This would avoid checking for `undefined` at every step

src/scalaDebugger.ts
```diff
@@ -46,15 +50,169 @@ export async function start(
...
+        if (result === mainClassPick) {
+          return this.provideDebugMainClassConfiguration().then((config) => {
+            if (config === undefined) return [];
+            return [config];
+          });
+        } else if (result === testClassPick) {
+          return this.provideDebugTestClassConfiguration().then((config) => {
+            if (config === undefined) return [];
+            return [config];
+          });
```
gabro, line 70: if we start using `Promise.reject` (see my comment below) this can be:

```suggestion
          return this.provideDebugTestClassConfiguration().then((config) => [config])
```

and then after the most outer `then` we can add

```ts
.catch(() => [])
```

src/scalaDebugger.ts
gabro: This would become

```suggestion
  private provideDebugMainClassConfiguration(): Thenable<DebugConfiguration> {
    return this.askForBuildTarget().then((buildTarget) =>
      this.askForClassName().then((className) =>
        this.askForArguments().then((args) =>
          this.askForConfigurationName(className).then((name) =>
            const result: DebugConfiguration = {
              type: configurationType,
              name: name,
              request: launchRequestType,
              mainClass: className,
              buildTarget: buildTarget,
              args: args,
            };
            return result;
          );
        );
      );
    );
  }
```

(provided that all sub-functions return `Thenable<A>` instead of `Thenable<A | undefined>` and fail with `Promise.reject` instead)

src/scalaDebugger.ts
alekseiAlefirov: done. I've put `host` instead of `protocol` -- at least it corresponds with `DebugAdapterServer` parameters.

src/scalaDebugger.ts
alekseiAlefirov: Done. I too had an itch on mutating, but lacked experience with JS.

src/scalaDebugger.ts
alekseiAlefirov: answered above.

src/scalaDebugger.ts
alekseiAlefirov: fixed with `onrejected` argument of `.then`

src/scalaDebugger.ts
alekseiAlefirov: done.

src/scalaDebugger.ts
alekseiAlefirov: done.

src/scalaDebugger.ts
alekseiAlefirov: now with `reject` we can :) previously there was collision -- we had to carry information, if user rejected input, and if they just provided no build target.

src/scalaDebugger.ts
alekseiAlefirov: done.

src/scalaDebugger.ts
alekseiAlefirov: Thanks for the `Promise.reject` advice! I miss experience in JS :)
Since it's `Thenable`, it misses `catch` method, but have second argument callback `onrejected`.

alekseiAlefirov:
@gabro , thanks for the review, PR reworked.

src/scalaDebugger.ts
```diff
@@ -46,15 +50,161 @@ export async function start(
...
+            (_) => []
+          );
+        } else if (result === testClassPick) {
+          return this.provideDebugTestClassConfiguration().then(
+            (config) => [config],
+            (_) => []
+          );
+        }
+        return [];
+      });
```
gabro, line 73:
```suggestion
      .then((result) => {
        switch (result) {
          case mainClassPick: return this.provideDebugMainClassConfiguration().then(config => [config]);
          case testClassPick: return this.provideDebugTestClassConfiguration().then(config => [config];
          default: return [];
        }
      }, () => [])
```

gabro: The diff isn't very helpful, but I've simply moved the fallback to the outer `then` and replaced the if/else with a switch (which is a bit more readable IMO)

gabro: (you'll need to run prettier, I think)

src/scalaDebugger.ts
```diff
@@ -46,15 +50,161 @@ export async function start(
...
+            testClass: className,
+            buildTarget: buildTarget,
+          };
+          return result;
+        })
+      )
+    );
+  }
+
+  private askForBuildTarget(): Thenable<string | undefined> {
```
gabro, line 121: do we still need `undefined` here? Can't we just use `Promise.reject()` for that case too?

src/scalaDebugger.ts
```diff
@@ -46,15 +50,161 @@ export async function start(
...
+
+  private askForClassName(): Thenable<string> {
+    return vscode.window
+      .showInputBox({
+        prompt: "Enter the name of the class to debug",
+        placeHolder: "<package>.<Class>",
+      })
+      .then((name) => {
+        if (name === undefined) return Promise.reject();
+        return name;
```
gabro, line 146: you can use the new fancy `??` operator if you want:


```suggestion
        return name ?? Promise.reject();
```

src/scalaDebugger.ts
```diff
@@ -46,15 +50,161 @@ export async function start(
...
+
+  private askForConfigurationName(className: string): Thenable<string> {
+    return vscode.window
+      .showInputBox({
+        prompt: "Enter the name of the configuration",
+        value: `Debug ${className}`,
+      })
+      .then((name) => {
+        if (name === undefined) return Promise.reject();
+        return name;
```
gabro, line 158:
```suggestion
        return name ?? Promise.reject();
```

src/scalaDebugger.ts
```diff
@@ -46,15 +50,161 @@ export async function start(
...
+      .showInputBox({ prompt: "Enter argument or leave it empty" })
+      .then((argument) => {
+        if (argument === undefined) {
+          return Promise.reject();
+        } else if (argument === "") {
+          return [];
+        } else {
+          return this.askForArguments().then(
+            (rest) => [argument, ...rest],
+            (_) => [argument]
```
gabro, line 173:
```suggestion
            () => [argument]
```

tgodzik:
> Added dialog for creating the first (ever) debug configuration.
> launch json debug
> Not sure, if it's not too massive for the client.
> @tgodzik please, check it out.

I think we can skip the arguments for main classes, it should be easy enough for people to figure out that they need to add them to the array in the configuration.

src/scalaDebugger.ts
alekseiAlefirov: done `switch`, thanks. To catch rejection, had to add another following `then`, `onrejection` here is not enough.

src/scalaDebugger.ts
alekseiAlefirov: yes, `undefined` reflects no `buildTarget` provided.

src/scalaDebugger.ts
alekseiAlefirov: done, thanks.

src/scalaDebugger.ts
alekseiAlefirov: done

src/scalaDebugger.ts
alekseiAlefirov: done

alekseiAlefirov:
@gabro PR reworked,
@tgodzik arguments removed, just empty array provided in result.
Here's updated demo:
launch_debug3

PR #233, opened 2020-03-10
Goto Super Method functionality

kpbochenek:
Exposes functionality implemented in PR: https://github.com/scalameta/metals/pull/1487

All functionality is described in mentioned PR, here I only expose those 2 functions implemented in `metals`


src/extension.ts
```diff
@@ -461,6 +461,45 @@ function launchMetals(
       }
     });
 
+    registerCommand("metals-goto-location", args => {
```
tgodzik: Why do we need to register the command? I think this will also go to the settings and will be able to be used by the user?

package.json
```diff
@@ -134,6 +134,11 @@
             "type": "string"
           },
           "markdownDescription": "Optional list of custom resolvers passed to Coursier when fetching metals dependencies.\n\nFor documentation on accepted values see the [Coursier documentation](https://get-coursier.io/docs/other-repositories).\n\nThe extension will pass these to Coursier using the COURSIER_REPOSITORIES environment variable after joining the custom repositories with a pipe character (|)."
+        },
+        "metals.superMethodLenses": {
```
ckipp01: Out of curiosity, will there be a situation where people specifically don't want this? I feel like it may be more common for someone to just disable code lenses rather than in individual feature that uses code lenses? I sort of feel like up to this point we don't off this level of customize-ability, but this starts down the path of more configuration. Do we want this?

package.json
```diff
@@ -134,6 +134,11 @@
...
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Enable/disable goto super method code lens"
```
ckipp01: If this doesn't utilize markdown at all, I believe it can just be description as well.

package.json
gabro: I'm all for not adding configuration, but code lenses are a weird beast and some of them can be more annoying than others.

This specific one would appear very often and cause a lot of shifting in the UI, which can be more annoying than the "Run"|"Debug" one (which is very rare)

tgodzik: I think we should have this option, because it does heavily impact the user's experience. The same happened for VS Code java plugin, where there were immediately request for an option to disable them.  

I imagine some people might want to just use the shortcuts and it might just not be useful for them to have these code lenses. Besides, disabling all lenses would mean disabling them everywhere and they still might want to have some other like for example for git.

>but this starts down the path of more configuration

We do need some customization and this doesn't add any overhead really. I agree we need to be careful about it, but using a simple boolean value is nothing that will make us start with even more configuration.

ckipp01: Sounds good 👍

src/extension.ts
kpbochenek: not needed, removed :+1:

package.json
kpbochenek: okay :+1:

package.json
kpbochenek: okay I will leave as it is then

tgodzik: Shouldn't it be `superMethodLensesEnabled` ?

src/extension.ts
```diff
@@ -654,6 +684,22 @@ function launchMetals(
...
+    location.range.start.character,
+    location.range.end.line,
+    location.range.end.character
+  );
+  const selection = new Selection(range.start, range.start);
+
+  workspace
+    .openTextDocument(Uri.parse(location.uri))
+    .then((textDocument) =>
+      window.showTextDocument(textDocument, { selection: selection })
```
tgodzik: You can just use `selection : range`

kpbochenek: changes

package.json
kpbochenek: correct, renamed :+1:

kpbochenek:
unrelated build failure:
```
Executing prepublish script 'yarn run vscode:prepublish'...
 ERROR  Command failed: yarn run vscode:prepublish
warning metals@1.8.4: The engine "vscode" appears to be invalid.
error Command failed with exit code 2.
```

gabro:
Can you run yarn prepublish directly and see whether you get better logs?

gabro:
Actually, prepublish is just `yarn compile` so you can try that locally

kpbochenek:
Thanks it was actually by unused import :)

There is still one error but it doesn't cause build to fail <weird>

```
const serverOptions = getServerOptions({
    metalsClasspath,
    serverProperties,
    javaConfig,
    clientName: "vscode",
  });
```

 is missing `doctorFormat: "html" | "json";`

gabro:
> Ok I need to update packages 👍

You beat me to it :) We removed that option in a recent version of `metals-languageclient`

PR #246, opened 2020-03-22
Replace vscode with @types/vscode

ckipp01:
The main `vscode` dependency that we use has split up, and it's recommended to instead use `@types/vscode` and `vscode-test`. This pr replaces the `vscode` dependency with `@types-vscode`. I also update other various dependencies as a few of them were lagging behind which was causing some of the recent updates to not work right.

https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode

- Update main `vscode` dependency
- Update other various dependencies
- ~Format code with new version of prettier~

I did build locally and test installing the latest snapshot of metals, hovers, completions, debugging, and worksheets. Everything seems to be working as expected, but it's best if someone else also pulls this, builds it, and plays around with it, especially since I don't use vscode as my main editor.  I'm probably the worst person to do this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   },
   "homepage": "http://metals.rocks",
   "engines": {
-    "vscode": "^1.39.0"
+    "vscode": "^1.43.0"
   },
   "icon": "images/logo.png",
   "categories": [
@@ -266,7 +266,6 @@
     "vscode:prepublish": "yarn compile",
     "compile": "tsc -p ./",
     "watch": "tsc -watch -p ./",
-    "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "yarn compile",
     "build": "vsce package --yarn",
     "vscode:publish": "vsce publish --yarn",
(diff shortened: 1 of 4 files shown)


gabro:
Make sure to thoroughly test this, by publishing a vsix locally and installing it by hand. I’ve tried this in the past and the extension would not activate (i wasn’t able to pinpoint the exact issue), although it was working fine when launching it from the debugger

package.json
```diff
@@ -266,25 +266,24 @@
...
-    "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "yarn compile",
     "build": "vsce package --yarn",
     "vscode:publish": "vsce publish --yarn",
     "format": "prettier --write **/*.{ts,js,json,yml}",
     "format-check": "prettier --check **/*.{ts,js,json,yml}"
   },
   "devDependencies": {
-    "@types/node": "^13",
+    "@types/node": "^13.9.3",
```
gabro: since we’re at it, can you remove all the ^ from the dependencies? They’re not very useful since we’re using yarn.lock and dependabot

src/extension.ts
```diff
@@ -293,7 +293,7 @@ function launchMetals(
 
   context.subscriptions.push(client.start());
 
-  client.onReady().then(_ => {
+  client.onReady().then((_) => {
```
gabro: this should have been () instead of _. Can we make the change now?

package.json
```diff
@@ -266,25 +266,24 @@
...
     "@types/shell-quote": "^1.6.1",
-    "prettier": "1.19.1",
+    "prettier": "2.0.1",
     "typescript": "^3.8.3",
-    "vsce": "^1.54.0"
+    "vsce": "^1.74.0"
   },
   "dependencies": {
-    "metals-languageclient": "0.1.20",
+    "@types/vscode": "^1.43.0",
```
gabro: this should go in devDependencies now

ckipp01:
> Make sure to thoroughly test this, by publishing a vsix locally and installing it by hand.
Sounds good 👍  I was able to do this with no issues.

> The relevant code changes are 👍 I’ve left a few minor comments.
Addressed the comments and made the changes.

> However I find the prettier change a bit distracting and out of scope for the PR.
> Can we do the prettier 2 upgrade in a separate PR so that we keep this diff on this one more focused?
Sure! I undid that update and I'll send in another pr for it.

PR #247, opened 2020-03-22
Update prettier

ckipp01:
Format everything with new version of prettier
Add in section to contributing about publishing locally

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,6 +38,16 @@ application.
 When you make further changes to the extension, quit the "Extension Development
 Host" and run `F5` (or `Run > Start debugging`) again.
 
+## Publishing and testing locally
+
+To ensure that everything is working as expected after changes it's best to
+publish a vsix artifact locally. You can do this by `yarn build` which will result in
+the metals-<version>.vsix being built and placed in the root of your directory.
+You can then simply `code <path-to-vsix-file>` which will install the extension.
+You can install the extension by going to the extension menu in Visual Studio
+Code and under "More Actions..." choose to "Install from VSIX...". Then open a
+`*.scala` file to ensure that Metals starts and works correctly.
+
 ## Testing
 
 The extension currently has no tests. When you are happy with the functionality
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -38,6 +38,16 @@ application.
 When you make further changes to the extension, quit the "Extension Development
 Host" and run `F5` (or `Run > Start debugging`) again.
 
+## Publishing and testing locally
+
+To ensure that everything is working as expected after changes it's best to
+publish a vsix artifact locally. You can do this by `yarn build` which will result in
+the metals-<version>.vsix being built and placed in the root of your directory.
+After the artifact has successfully build you can install the extension by going to
+the extension menu in Visual Studio Code and under "More Actions..." choose to
```
gabro: There’s also a way to do this from the command line. I’m from mobile right now but it should be easy to find and it may make sense to document it here

ckipp01: Sure, I added that in as well. 👍

CONTRIBUTING.md
```diff
@@ -38,6 +38,16 @@ application.
 When you make further changes to the extension, quit the "Extension Development
 Host" and run `F5` (or `Run > Start debugging`) again.
 
+## Publishing and testing locally
+
+To ensure that everything is working as expected after changes it's best to
+publish a vsix artifact locally. You can do this by `yarn build` which will result in
+the metals-<version>.vsix being built and placed in the root of your directory.
+You can then simply `code <path-to-vsix-file>` which will install the extension.
```
gabro: does it work this way? I remember that was a flag for it like —install or something

ckipp01: Yea, you can actually just open up a vsix file via the file opener as well and it will install it. 
> Install a Packaged Extension (.vsix)

>You can manually install an VS Code extension packaged in a .vsix file. Simply install using the VS Code command line providing the path to the .vsix file.

>`code myExtensionFolder\myExtension.vsix`

>The extension will be installed under your user .vscode/extensions folder. You may provide multiple .vsix files on the command line to install multiple extensions at once.

>You can also install a .vsix by opening the file from within VS Code. Run File | Open File... or kb(workbench.action.files.openFile) and select the extension .vsix.

gabro: Ah, TIL! Thanks

PR #251, opened 2020-03-25
Add Ammonite build server-related commands

alexarchambault:
Related to https://github.com/scalameta/metals/pull/1538.

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,16 @@
         "command": "metals.go-to-super-method",
         "category": "Metals",
         "title": "Go to super method"
+      },
+      {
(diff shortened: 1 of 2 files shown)


ckipp01:
Now that https://github.com/scalameta/metals/pull/1538 is merged in we may want to merge this so people have these commands. Thoughts? These are still the only ones necessary right @alexarchambault?

alexarchambault:
I think we should merge this, yes. These aren't absolutely required for the Ammonite support, but can be handy if ever one wants to manually stop / start the Ammonite server. It's also the only way to stop the Ammonite server, apart from shutting down the metals server.

PR #269, opened 2020-04-10
Update default server versions to 0.8.4

tgodzik:


package.json
```diff
@@ -35,7 +35,7 @@
   },
   "homepage": "http://metals.rocks",
   "engines": {
-    "vscode": "^1.43.0"
+    "vscode": "^1.44.0"
```
gabro, line 38: we should not update @types/vscode automatically in the future (that was the root cause).
I'll configure dependabot to avoid it

tgodzik: Thanks!

PR #275, opened 2020-04-20
Add `go to super method` and `reveal method hierarchy`to palette

tgodzik:
@olafurpg  added this as mentioned in https://github.com/scalameta/metals-feature-requests/issues/101#issuecomment-615336020

Previously, I thought this wouldn't work correctly, but I tested and it seems fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,6 +196,16 @@
         "command": "metals.new-scala-file",
         "category": "Metals",
         "title": "New Scala File..."
+      },
+      {
+        "command": "metals.super-method-hierarchy",
+        "category": "Metals",
+        "title": "Reveal super method hierachy"
+      },
+      {
+        "command": "metals.go-to-super-method",
+        "category": "Metals",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -157,6 +157,16 @@
...
+      {
+        "command": "metals.super-method-hierarchy",
+        "category": "Metals",
+        "title": "Reveal super method hierachy"
+      },
+      {
+        "command": "metals.go-to-super-method",
+        "category": "Metals",
+        "title": "Go to super method"
+      },
```
gabro, line 169: can you also add them to the list below, in order to enable them only when Metals is enabled?

tgodzik: Good idea! We actually forgot about that for some more commands. I added all of the missing ones, just left the `Reveal metals view` ones, since they do work without the server and still do something in that case.

gabro: Good catch 👍

PR #289, opened 2020-04-29
Add new icons for the metals view and a separate compilations section

tgodzik:
Change needed for https://github.com/scalameta/metals/pull/1677


src/treeview.ts
```diff
@@ -243,6 +244,7 @@ class MetalsTreeDataProvider implements TreeDataProvider<string> {
         }
       }
     }
+    return new ThemeIcon(icon);
```
tgodzik, line 247: If `icon` is an existing VS Code icon it will now show up here. We could probably remove some icons and replace with existing ones, but I think it's fine as is now.

tgodzik:
Hopefully no one minds if I merge this for the release. There is mostly moving things around and is a extension of the metals PR

PR #290, opened 2020-05-03
Move from ClientExperimentalCapabilities to IntializationOptions

ckipp01:
This pr sort of demonstrates part of the reason I wanted to go more the `InitializationOptions` route. It's incredibly less verbose to set these this way. I did test the existing settings locally with this, but we need to wait for the few to be merged in here: https://github.com/scalameta/metals/pull/1684, then we can try debugging, treeview, etc to ensure it's working as expected.

Now that Metals is able to set the majority of the settings via InitializationOptions
we can move away from setting them via ClientExperimentalCapabilities towards
setting them in a much easier way via InitializationOptions


src/extension.ts
```diff
@@ -634,56 +640,45 @@ function launchMetals(
         );
       });
     });
-    if (features.treeViewProvider) {
```
ckipp01, line 637: So I see that there are a few checks here to ensure that these features are on, but AFIK, these are sort of pointless right? Since we are the ones setting them, and we always have them, these should be fine to remove right?

tgodzik: Should be fine I think. The only issue might be if someone want's to go waaaay back.

ckipp01:
Alright, just verifying now that https://github.com/scalameta/metals/pull/1684 was merged. The things that we carried over are indeed working. For example:

Worksheets
Screenshot 2020-05-04 at 15 33 31

Debugging
Screenshot 2020-05-04 at 15 32 57

Tree View
Screenshot 2020-05-04 at 15 31 48

Issue #299, opened 2020-05-13
it generates *always* .metals directory in VSCode


Raffaello:
**Describe the bug**
it just always running even in non JVM / scala projects.

**To Reproduce**
install scala metals
and for example create a workspace that using go or python languages.

**Expected behavior**
to do not have .metals directory that is completely irrelevant for the project.

**Screenshots**

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code
- Metals version: v1.9.0

**Additional context**

**Search terms**

olafurpg:
A workaround is to manually disable the Metals extension per-workspace, or go the other way around and manually enable the extension per-workspace

Raffaello:
i noticed now: could be related to this repo the issue too:
https://github.com/scalameta/metals-vscode

FranklinChen:
I just noticed this bug when opening a TypeScript project in VS Code and finding that Metals was launched and `.metals` created.

tgodzik:
@FranklinChen did it happen with the newest version of the extension too? We reduced the number of patterns now, which should help out in situations where Metals is loaded due to timeout.

These are the current activasion events:
```
  "activationEvents": [
    "onCommand:metals.new-scala-project",
    "onDebugResolve:scala",
    "onLanguage:scala",
    "workspaceContains:build.sbt",
    "workspaceContains:build.sc",
    "workspaceContains:project/build.properties",
    "workspaceContains:src/main/scala",
    "workspaceContains:*/src/main/scala",
    "workspaceContains:*/*/src/main/scala"
  ],
```

If the change didn't help, we might need to remove the last 2 ones.

FranklinChen:
It seems to be OK for me now.

tgodzik:
Ok, I will close it for now, but if this ever happens again please report the issue here. Thanks!

PR #300, opened 2020-05-13
Add a feature to create a new Project

tgodzik:
This is a follow up of https://github.com/scalameta/metals/pull/1728

In particular this adds:
- a welcome view with the new project button - it requires to have an additional activation command, since the normal "New Scala Project" command will only be added after the client is ready. 
- new extension to open a new window automatically (without it, there will only be a message saying that the new project got create)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "Programming Languages"
   ],
   "activationEvents": [
+    "onCommand:metals.new-scala-project-activate",
     "onLanguage:scala",
     "workspaceContains:build.sbt",
     "workspaceContains:build.sc",
@@ -65,6 +66,18 @@
         }
       ]
     },
+    "viewsWelcome": [
+      {
+        "view": "metalsPackages",
+        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project-activate)",
+        "when": "!metals:enabled"
+      },
+      {
+        "view": "metalsPackages",
+        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project)",
+        "when": "metals:enabled && metals:no-build-tool"
+      }
+    ],
     "views": {
       "metals-explorer": [
         {
@@ -202,6 +215,11 @@
         "title": "Run doctor"
       },
       {
+        "command": "metals.new-scala-project",
+        "category": "Metals",
+        "title": "New Scala Project"
+      },
+      {
         "command": "metals.new-scala-file",
         "category": "Metals",
         "title": "New Scala File..."
@@ -268,6 +286,10 @@
         },
         {
           "command": "metals.new-scala-file",
+          "when": "metals:enabled"
+        },
+        {
+          "command": "metals.new-scala-project",
           "when": "metals:enabled"
         },
         {
--- a/package.json
+++ b/package.json
@@ -42,12 +42,15 @@
     "Programming Languages"
   ],
   "activationEvents": [
-    "onCommand:metals.new-scala-project-activate",
+    "onCommand:metals.new-scala-project",
+    "onDebugResolve:scala",
     "onLanguage:scala",
     "workspaceContains:build.sbt",
     "workspaceContains:build.sc",
     "workspaceContains:project/build.properties",
-    "workspaceContains:**/scala/**"
+    "workspaceContains:src/main/scala",
+    "workspaceContains:*/src/main/scala",
+    "workspaceContains:*/*/src/main/scala"
   ],
   "contributes": {
     "configurationDefaults": {
@@ -69,13 +72,7 @@
     "viewsWelcome": [
       {
         "view": "metalsPackages",
-        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project-activate)",
-        "when": "!metals:enabled"
-      },
-      {
-        "view": "metalsPackages",
-        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project)",
-        "when": "metals:enabled && metals:no-build-tool"
+        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project)"
       }
     ],
     "views": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -33,7 +33,6 @@ import {
   Location,
   TextDocumentPositionParams,
   TextDocument,
-  Disposable,
 } from "vscode-languageclient";
 import { ClientCommands } from "./client-commands";
 import {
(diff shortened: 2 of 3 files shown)


src/extension.ts
```diff
@@ -80,17 +82,55 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
+  shouldCreateNewProject: boolean;
+  status: Disposable;
+}
+
 export async function activate(context: ExtensionContext) {
   detectLaunchConfigurationChanges();
   checkServerVersion();
+  let activation: Activation = {
+    shouldCreateNewProject: false,
+    status: window.setStatusBarMessage("$(gear~spin) Activating Metals"),
```
tgodzik, line 95: Added this to show the user that something is happening after clicking the `New Scala Project` button.

src/extension.ts
```diff
@@ -80,17 +82,55 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
+      activation.status.dispose();
     });
   commands.executeCommand("setContext", "metals:enabled", true);
+  setHasBuildTool();
+}
+
+function setHasBuildTool(): void {
+  let allPromises = [
+    "build.sbt",
+    " build.sc",
```
olafurpg, line 118: Is this space intentional?
```suggestion
    "build.sc",
```

src/extension.ts
```diff
@@ -80,17 +82,55 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
+    "project/build.properties",
+    "**/scala/**",
```
olafurpg, line 120: Is this pattern necessary? I am concerned this glob will timeout in large workspaces causing unexpected behavior. I work on a daily basis in a git monorepo where the following command takes >1min to complete `time find . -type d -name scala | wc -l`

src/extension.ts
olafurpg: Did you look into using `window.withProgress(..)` instead? (I'm not sure it makes sense, just throwing out ideas)

olafurpg:
I am really excited about this change! Thank you for working on this. A lot of beginner users have asked questions about how to get Metals working, which is not surprising since Metals currently assumes the user has already created an sbt build and has no features to help users create the sbt build in the first place.

src/extension.ts
tgodzik: I looked into it, but it seems the whole task needs to be included inside the `withProgress`. Haven't found a way yet to show/hide `withProgress` But that would be much better.

tgodzik: I looked into it, but it seems the whole task needs to be included inside the `withProgress`. Haven't found a way yet to show/hide `withProgress` But that would be much better.

src/extension.ts
tgodzik: No! Thanks a lot, no idea how it got there.

src/extension.ts
tgodzik: It's more fool proof, but it does cause us some issues. We could maybe have a couple of more specific ones? Something like `src/main/scala/**` and  `*/src/main/scala/**` should cover more bases and be much faster.

tgodzik: Otherwise, I have no idea how to support maven, pants and gradle.

src/extension.ts
```diff
@@ -80,17 +82,55 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
+}
+
+function setHasBuildTool(): void {
+  let allPromises = [
+    "build.sbt",
+    "build.sc",
+    "project/build.properties",
+    "**/scala/**",
+  ].map(async (glob) => {
+    const files = await workspace.findFiles(glob);
```
kpbochenek, line 122: You can concat all patterns into one glob: `{build.sbt, build.sc, ...}`
Then you can specify excluded globs : `{.git, .bloop, **target/, ...}`
Then you can specify max results: `maxResults=1` (hopefully it will stop searching whole workspace if anything will be found)

src/extension.ts
```diff
@@ -318,6 +368,11 @@ function launchMetals(
   context.subscriptions.push(client.start());
 
   client.onReady().then(() => {
+    activation.status.dispose();
+    if (activation.shouldCreateNewProject)
+      client.sendRequest(ExecuteCommandRequest.type, {
```
kpbochenek, line 373: You can extract this to method as you duplicated this part also in `registerCommand`

src/extension.ts
```diff
@@ -80,17 +82,55 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
       outputChannel.appendLine(err);
       showMissingJavaMessage();
+      activation.status.dispose();
     });
   commands.executeCommand("setContext", "metals:enabled", true);
+  setHasBuildTool();
+}
+
+function setHasBuildTool(): void {
+  let allPromises = [
```
gabro, line 116:
```suggestion
  const detectedBuildTools = [
```

src/extension.ts
```diff
@@ -80,17 +82,55 @@ let decorationType: TextEditorDecorationType = window.createTextEditorDecoration
...
+  Promise.all(allPromises).then((all) => {
+    if (
+      all.every((a) => {
+        return a;
+      })
+    ) {
+      commands.executeCommand("setContext", "metals:no-build-tool", true);
+    }
+  });
 }
```
gabro, line 134: Absolutely optional proposal to rewrite this. It's easier to read for me this way, but it's clearly subjective:

```suggestion
function setHasBuildTool(): Promise<void> {
  const buildToolGlobs = [
    "build.sbt",
    "build.sc",
    "project/build.properties",
    "**/scala/**",
  ]
 
  return Promise.all(buildToolGlobs.map(workspace.findFiles)).then(matchResults => {
    if (matchResults.every(matchedFiles => matchedFiles == 0) {
      return commands.executeCommand("setContext", "metals:no-build-tool", true);
    }
  })
}
```

(in Scala this would be a `traverse`, obviously :P)

tgodzik: Thanks! I wasn't sure how to write this properly in type script.

src/extension.ts
tgodzik: I am not sure it would help a lot here. It would be around the same amount of code.

src/extension.ts
tgodzik: I actually refactored the glob pattern as suggested by @kpbochenek and it seems it should be better now.

src/extension.ts
tgodzik: Fixed! Great idea, thanks!

src/extension.ts
tgodzik: I simplified the patterns a bit, which should be much more efficient:
```
    "workspaceContains:build.sbt",
    "workspaceContains:build.sc",
    "workspaceContains:project/build.properties",
    "workspaceContains:src/main/scala",
    "workspaceContains:*/src/main/scala",
    "workspaceContains:*/*/src/main/scala"
```
This should cover most cases and in all other Metals will still activate after opening a file.

src/extension.ts
tgodzik: Ok, actually coming back to it, changed to withProgress and added a few returns so `then` would actually work with a promise not void.

package.json
```diff
@@ -42,11 +42,15 @@
     "Programming Languages"
   ],
   "activationEvents": [
+    "onCommand:metals.new-scala-project-activate",
+    "onDebugResolve:scala",
```
tgodzik, line 46: Added `onDebugResolve`, which might be useful in some cases.

tgodzik:
Ok, turns out that I don't actually have to add an additional command for running when activating the extension as long as we properly add returns everywhere, so that activation actually finishes when the client is ready.

I removed the aditional `new-scala-project-activate` and the additional setting of the `context.no-build-tool`, which greatly simplified things.

src/extension.ts
tgodzik: Aaaand I removed it altogether :sweat_smile:

src/extension.ts
```diff
@@ -317,7 +331,7 @@ function launchMetals(
 
   context.subscriptions.push(client.start());
 
-  client.onReady().then(() => {
+  return client.onReady().then(() => {
```
tgodzik, line 334: Added return to most functions so that both `withProgress` will work properly and the activasion function function will indeed finish at the proper time.

src/extension.ts
```diff
@@ -418,6 +432,17 @@ function launchMetals(
         case "metals-model-refresh":
           compilationDoneEmitter.fire();
           break;
+        case "metals-open-window":
```
kpbochenek, line 435: just left a comment in metals that it might be better to rename it to `open-folder` / `open-directory` or something like that. If you rename it in metals it needs change here too. If not then ignore :)

src/extension.ts
```diff
@@ -418,6 +432,17 @@ function launchMetals(
...
+          const openWindowParams =
+            params.arguments && (params.arguments[0] as MetalsOpenWindowParams);
```
gabro, line 437:
```suggestion
          const openWindowParams = params.arguments?[0] as MetalsOpenWindowParams;
```

tgodzik: Changed to `params.arguments?.[0] as MetalsOpenWindowParams;`

tgodzik:
Also fixes https://github.com/scalameta/metals-vscode/issues/299

PR #301, opened 2020-05-13
Fix recent bad merge and make sure we have 3 checks on automatic merges

tgodzik:


gabro:
Thanks for fixing this!

PR #312, opened 2020-05-28
Add configuration for enableStripMarginOnTypeFormatting

mlachkar:


package.json
```diff
@@ -143,6 +143,11 @@
           "type": "boolean",
           "default": false,
           "description": "Enable/disable goto super method code lens."
+        },
+        "metals.enableStripMarginOnTypeFormatting": {
+          "type": "boolean",
+          "default": true,
+          "description": "When enabled, if you press the return key from the first line of a multiline string containing a pipe, it will automatically add .stripMargin."
```
ckipp01, line 150: Super nitpicky, but instead of using `description` here, you could use `markdownDescription` and then actually put `stripMargin` in backticks which might appear a bit nicer in vscode.

mlachkar: ok ! thanks for your review

package.json
```diff
@@ -377,7 +382,7 @@
     "@types/node": "14.0.5",
     "@types/shell-quote": "1.6.2",
     "@types/vscode": "1.43.0",
-    "prettier": "2.0.5",
+    "prettier": "^2.0.5",
```
gabro, line 385: minor: why this change?
For context: we rely on dependabot to send patches for the updates so we prefer to use fixed versions in the `package.json`

mlachkar: I thought it was added by `yarn format`. 
I fixed it now.

gabro: thanks :)

Issue #332, opened 2020-06-24
Command "Reveal Active File in Side Bar" works only once


mgibowski:
**Describe the bug**

**To Reproduce** 

1. Open any Scala file in vs code with metals extension enabled
2. Run command "Reveal Active File in Side Bar"
3. Close side bar (cmd + B)
4. Repeat running command "Reveal Active File in Side Bar"

**Expected behavior**

Having the side bar opened with metals focused on active file.

Currently nothing gets displayed, but it seems that this command triggers recompilation of the project.

**Installation:**

- Operating system: macOS
- VSCode version: 1.46.1
- VSCode extension version: 1.9.0
- Metals version: 0.9.0

tgodzik:
Thanks for reporting! I can confirm the issue, if the class is revealed, it will not be focused later even if we run the command again.

Not sure if it's on the server or client side yet.

PR #333, opened 2020-06-24
Focus item in tree view even if it's already expanded

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/332

### DIFF
--- a/src/treeview.ts
+++ b/src/treeview.ts
@@ -112,8 +112,14 @@ export function startTreeView(
         if (i < params.uriChain.length) {
           const uri = params.uriChain[i];
           const isExpanded = expandedNode(params.viewId).has(uri);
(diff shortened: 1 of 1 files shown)


src/treeview.ts
```diff
@@ -112,8 +112,14 @@ export function startTreeView(
         if (i < params.uriChain.length) {
           const uri = params.uriChain[i];
           const isExpanded = expandedNode(params.viewId).has(uri);
+          const isDestinationNode = i == 0;
```
gabro, line 115: === for consistency (and safety)

gabro: (I see you just moved the line, but it’s a good chance to fix it nonetheless)

tgodzik: Fixed!

PR #336, opened 2020-06-25
Add a link to show logs when running slow task 

tgodzik:
Previously, we would always open Metals logs when running slow task, which could cause the side panel to be clsoed quite randomly. Now we offer a link to the logs within the notification.

It's not possible to add additional buttons to the notification unfortunately.

show-logs

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -597,19 +597,14 @@ function launchMetals(
     // bar with a "cancel" button.
     client.onRequest(MetalsSlowTask.type, (params, requestToken) => {
       return new Promise((requestResolve) => {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -597,19 +597,14 @@ function launchMetals(
     // bar with a "cancel" button.
     client.onRequest(MetalsSlowTask.type, (params, requestToken) => {
       return new Promise((requestResolve) => {
+        const showLogs = ` ([show logs](command:${ClientCommands.toggleLogs} "Show Metals logs"))`;
```
tgodzik, line 600: I was also thinking of `- show logs`, but not sure what is best

olafurpg: This looks good! We can refine it later if it turns out to be weird

Issue #342, opened 2020-07-02
"New Scala Project" Button on VSCode doesn't do anything when pressed 


siddhant3s:
The new "New Scala Project" that was added in the last Cobalt release doesn't do anything when clicked from a new VSCode Window.

It seems that the button only functions when we have already opened a directory. 

**To Reproduce** Steps to reproduce the behavior:

1. Open a brand new VSCode Window by going to File -> New Window
2. Go to the Metals tab on the left hand side bar
3. You should see a blue "New Scala Project" button under Packages pane.
4. Click on the button

Expected behavior: The "new project button" should start the new project flow or show an error explaining that the user needs to open a blank directory

**Screenshots**

<img width="1028" alt="image" src="https://user-images.githubusercontent.com/295892/86388233-42593b00-bc49-11ea-92c0-b12561bc0a35.png">

**Installation:**

- Operating system: macOS/
- VSCode version:1.46.1
- VSCode extension version:1.9.1
- Metals version: 0.9.1

kpbochenek:
Thanks for reporting!

Confirmed nothing is happening if no directory is opened.

PR #346, opened 2020-07-08
Make sure the new project button doesn't show up in an empty workspace

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/342

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
     "viewsWelcome": [
       {
         "view": "metalsPackages",
-        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project)"
+        "contents": "No Scala project found.\n[New Scala project](command:metals.new-scala-project)",
+        "when": "workbenchState != empty"
       }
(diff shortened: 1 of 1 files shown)


kpbochenek:
What is blocking us from actually making this button work? 

I think this button is most useful when you have no workspace open.

Also it might mislead people that sometimes they see this button and sometimes not(how would they know they need to have any workspace open to see this button?)

Maybe if this is technically not possible to make it work, then on pressing show message that to make it work any workspace needs to be opened?

tgodzik:
> What is blocking us from actually making this button work?
> 
> I think this button is most useful when you have no workspace open.
> 
> Also it might mislead people that sometimes they see this button and sometimes not(how would they know they need to have any workspace open to see this button?)
> 
> Maybe if this is technically not possible to make it work, then on pressing show message that to make it work any workspace needs to be opened?

I was playing around with it, but since we are lacking a workspace this would require a lot more work on the server side.

PR #349, opened 2020-07-09
Reset popup choices support

kpbochenek:
This is work related to PR in metals (https://github.com/scalameta/metals/pull/1897)

Links will be displayed thanks to PR in metals, they will become functional when this PR will be merged.
2020-07-09-110034_595x117_scrot

Links execute LSP command e.g. `metals.reset-choice [import build]`

It also exposes `metals.reset-choice-interactive` that can be mapped to a keyboard shortcut that allows to reset available choices interactively:
2020-07-09-110426_543x202_scrot

Example binding:
2020-07-09-110627_590x63_scrot

I have a feeling it could be good to document somewhere but I have no clue where would be best place.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -338,7 +338,8 @@ function launchMetals(
         doctor = window.createWebviewPanel(
           "metals-doctor",
           "Metals Doctor",
-          ViewColumn.Active
+          ViewColumn.Active,
+          { enableCommandUris: true }
         );
         context.subscriptions.push(doctor);
         doctor.onDidDispose(() => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -528,6 +529,20 @@ function launchMetals(
...
+      });
+    });
+
+    registerCommand("metals.reset-choice", (args) => {
+      client.sendRequest(ExecuteCommandRequest.type, {
+        command: "reset-choice",
+        arguments: args,
+      });
+    });
+
```
ckipp01, line 545: Shouldn't these also be added under the `commands` in the `package.json`?

kpbochenek: good catch!
only `metals.reset-choice-interactive` should be added?
as I understand commands in package.json will be added to 'Show all commands' popup?
And those commands must be parameterless?

Will double check

ckipp01: Correct. 🤔 Thinking about this I now for myself, since `reset-choice-interactive` is really the only one I will use, I may actually only register `reset-choice` as the name.

kpbochenek: added :)
2020-07-10-111555_688x143_scrot

PR #359, opened 2020-07-16
Add a step to publish to Open VSX registry

tgodzik:
I added the OVSX_TOKEN to secrets and it should all work. I will test it out with the next release.

Fixes https://github.com/scalameta/metals-vscode/issues/360

A follow up from https://github.com/eclipse/open-vsx.org/issues/83

### DIFF
--- a/package.json
+++ b/package.json
@@ -429,6 +429,7 @@
     "test": "yarn compile",
     "build": "vsce package --yarn",
     "vscode:publish": "vsce publish --yarn",
+    "ovsx:publish": "ovsx publish",
     "format": "prettier --write '**/*.{ts,js,json,yml}'",
     "format-check": "prettier --check '**/*.{ts,js,json,yml}'"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -429,6 +429,7 @@
     "test": "yarn compile",
     "build": "vsce package --yarn",
     "vscode:publish": "vsce publish --yarn",
+    "ovsx:publish": "ovsx publish",
```
spoenemann, line 432: You could add the `--yarn` option here, too.

tgodzik: I am reusing the previously build package and running publish on a file:

`yarn ovsx:publish ./metals-${{ steps.get_version.outputs.NEW_VERSION }}.vsix`

Would that also benefit from `--yarn`? Or since that file is already built it doesn't matter?

spoenemann: Right, in that case it doesn't matter.

tgodzik: Ok, thanks!

Issue #366, opened 2020-07-27
java.lang.ClassNotFoundException: scala.meta.metals.Main


tym-network:
**Describe the bug**

Metals server crashes when opening a Scala project.

**To Reproduce**

1. Open a Scala project in VSCode
2. In the Notification drawer, it says "Starting Metals server..."
3. After a few seconds, the following error in displayed in the notification server: "The Metals server crashed 5 times in the last 3 minutes. The server will not be restarted."

**Expected behavior**

The Metals server starts properly.

**Screenshots**
![Metals1](https://user-images.githubusercontent.com/2879545/88508390-8e22aa00-cf93-11ea-9156-87a5ca64988c.png)
![Metals2](https://user-images.githubusercontent.com/2879545/88508391-8ebb4080-cf93-11ea-8c64-46a3bcee4fd6.png)

**Installation:**

- Operating system: Windows 10
- VSCode version: 1.47.2
- VSCode extension version: 1.9.2
- Metals version: 0.9.2

**Additional context**

Here is the content of the log file located in AppData/Roaming/Code/logs/xxx/exthost1/output_logging_xxx:
```
Java home: C:\Program Files\Java\jdk-11.0.8
Error: Could not find or load main class scala.meta.metals.Main
Caused by: java.lang.ClassNotFoundException: scala.meta.metals.Main
[Info  - 22:51:16] Connection to server got closed. Server will restart.
Error: Could not find or load main class scala.meta.metals.Main
Caused by: java.lang.ClassNotFoundException: scala.meta.metals.Main
[Info  - 22:51:17] Connection to server got closed. Server will restart.
Error: Could not find or load main class scala.meta.metals.Main
Caused by: java.lang.ClassNotFoundException: scala.meta.metals.Main
[Info  - 22:51:17] Connection to server got closed. Server will restart.
Error: Could not find or load main class scala.meta.metals.Main
Caused by: java.lang.ClassNotFoundException: scala.meta.metals.Main
[Info  - 22:51:17] Connection to server got closed. Server will restart.
Error: Could not find or load main class scala.meta.metals.Main
Caused by: java.lang.ClassNotFoundException: scala.meta.metals.Main
[Error - 22:51:17] Connection to server got closed. Server will not be restarted.
```

I installed Java 11 and specified the path in VSCode settings `metals.javaHome`.
I also installed Scala using the .exe

Could the error be because I have an accent in my first name (which is also in the name of my home folder)? 

**Search terms**
[...]

PR #368, opened 2020-07-28
Log command used to start metals

kpbochenek:
Output:
```
Java home: /usr/lib/jvm/default
Starting metals using command '/usr/lib/jvm/default/bin/java' with args '-Xss4m,-Xms100m,-Dmetals.client=vscode,-classpath,/home/kp...
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -311,6 +311,14 @@ function launchMetals(
     clientOptions
   );
 
+  {
+    const cmd = serverOptions.run.command;
+    const args = serverOptions.run.args;
+    outputChannel.appendLine(
+      `Starting metals using command '${cmd}' with args '${args}'`
+    );
+  }
+
   function registerCommand(command: string, callback: (...args: any[]) => any) {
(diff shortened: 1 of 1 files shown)


kpbochenek:
@tgodzik logging a java path and arguments used could be helpful debugging issues like https://github.com/scalameta/metals-vscode/issues/366 ? WDYT?

Extension internally constructs a command to be executed from cmd and args so I can't log exact command but maybe this will be helpful enough?

Not 100% sure about this that's why 'draft' and waiting for your opinion

src/extension.ts
```diff
@@ -311,6 +311,14 @@ function launchMetals(
     clientOptions
   );
 
+  {
```
tgodzik, line 314: Does it need to be inside brackets?

kpbochenek: nope, just scoping cmd and args

tgodzik:
I think we don't necessarily need this now. Users can easily check the java process if needed. I will close it for now and reopen if more people want this change.

PR #371, opened 2020-08-01
Analyze stacktrace

kpbochenek:
(see description and discussion in scalameta/metals PR)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,6 +6,7 @@ import {
   workspace,
   ExtensionContext,
   window,
+  env,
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -539,6 +570,31 @@ function launchMetals(
       }
     );
 
+    registerCommand("metals.analyze-stacktrace", () => {
```
ckipp01, line 573: When this is merged, https://github.com/scalameta/metals-languageclient/pull/184, let's use it here.

PR #372, opened 2020-08-03
Add files.watcherExclude and search.exclude config

### LINKED ISSUE #365: Consider disabling file watching on some directories for better performance
See https://gitter.im/scalameta/metals?at=5f1c5fafc1c0090e64a16354

It seems that disabling file watching on .bloop, .metals, and target might improve editor responsiveness so we may consider shipping that piece of configuration as a default

gabro:
Closes #365 

/cc @ChristopherDavenport

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -86,7 +86,7 @@ const config = workspace.getConfiguration("metals");
 export async function activate(context: ExtensionContext) {
   detectLaunchConfigurationChanges();
   checkServerVersion();
-  configureGlobalSettingsDefaults();
+  configureSettingsDefaults();
 
   return window.withProgress(
     {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -885,3 +887,33 @@ function isSupportedLanguage(languageId: TextDocument["languageId"]): boolean {
...
+    config.update(
+      propertyKey,
+      { ...currentValues, ...newValues },
+      ConfigurationTarget.Global
+    );
+  }
+  const ignoredMetalsDirs = {
+    "**/.bloop": true,
+    "**/.metals": true,
+    "**/target": true,
```
gabro, line 915: This should cover sbt, is there any other generated directory from other build tools we should include?

src/extension.ts
```diff
@@ -885,3 +887,33 @@ function isSupportedLanguage(languageId: TextDocument["languageId"]): boolean {
...
+  };
+  updateFileConfig("files", "watcherExclude", ignoredMetalsDirs);
+  updateFileConfig("search", "exclude", ignoredMetalsDirs);
```
gabro, line 918: I've limited it to `files.watcherExclude` (avoid file watching, for performance) and `search.exclude` (avoid showing generated files in search results).

I've not added `files.exclude` because that's a broader settings that would also hide those dirs from the file explorer, which I don't think is a good default.

src/extension.ts
ckipp01: `"**/.ammonite"`?

src/extension.ts
```diff
@@ -885,3 +887,33 @@ function isSupportedLanguage(languageId: TextDocument["languageId"]): boolean {
       return false;
   }
 }
+
+// NOTE(gabro): we would normally use the `configurationDefaults` contribution point in the
+// extension manifest but that's currently limited to language-scoped settings in VSCode.
+// We use this method to set global configuration settings such as `files.watcherExclude`
+// and `search.exclude`.
+function configureGlobalSettingsDefaults() {
```
tgodzik, line 895:
```suggestion
function configureWorkspaceSettingsDefaults() {
```
?

src/extension.ts
tgodzik: We could also exclude `out` directory - I think that is used for mill. We could add an if workspace contains `build.sc` only.

src/extension.ts
tgodzik: Good idea! I was worried this could mean `.bloop` would just disappear and cause additional confusion. I had a problem like that previously and I did not fully understood what happened for a while.

src/extension.ts
gabro: These are actually global, i.e. they're scoped to the "User" configuration, since they're fairly general.

tgodzik: Ach, maybe it would be safer to scope it to user? Especially if we include build tool specific excludes?

gabro: it is scoped to "user" right now.

To recap, the scopes are:

- Global (i.e. User in the UI)
- Workspace
- WorkspaceFolder

Also, some settings can be scoped to a specific language (but these specific settings cannot).

Since these are very general, I would lean towards scoping them to Global. My idea is that it makes sense to always ignore `.bloop` / `.metals` / `target` and the like in file watching and search results.

Actually, now that I remembered, it's useless to add `search.exclude` since there's an option in VSCode to automatically exclude files ignored by git.

gabro: Ok, removed now.

Bottom line: as long as we're configure VSCode to exclude very specific files from the file watching such as `.bloop` and `.metals` I think we can safely do that globally.

tgodzik: Ok, I think those 3 plus `.ammonite` should be safe to include. The mill's out is usually not that big of a problem and `out` might be too general.

gabro: I agree 👍 

Are you sure `.ammonite` is used at the project level? I've tried googling and I could only find references for a `~/.ammonite` dir

tgodzik: It's only there if you start a BSP script connection.

gabro: 👍 added!

tgodzik:
Actually, maybe we should not be adding `target` -> we do have things like `target/src-managed`, which will be ignored in this case.

gabro:
What's the harm in ignoring file watching for `src-managed`? We don't use VSCode's file watcher anyway IIRC

tgodzik:
> What's the harm in ignoring file watching for `src-managed`? We don't use VSCode's file watcher anyway IIRC

True, but what about other extensions? If an extension is actually using the VS Code file watcher and target is a normal directory this will cause some unexpected issues. I would maybe only ignore it for a workspace then,

gabro:
I see. That's a good point. I'll ignore `target` for specific workspaces and `.bloop`/`.metals` globally

tgodzik:
> I see. That's a good point. I'll ignore `target` for specific workspaces and `.bloop`/`.metals` globally

Thanks! Sorry, for the my great indecision showed in the comments/approves :sweat:

gabro:
No, thank you for the discussion, those were all legitimate doubts I also had!

PR #375, opened 2020-08-09
Add in extensions and commands from metals-languageclient.

ckipp01:
This goes along with the work in https://github.com/scalameta/metals-languageclient/pull/163. Mainly this takes out quite a bit of shared boilerplate from this extension and coc-metals.


src/extension.ts
```diff
@@ -363,63 +362,58 @@ function launchMetals(
...
+      ServerCommands.AmmoniteStop,
     ].forEach((command) => {
       registerCommand("metals." + command, async () =>
         client.sendRequest(ExecuteCommandRequest.type, { command: command })
       );
     });
 
     let channelOpen = false;
-    const clientCommands: {
-      [k in keyof typeof ClientCommands]: (...args: unknown[]) => unknown;
```
ckipp01, line 384: Since I now have `ClientCommands` as an enum containing all client commands, I had to change this around a bit. It's not as magical now as I just register each one individually, but I couldn't figure out a good way to do this, so I just opted for the straight forward approach.

gabro: I see, although I still think we should make this exhaustive, so that the compiler can help us if we add a new command. Depending on how we encode the commands (see my comment in the upstream PR) we can have slightly different strategies. If we use a literal object:

```typescript
import { pipe } from 'fp-ts/lib/function'
import { record } from 'fp-ts'
import { clientCommands } from 'metals-languageclient'

pipe(clientCommands, record.map((command): void => {
  switch (command) {
    case 'metals-echo-command':
      // ...
    case 'metals-diagnostics-focus':
      // ...
  }
}))
```

This way you have exhaustive matching of commands (and autocompletion) and the compiler will warn if we forget to handle a case.

the `record.map` incantation is needed in order to preserve the precise types (stock object methods such as `Object.keys` or `Object.entries` widen everything to `string`)

gabro: Actually... `Object.entries` seems to at least preserve the narrow types of the values, so the above can become:

```typescript
import { clientCommands } from 'metals-languageclient'

Object.entries(clientCommands).map(([, v]): void => {
  switch (command) {
    case 'metals-echo-command':
      // ...
    case 'metals-diagnostics-focus':
      // ...
  }
})
```

ckipp01: But this will never be exhaustive right? Not all the cases _can_ be handled here. I think that's part of the reason why the `ClientCommands` in protocol here before only held a subset of the actual commands because some are handled down further in `client.onNotifications` as well. So afaik, we can't have every command handled here in a nice fashion can we? Also, a check for exhaustiveness may also not be what we want. VS Code may implement every command, so we'd want that check, however, another client may only account for a subset of those commands. So seemingly the exhaustive check would then not be helpful. What do you think?

gabro: ah, I may not have some context, definitely.

If not all commands can be handled here, then we don't need an exhaustiveness check. The same structure can also work without exhaustiveness (we just need to remove the return type from the lambda and TS won't complain anymore) but that's just up to personal taste at that point.

src/extension.ts
```diff
@@ -541,23 +535,23 @@ function launchMetals(
...
+    registerCommand(`metals.${ServerCommands.ResetChoice}-interactive`, () => {
       client.sendRequest(ExecuteCommandRequest.type, {
-        command: "reset-choice",
+        command: ServerCommands.ResetChoice,
         arguments: [],
       });
     });
 
-    registerCommand("metals.reset-choice", (args) => {
+    registerCommand(`metals.${ServerCommands.ResetChoice}`, (args) => {
```
tgodzik, line 545: Can we actually just check if args is defined, otherwise use empty array? Something like:
`let realArgs = args || []`

We wouldn't need to define it twice and we could drop this weird `metals.${ServerCommands.ResetChoice}-interactive`

src/extension.ts
ckipp01: Sounds good, and yea, since the exhaustive check isn't needed here, I think the individual registers may be the most clear way to do this.

src/extension.ts
ckipp01: Ah, good call. I actually changed this up a bit to instead have a default `[]` parameter. Which testing seems to work fine.

PR #378, opened 2020-08-12
Add deactivate to stop the client once the extension closes

tgodzik:
I've actually notice that we sometimes don't shutdown the server correctly and I did some digging and found that we don't actually have a deactivate function that closes the client explicitly. 

It seems to work most of the time, but maybe adding deactivate might help anyway? Perhaps client wouldn't shutdown properly due to some race condition?

@olafurpg What do you think? Is it just an omission or was it excluded on purpose?

CC @jvican I think this might cause the issue you were talking about.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -73,6 +73,8 @@ const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
 const openSettingsCommand = "workbench.action.openSettings";
 let treeViews: MetalsTreeViews | undefined;
+let currentClient: LanguageClient | undefined;
+
 let decorationType: TextEditorDecorationType = window.createTextEditorDecorationType(
   {
     isWholeLine: true,
@@ -105,6 +107,10 @@ export async function activate(context: ExtensionContext) {
       }
     }
   );
+}
+
+export function deactivate(): Thenable<void> | undefined {
+  return currentClient?.stop();
 }
 
 function showMissingJavaMessage(): Thenable<void> {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -107,6 +109,13 @@ export async function activate(context: ExtensionContext) {
   );
 }
 
+export function deactivate(): Thenable<void> | undefined {
+  if (!currentClient) {
+    return undefined;
+  }
+  return currentClient.stop();
```
gabro, line 116: return currentClient?.stop()
```suggestion
  return currentClient?.stop();
```

src/extension.ts
```diff
@@ -565,7 +575,8 @@ function launchMetals(
         const editor = window.visibleTextEditors.find((e) =>
           isSupportedLanguage(e.document.languageId)
         );
-        if (editor) {
+        if (editor && client) {
+          const definedClient = client;
```
gabro, line 579: you can probably avoid this bind here, and apply the suggestion below

src/extension.ts
```diff
@@ -577,7 +588,7 @@ function launchMetals(
               title: "Metals: Reveal Active File in Side Bar",
             },
             (progress) => {
-              return client
+              return definedClient
```
gabro, line 591:
```suggestion
              return client?.
```

src/extension.ts
tgodzik: Fixed, I haven't noticed I could tidy it up further. Thanks!

PR #383, opened 2020-08-19
Publish to ovsx with explicit --pat argument

tgodzik:
I published 1.9.3 manually to test it out, but we will know for sure during the next release I guess.


tgodzik:
Also move it to the last step so that if it fails we will still have uploaded the binary

tgodzik:
I will merge that since it's only adding --pat, which I tested while publishing locally.

PR #384, opened 2020-08-20
Remove jvmOptions from testClass in json example

kpbochenek:


tgodzik (approved):
Thanks for noticing this! I thought I I removed it already :sweat:

PR #393, opened 2020-09-03
Add in configuration for excludedPackages

ckipp01:
This is a follow-up to https://github.com/scalameta/metals/pull/2012 so that users are able to customize the package exclusions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,6 +118,11 @@
           },
           "markdownDescription": "Optional list of JVM properties to pass along to the Ammonite server. Each property needs to be a separate item.\n\nExample: `-Xmx1G` or `-Xms100M`"
         },
+        "metals.excludedPackages": {
+          "type": "array",
+          "default": [],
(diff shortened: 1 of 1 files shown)


kpbochenek (approved):
LGTM :+1:

(needs to wait for metals release?)

ckipp01:
> LGTM 👍
> 
> (needs to wait for metals release?)

Since we are waiting for other things as well, sure. I don't think it matters _that_ much since we never release the plugin apart fro Metals anyways.

gabro:
> I don't think it matters _that_ much since we never release the plugin apart fro Metals anyways.

It's rare, but it happened in the past and can technically happen again in case of bug fixes or VSCode-specific improvements.

ckipp01:
> It's rare, but it happened in the past and can technically happen again in case of bug fixes or VSCode-specific improvements.

Maybe something to consider is doing a `release`, `development` branch strategy so that we could merge in features as they are added to `development` and then into `release` when we release. That way individuals that want to potentially use certain features in the Metals snapshot can use them if they build themselves. I'm not sure how common it is that someone wants to do that, but it's not the easiest to do that now.

gabro:
It's certainly something that makes sense if we want to support that use case. For the time being I think the changes to the extension are few enough that the current "simplistic" approach still works decently.

PR #403, opened 2020-09-18
Add command to quickly create worksheet

tgodzik:
Depends on https://github.com/scalameta/metals-languageclient/pull/183

It will still ask for name if run from command palette, since it doesn't know the directory. I wonder if we should suggest some default name, or maybe a random one? We are not sure where we are creating the new file, so we will not even be able to check if we can create a new worksheet with the specified file. 

Maybe, we should suggest a default name for worksheets inside Metals server? I think input box has an option for that.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -607,6 +607,34 @@ function launchMetals(
       }
     );
 
+    registerCommand(`metals.new-scala-worksheet`, async () => {
+      const sendRequest = (args: Array<string | undefined>) => {
+        return client.sendRequest(ExecuteCommandRequest.type, {
+          command: ServerCommands.NewScalaFile,
+          arguments: args,
+        });
+      };
+      const currentUri = window.activeTextEditor?.document.uri;
+      if (currentUri != null) {
+        const parentUri = path.dirname(currentUri.toString());
+        const name = path.basename(parentUri);
+        const parentPath = Uri.parse(parentUri).fsPath;
+        const fullPath = path.join(parentPath, `${name}.worksheet.sc`);
+        if (fs.existsSync(fullPath)) {
+          window.showWarningMessage(
+            `A worksheet ${name}.worksheet.sc already exists, opening it instead`
+          );
+          return workspace
+            .openTextDocument(fullPath)
+            .then((textDocument) => window.showTextDocument(textDocument));
+        } else {
+          return sendRequest([parentUri, name, "worksheet"]);
+        }
+      } else {
+        return sendRequest([undefined, undefined, "worksheet"]);
+      }
+    });
+
     registerCommand(`metals.${ServerCommands.NewScalaProject}`, async () => {
       return client.sendRequest(ExecuteCommandRequest.type, {
         command: ServerCommands.NewScalaProject,
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
       "properties": {
         "metals.serverVersion": {
           "type": "string",
-          "default": "0.9.3",
+          "default": "0.9.4",
           "markdownDescription": "The version of the Metals server artifact. Requires reloading the window.\n\n**Change only if you know what you're doing**"
         },
         "metals.serverProperties": {
--- a/package.json
+++ b/package.json
@@ -235,6 +235,11 @@
         "title": "New Scala File..."
       },
       {
+        "command": "metals.new-scala-worksheet",
+        "category": "Metals",
+        "title": "Create Worksheet"
+      },
+      {
         "command": "metals.super-method-hierarchy",
         "category": "Metals",
         "title": "Reveal super method hierachy"


kpbochenek:
I tried with `New Scala File` and it looks like it gets directory when invoked from pallete(it will create new file next to the one I have opened).

When exactly it would ask for a name?

On already existing worksheet this command will fail, correct?

ckipp01:
> It will still ask for name if run from command palette, since it doesn't know the directory. I wonder if we should suggest some default name, or maybe a random one? We are not sure where we are creating the new file, so we will not even be able to check if we can create a new worksheet with the specified file.

Well it doesn't have to right? Since when the user triggers it via the command, you can get the current file they are in, and then grab the parent of that file. That's how I have it working in `coc-metals`.
```js
    registerCommand(`metals.quick-worksheet`, async () => {
      const currentDoc = await workspace.document;
      const currentPath = currentDoc.uri;
      const parentDir = path.dirname(currentPath);
      const name = path.basename(parentDir);

      client.sendRequest(ExecuteCommandRequest.type, {
        command: ServerCommands.NewScalaFile,
        arguments: [parentDir, name, "worksheet"],
      });
    });
```
> On already existing worksheet this command will fail, correct?

Correct

src/extension.ts
```diff
@@ -607,6 +607,18 @@ function launchMetals(
       }
     );
 
+    registerCommand(
+      `metals.${ServerCommands.NewScalaWorksheet}`,
+      async (directory: Uri) => {
+        let defaultName = directory?.path.split(path.sep).pop();
```
ckipp01, line 613:
```suggestion
        const defaultName = directory?.path.split(path.sep).pop();
```

src/extension.ts
```diff
@@ -607,6 +607,18 @@ function launchMetals(
...
+        console.log(defaultName);
```
ckipp01, line 614: Make sure to remove this.

src/extension.ts
```diff
@@ -607,6 +607,31 @@ function launchMetals(
...
+      const currentUri = window.activeTextEditor?.document.uri;
+      if (currentUri) {
+        const parentUri = path.dirname(currentUri.toString());
+        const name = path.basename(parentUri);
+        const parentPath = Uri.parse(parentUri).fsPath;
+        const fullPath = path.join(parentPath, `${name}.worksheet.sc`);
+        if (fs.existsSync(fullPath)) {
+          return workspace
+            .openTextDocument(fullPath)
+            .then((textDocument) => window.showTextDocument(textDocument));
```
tgodzik, line 626: Added an option to open the document if it exists.

ckipp01: Ah, good idea 👍 

I'll probably add this check into `coc-metals` as well.

src/extension.ts
```diff
@@ -607,6 +607,31 @@ function launchMetals(
...
     );
 
+    registerCommand(`metals.new-scala-worksheet`, async () => {
+      const sendRequest = (args: Array<string | undefined>) => {
+        return client.sendRequest(ExecuteCommandRequest.type, {
+          command: ServerCommands.NewScalaFile,
+          arguments: args,
+        });
+      };
+      const currentUri = window.activeTextEditor?.document.uri;
```
ckipp01, line 617: I'm not sure I understand this. When would this return false?

ckipp01 (approved):
Just a quick question, but then LGTM!

src/extension.ts
gabro: if there's no active text editor, I suppose. It can happen if (for instance) VSCode is open but there are no open files

src/extension.ts
```diff
@@ -607,6 +607,31 @@ function launchMetals(
...
+      if (currentUri) {
```
gabro, line 618: nitpick, but since `""` is falsy in JavaScript, I usually make these checks explicit, to avoid weird edge-case behaviors:

```suggestion
      if (currentUri != null) {
```

`!= null` will check for both `null` and `undefined` (different than `!== null`, sigh, JS...)

src/extension.ts
gabro: I'm wondering whether we should also warn the user in this case. Something along the lines of

```ts
window.showWarningMessage(`A worksheet ${name}.worksheet.sc already exists, opening it instead`)
```

gabro (changes requested):
Nice! I've left some minor comments

src/extension.ts
tgodzik: Yes, that's exactly the case.

ckipp01: Sounds good, thanks!

src/extension.ts
tgodzik: Added, I think it makes sense. It shouldn't be too annoying as most people would delete the worksheets afterwards

src/extension.ts

gabro (approved):
LGTM 👍

PR #408, opened 2020-09-21
Run release on publish so that files are properly uploaded

tgodzik:
When we publish a release the push tag event is not matched, so trying out a different event.


tgodzik:
Merging it to test it out on the release.

PR #409, opened 2020-09-21
Update ovsx

tgodzik:
So it seems that without this module we don't even get an error message, but the exact same command seems to be passing locally.

Basically, during the next release we should at least get a real error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -456,7 +456,7 @@
     "@types/node": "14.11.1",
     "@types/shell-quote": "1.7.0",
     "@types/vscode": "1.43.0",
-    "ovsx": "0.1.0-next.9321255",
+    "ovsx": "0.1.0-next.bcce4bc",
     "prettier": "2.1.2",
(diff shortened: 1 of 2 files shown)


tgodzik:
Failed release: https://github.com/scalameta/metals-vscode/runs/1146151220?check_suite_focus=true

gabro:
There's something fishy going on here. Either ovsx is missing a dependency or some its dependencies is. Were you able to track down who requires that module?

gabro:
Ok, this seems to be fixed in the @next version of the package, according to https://github.com/eclipse/openvsx/issues/60

tgodzik:
> Ok, this seems to be fixed in the @next version of the package, according to [eclipse/openvsx#60](https://github.com/eclipse/openvsx/issues/60)

Thanks for checking that out! I thought we have the newest version somehow. Updated it and it seems the errors are shown properly.

gabro (approved):

PR #413, opened 2020-09-28
Support implicit decorations from the Metals server

tgodzik:
This only influences decorations that are inserted inside text. As I checked it does not break worksheets.

Depends on https://github.com/scalameta/metals-languageclient/pull/188

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,14 @@
           "default": [],
           "markdownDescription": "List of packages you'd like to be left out of completions, symbol searches, and code actions.\n\nEx. `akka.actor.typed.javadsl` will ensure nothing in the `javadsl` package gets recommended to you.\n\nYou can find the list of default exclusions [here on the Metals website](https://scalameta.org/metals/docs/editors/new-editor.html#excluded-packages).\n\nIf you need to remove one of the defaults, you can simply include it and preface it with `--`."
         },
+        "metals.showInferredType": {
+          "type": "boolean",
+          "markdownDescription": "When this option is enabled, for each method that have inferred types that are not explicitely specified they are displayed as additional decorations."
+        },
+        "metals.showImplicitArguments": {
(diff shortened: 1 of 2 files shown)


gabro (approved):
👍

kpbochenek:
Should we expose those 2 configuration parameters to the user?

tgodzik:
> Should we expose those 2 configuration parameters to the user?

Yeah, it was there at some point, but disappeared :/

tgodzik:
> Should we expose those 2 configuration parameters to the user?

And fixed! Thanks for noticing!

package.json
```diff
@@ -123,6 +123,14 @@
           "default": [],
           "markdownDescription": "List of packages you'd like to be left out of completions, symbol searches, and code actions.\n\nEx. `akka.actor.typed.javadsl` will ensure nothing in the `javadsl` package gets recommended to you.\n\nYou can find the list of default exclusions [here on the Metals website](https://scalameta.org/metals/docs/editors/new-editor.html#excluded-packages).\n\nIf you need to remove one of the defaults, you can simply include it and preface it with `--`."
         },
+        "metals.typeAnnotationsEnabled": {
+          "type": "boolean",
+          "markdownDescription": "When this option is enabled, for each method that can have infered types they are displayed either as additional decorations if they are supported for the editor or as info diagnostics otherwise."
```
kpbochenek, line 128: Okay so typo 'inferred'
But we can also be more specific, sth like:
`for each method that can have inferred types they are displayed as additional decorations.`

package.json
```diff
@@ -123,6 +123,14 @@
...
+        },
```
kpbochenek, line 129: And a question to 'can have inferred types', when will method have type parameter but it won't be displayed when this is enabled? Ignoring dotty :rofl:

PR #414, opened 2020-10-01
Add env parameter for adding environement variables

tgodzik:


package.json
```diff
@@ -409,6 +409,13 @@
                   "type": "string"
                 },
                 "default": []
+              },
+              "env": {
```
kpbochenek, line 413: [question]
would it make sense to keep it in alignment with how java env are defined? 
use Map[String, String] ?

tgodzik: what do you mean? `jvmOpts` parameter is the same.

kpbochenek: ahh sorry it was mentioned in issue this fixes:
https://code.visualstudio.com/docs/java/java-debugging
https://code.visualstudio.com/docs/java/java-debugging/cwd-env.mp4

kpbochenek: yeah lol I was far from being precise, keep it the same as in java extension that's what I meant xD

tgodzik: We could do it, I will take a look tomorrow.

tgodzik: Added!

package.json
```diff
@@ -409,6 +409,13 @@
...
+                "type": "object",
+                "items": {
+                  "type": "string"
+                },
+                "default": []
```
ckipp01, line 418: Should the default still be `[]` or `{}`?

package.json
```diff
@@ -423,7 +430,8 @@
               "name": "Untitled",
               "mainClass": "???",
               "args": [],
-              "jvmOptions": []
+              "jvmOptions": [],
+              "env": []
```
ckipp01, line 434: Same as above, shouldn't this be `{}` now?

package.json
tgodzik: Fixed!

package.json
tgodzik: And fixed!

ckipp01 (approved):
Looks good! LGTM!

kpbochenek (approved):
LGTM! :+1:

PR #436, opened 2020-11-02
Add in missing `bsp-switch` command

ckipp01:
These ~two command~ command (I just removed the disconnect one) have been in Metals for a while, just not in the extension. So this won't break anything for users sticking on the stable Metals release, but will allow for users to more easily test the sbt BSP support.


tgodzik (approved):
LGTM

PR #437, opened 2020-11-04
Fix homepage which is shown in marketplace

ckipp01:
A colleague pointed this out to me today. When you are looking at the plugin in [Marketplace](https://marketplace.visualstudio.com/items?itemName=scalameta.metals) and click the homepage, it leads you to the old `metals.rocks` url.


tgodzik (approved):
LGTM! Good catch!

PR #442, opened 2020-11-10
Reword the build-restart description

ckipp01:
Ugh, I should have done this before the release but didn't catch it. If a user is using sbt BSP and the trigger this, it does a `workspace/reload`, so it's not just for Bloop anymore. Probably best to update the description to reflect this.


tgodzik (approved):
LGTM

PR #447, opened 2020-11-16
Add option to show implicit methods and classes

tgodzik:
Follow up after https://github.com/scalameta/metals/pull/2232

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,10 @@
           "type": "boolean",
           "markdownDescription": "When this option is enabled, for each method that has implicit arguments they are displayed as additional decorations."
         },
+        "metals.showImplicitConversionsAndClasses": {
+          "type": "boolean",
+          "markdownDescription": "When this option is enabled, each implicit method and class is displayed as additional decorations at the usage site."
+        },
         "metals.javaHome": {
           "type": "string",
           "markdownDescription": "Optional path to the Java home directory. Requires reloading the window.\n\nDefaults to the most recent Java version between 8 and 11 (inclusive) computed by the `locate-java-home` npm package."
--- a/package.json
+++ b/package.json
@@ -304,6 +304,11 @@
         "command": "metals.ammonite-stop",
         "category": "Metals",
         "title": "Stop Ammonite build server"
+      },
+      {
+        "command": "metals.toggle-implicit-conversions-and-classes",
+        "category": "Metals",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -131,6 +131,10 @@
           "type": "boolean",
           "markdownDescription": "When this option is enabled, for each method that has implicit arguments they are displayed as additional decorations."
         },
+        "metals.showImplicitConversionsAndClasses": {
+          "type": "boolean",
+          "markdownDescription": "When this option is enabled, for each method that has implicit arguments they are displayed as additional decorations."
```
ckipp01, line 136: I'm assuming you don't want the copied description from up above? 😆

tgodzik: Argh, I haven't actually saved the file before pushing :sweat_smile:

tgodzik: Fixed!

ckipp01 (approved):
Looks good 👍

tgodzik:
I added an option to toggle the option on and off, which might be useful to users. I will need to do a small change in the server too.

src/extension.ts
```diff
@@ -656,6 +656,14 @@ function launchMetals(
       });
     });
 
+    registerCommand("metals.toggle-implicit-conversions-and-classes", () => {
```
ckipp01, line 659: I'm trying to think if this will be used by any other editors that may use `metals-languageclient`. Do you think we should add this there, or is this probably pretty VS Cod specfic?

tgodzik: I think currently it's pretty much VS Code specific, it could be useful in cases that we show decorations, but emacs does not used the language client

tgodzik: For hovers, it will just start working with the next hover.

src/extension.ts
```diff
@@ -1027,3 +1039,10 @@ function configureSettingsDefaults() {
     ConfigurationTarget.Workspace
   );
 }
+
+function toggleBooleanWorkspaceSetting(setting: string) {
+  const config = workspace.getConfiguration("metals");
+  const configProperty = config.inspect<boolean>(setting);
+  const currentValues = configProperty?.workspaceValue ?? false;
+  config.update(setting, !currentValues, ConfigurationTarget.Workspace);
```
tgodzik, line 1047: @gabro I set it to use workspace by default, I think it makes sense for toggle. What do you think?

PR #457, opened 2020-11-27
Add `isWholeLine` back in case of worksheets 

tgodzik:
This fixes the behaviour when writting at an end of an evaluated worksheet:

image

and now will move the whole decoration right:

image

If we ever support implicits inside worksheets we might need auto update the position to the end of line for worksheet decorations.


gabro (approved):

tgodzik:
I will publish the fix already as the current behaviour can be annoying.

PR #466, opened 2020-12-14
Add in metals.copy-worksheet-ouput command

ckipp01:
This pr adds in the `metals.copy-worksheet-output` command that is needed for https://github.com/scalameta/metals/pull/2290.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -598,6 +598,34 @@ function launchMetals(
       });
     });
 
+    // TODO replace with ServerCommand when done
+    registerTextEditorCommand(
+      `metals.copy-worksheet-output`,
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -598,6 +598,34 @@ function launchMetals(
       });
     });
 
+    // TODO replace with ServerCommand when done
+    registerTextEditorCommand(
+      `metals.copy-worksheet-output`,
```
ckipp01, line 603: - [ ] Replace the hardcoded value with `CopyWorksheetOutput` after `metals-languageclient` is release.

tgodzik (approved):

PR #474, opened 2020-12-19
Revert to an older version of openvsx to fix release issue

tgodzik:
It seems that the recently updated version fails without a clear reason (https://github.com/scalameta/metals-vscode/actions/runs/432894310). Should be safe to just stay on this one for the time being.


ckipp01 (approved):
👍🏼 annoying that we keep having issues with this one. LGTM!

gabro (approved):

PR #492, opened 2021-01-26
Make debug console show up always when running or debugging

tgodzik:
I noticed recently, that the debug console will not show up always when users run or test their application, but turns we can force the view to appear whenever we are successfully starting a DAP session.


gabro (approved):
Nice, it always bugged me but I've never got a chance to fix it! Thanks!

PR #494, opened 2021-01-28
Add `fallback-scala-version` configuration option

dos65:
An addition to scalameta/metals#2417

### DIFF
--- a/package.json
+++ b/package.json
@@ -187,6 +187,29 @@
           "type": "boolean",
           "default": true,
           "markdownDescription": "When enabled, if you press the return key from the first line of a multiline string containing a pipe, it will automatically add `.stripMargin`."
+        },
+        "metals.fallbackScalaVersion": {
+          "type": "string",
+          "default": "automatic",
+          "enum": [
+            "automatic",
+            "2.13.4",
+            "2.13.3",
+            "2.13.2",
+            "2.13.1",
+            "2.13.0",
+            "2.12.13",
+            "2.12.12",
+            "2.12.11",
+            "2.12.10",
+            "2.12.9",
+            "2.12.8",
+            "2.11.12",
+            "3.0.0-M3",
+            "3.0.0-M2",
+            "3.0.0-M1"
+          ],
+          "markdownDescription": "The Scala compiler version that is used as the default or fallback in case a file doesn't belong to any build target or the specified Scala version isn't supported by Metals.\n\nThis applies to standalone Scala files, worksheets, and Ammonite scripts.\n\n The `automatic` value means that the Scala version for these files will be inferred from the highest supported Scala version in your projects build definition"
         }
       }
     },


package.json
```diff
@@ -187,6 +187,28 @@
...
+            "2.12.13",
+            "2.12.12",
+            "2.12.11",
+            "2.12.10",
+            "2.12.9",
+            "2.12.8",
+            "2.11.12",
+            "3.0.0-M3",
+            "3.0.0-M2",
+            "3.0.0-M1"
```
tgodzik, line 209: Let's add at least in the main Metals repo `docs/contributors/releasing.md` information that we should update this list.

dos65: Added to the same PR -  [commit](https://github.com/scalameta/metals/pull/2417/commits/8a9549bd4e0909f26f5e1392f3e64c7a8cfd4621)

package.json
```diff
@@ -187,6 +187,28 @@
...
+          ],
+          "markdownDescription": "The Scala compiler version that is used as default or fallback in case a file doesn't belong to any build target or specified Scala version isn't supported by Metals.\n\nThis applies for standalone Scala files, worksheets, and Ammonite scripts"
```
ckipp01, line 211:
```suggestion
          "markdownDescription": "The Scala compiler version that is used as the default or fallback in case a file doesn't belong to any build target or the specified Scala version isn't supported by Metals.\n\nThis applies to standalone Scala files, worksheets, and Ammonite scripts."
```

dos65: Ough, thanks again!

package.json
```diff
@@ -187,6 +187,28 @@
           "type": "boolean",
           "default": true,
           "markdownDescription": "When enabled, if you press the return key from the first line of a multiline string containing a pipe, it will automatically add `.stripMargin`."
+        },
+        "metals.fallbackScalaVersion": {
+          "type": "string",
+          "default": "",
```
tgodzik, line 193: We should not have any default, otherwise I think empty string will be sent to the server.

tgodzik (changes requested):

package.json
dos65: Thx, I forgot to remove it.

By the way, I remembered that there is a minor issue with this selector.
What do you think about adding an option that will be called `default`?

There is a logic that tries to select the highest Scala version from build targets. 
It applies in case if this property is empty. But `enum` type doesn't allow you to choose empty so there will be no way to clean up this property.

tgodzik: Maybe something like `automatic` ?

dos65: Added `automatic`.

Also mentioned it in the description section. @ckipp01 could you check grammar there one more time?)

ckipp01: Sorry, late to the party on this. I hate to say this now, but what happens if you just don't have a default, what does VS Code send? In some ways, even if it is an empty string, I'd almost rather that, and instead have Metals just filter out an empty string. Mainly because now in this meaningful list of Scala versions we have a random `automatic`. It's not the worst, but just treating `""` as None on the Metals side instead of introducing another value makes more sense to me. It's not a hill I'll die but just my two cents.

dos65: @ckipp01
In VS Code `enum` option can't be manually edited. So `automatic` value solves the problem when you might decide to set `None` value back. Otherwise, you will have to manually edit `.vscode/settings.json`.

It might represented be just `""` but I'm afraid it will more confusing than `automatic` or smth like that.
Unfortunately, there is no way to specify how options should be rendered in VS Code selector. Only plain values.

> Does this mean that if I have set 2.12 set for one module and 2.13 set for another, you'll choose 2.13? I'm just afraid that this may cause confusion

Yes, it will select 2.13. 
I don't think that is confusing because most of the users use a single Scala version per project and will never face the need to change this option.

dos65: For an Ammonite script, it will be Scala 2 because it doesn't support scala 3 yet.
For a worksheet or scala file, it will be Scala 3

package.json
```diff
@@ -187,6 +187,29 @@
...
+            "2.12.11",
+            "2.12.10",
+            "2.12.9",
+            "2.12.8",
+            "2.11.12",
+            "3.0.0-M3",
+            "3.0.0-M2",
+            "3.0.0-M1"
+          ],
+          "markdownDescription": "The Scala compiler version that is used as the default or fallback in case a file doesn't belong to any build target or the specified Scala version isn't supported by Metals.\n\nThis applies to standalone Scala files, worksheets, and Ammonite scripts.\n\n The `automatic` value means that the Scala version for these files might be inferred as the highest supported Scala version from your project build definition"
```
ckipp01, line 212:
```suggestion
          "markdownDescription": "The Scala compiler version that is used as the default or fallback in case a file doesn't belong to any build target or the specified Scala version isn't supported by Metals.\n\nThis applies to standalone Scala files, worksheets, and Ammonite scripts.\n\n The `automatic` value means that the Scala version for these files will be inferred from the highest supported Scala version in your projects build definition"
```

ckipp01 (approved):

package.json
dos65: Thanks!

tgodzik (approved):

PR #505, opened 2021-02-12
Delay notification for slowTask for 2 seconds

### LINKED ISSUE #234: Delay metals/slowTask until task has been running for more than a few seconds
**Describe the bug**

Currently, Metals starts a lot of short-lived window progress notifications when everything is cached:

* "Loading presentation compiler"
* "Downloading Metals"
* "Preparing worksheets"

These tasks can take a few hundred milliseconds to complete when the coursier cache is full. It would be nice if the VS Code extension waited ~2 seconds to start the progress notification

1. Open a worksheet
2. "Reload window"

Screenshot 2020-03-10 at 10 56 23
Screenshot 2020-03-10 at 10 56 16

**Expected behavior**

I expected those progress windows not to appear because they complete in <1 second.

**Installation:**

- Operating system: macOS
- Metals version: 0.8.1

**Additional context**

Docs for metals/slowTask https://scalameta.org/metals/docs/editors/new-editor.html#metalsslowtask

This change can be implemented by delaying `window.withProgress` here
https://github.com/scalameta/metals-vscode/blob/abff1d30fa0354dbe3c2d160039c412269ad3c35/src/extension.ts#L560

It shouldn't be necessary to change anything on the Metals server side.

gourlaysama:
This waits 2 seconds before showing the full progress notification for slow tasks (assuming the task is still running by then).

It avoids multiple notifications like "Loading Presentation Compiler" appearing and immediately leaving because everything was ready and there was almost nothing to do.

Fixes #234

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -69,6 +69,7 @@ import {
   DecorationTypeDidChange,
   DecorationsRangesDidChange,
 } from "./decoration-protocol";
+import { clearTimeout } from "timers";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Thanks for the contribution! Looks good! I will try it out before the next release.

src/extension.ts
```diff
@@ -784,41 +785,52 @@ function launchMetals(
...
-              clearInterval(interval);
-            }
 
-            // Client triggered cancelation from the progress notification.
-            progressToken.onCancellationRequested(() => {
-              onComplete();
-              requestResolve({ cancel: true });
-            });
+        // Wait a bit before showing the progress notification
+        const waitTime = 2;
```
tgodzik, line 790: Let's make sure that if there were already more than 2 seconds elapsed that we show the notification, otherwise we will be always adding it.

```suggestion
        const waitTime = 2;
        let delay = Math.max(0, waitTime - (params.secondsElapsed || 0))
```

src/extension.ts
```diff
@@ -784,41 +785,52 @@ function launchMetals(
...
+                // Server completed long running task.
+                requestToken.onCancellationRequested(() => {
+                  onComplete();
+                  progress.report({ increment: 100 });
+                  setTimeout(() => progressResolve(undefined), 1000);
+                });
+              });
+            }
+          );
+        }, waitTime * 1000);
```
tgodzik, line 828:
```suggestion
        }, delay * 1000);
```

tgodzik (changes requested):
Just checked it out and I would include a small change here. Thanks a lot for the fix!

src/extension.ts
gourlaysama: Good point. Done.

tgodzik (approved):
LGTM! Thanks!

PR #513, opened 2021-03-02
Show errors from launching and fetching Metals

tgodzik:
Not sure if it's the right way to do it in typescript, but it would be good to have more information when something fails. Currently, we can only guess what happened. This should allow the user to at least paste us the error message.

I recommend using `Hide whitespace changes` option.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -232,7 +232,13 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
         javaConfig
       );
     },
-    () => {
+    (reason) => {
+      if (reason instanceof Error) {
+        outputChannel.appendLine(
+          "Downloading Metals failed with the following:"
+        );
+        outputChannel.appendLine(reason.message);
+      }
       const msg = (() => {
         const proxy =
           `See https://scalameta.org/metals/docs/editors/vscode.html#http-proxy for instructions ` +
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -232,7 +232,14 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
         javaConfig
       );
     },
-    () => {
+    (reason) => {
+      if (reason instanceof Error) {
+        const err = reason as Error;
```
gabro, line 237: No need for casting, if you already used `instanceof` (flow typing should understand that)

src/extension.ts
```diff
@@ -361,523 +368,532 @@ function launchMetals(
...
+            editor.setDecorations(worksheetDecorationType, options);
+          else editor.setDecorations(decorationType, options);
+        } else {
+          outputChannel.appendLine(
+            `Ignoring decorations for non-active document '${params.uri}'.`
+          );
+        }
+      });
+    },
+    (err: Error) => {
```
gabro, line 892: if `err` is already of type `Error`, then no need to add the type here:

```suggestion
    (err) => {
```

If not, then we need to check it at runtime (like above)

gabro (changes requested):
Couple of suggestions

src/extension.ts
tgodzik: That's neat, fixed!

src/extension.ts
tgodzik: Ach, right. I added a check.

gabro (approved):

PR #519, opened 2021-03-12
Update coursier bootstrap to 2.0.13

tgodzik:
Follow up from https://github.com/scalameta/metals/pull/2597


tgodzik:
The file is not actually empty, but git keeps showing it is :/

PR #520, opened 2021-03-12
Add commands to run or test the current file or target.

tgodzik:
Thanks to @ckipp01 it's now possible to run test or main classes inside the current file.

The Metals PR that added this functionality is https://github.com/scalameta/metals/pull/2532

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -61,6 +61,8 @@ import {
   MetalsWindowStateDidChange,
   MetalsInputBox,
   MetalsQuickPick,
+  DebugDiscoveryParms,
+  RunType,
 } from "metals-languageclient";
 import * as metalsLanguageClient from "metals-languageclient";
 import { startTreeView } from "./treeview";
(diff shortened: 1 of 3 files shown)


src/scalaDebugger.ts
```diff
@@ -247,3 +247,14 @@ export interface DebugSession {
   name: string;
   uri: string;
 }
+
+export interface DebugDiscoveryParams {
```
tgodzik, line 251: @ckipp01 should we move these classes to the language client?

tgodzik:
It works pretty cool! :heart_eyes: 

https://user-images.githubusercontent.com/3807253/110937048-1cd11500-8332-11eb-893a-b6732f5dd20c.mp4

src/extension.ts
```diff
@@ -562,6 +563,51 @@ function launchMetals(
         }
       });
 
+      registerTextEditorCommand(
+        `metals.run-current-file`,
+        (editor, _edit, _args) => {
+          let args: DebugDiscoveryParams = {
```
gabro, line 569: - 
```suggestion
          const args: DebugDiscoveryParams = {
```

src/extension.ts
```diff
@@ -562,6 +563,51 @@ function launchMetals(
...
+              window.showErrorMessage("Debug session not started");
+            }
+          });
+        }
+      );
+
+      registerTextEditorCommand(
+        `metals.test-current-file`,
+        (editor, _edit, _args) => {
+          let args: DebugDiscoveryParams = {
```
gabro, line 584:
```suggestion
          const args: DebugDiscoveryParams = {
```

src/extension.ts
```diff
@@ -562,6 +563,51 @@ function launchMetals(
...
+              window.showErrorMessage("Debug session not started");
+            }
+          });
+        }
+      );
+
+      registerTextEditorCommand(
+        `metals.test-current-target`,
+        (editor, _edit, _args) => {
+          let args: DebugDiscoveryParams = {
```
gabro, line 599:
```suggestion
          const args: DebugDiscoveryParams = {
```

src/scalaDebugger.ts
ckipp01: Yea probably. I'll open a pr for that. Also, should we put in the other optional fields?

```
 "args"?: string[],
 "jvmOptions"?: string[],
 "env"?: {}, // <-- I actually don't know how this is is done in TS to signify it's an object with unknown k v pairs.
 "envFile"?: string
```

tgodzik: Sure, we can. I haven't added them since they would require using `launch.json`, which I think might not work when running current file.  Also it's kind of weird to set the same options for each file we would run. I think if a user requires arguments etc. then it makes sense to create a dedicated run configuration.

ckipp01: Yea true sounds good.

ckipp01: https://github.com/scalameta/metals-languageclient/pull/248

src/extension.ts

src/extension.ts

PR #525, opened 2021-03-21
Save file before starting debug session

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/524

### DIFF
--- a/src/scalaDebugger.ts
+++ b/src/scalaDebugger.ts
@@ -31,10 +31,13 @@ export async function start(
   noDebug: Boolean,
   ...parameters: any[]
 ): Promise<Boolean> {
-  return vscode.commands
-    .executeCommand<DebugSession>(
-      ServerCommands.DebugAdapterStart,
-      ...parameters
+  return commands
+    .executeCommand("workbench.action.files.save")
+    .then(() =>
+      vscode.commands.executeCommand<DebugSession>(
+        ServerCommands.DebugAdapterStart,
+        ...parameters
+      )
     )
     .then((response) => {
       if (response === undefined) return false;


tgodzik:
@ckipp01 how is this handled in the vim DAP extension?

src/scalaDebugger.ts
```diff
@@ -31,10 +31,13 @@ export async function start(
   noDebug: Boolean,
   ...parameters: any[]
 ): Promise<Boolean> {
-  return vscode.commands
-    .executeCommand<DebugSession>(
-      ServerCommands.DebugAdapterStart,
-      ...parameters
+  return commands
+    .executeCommand("workbench.action.files.save")
```
gabro, line 35: I wonder if we should do this unconditionally, or prompt the user and offer to save.

tgodzik: So it seems that vscode does it automatically anyways when the debug session is run, which means that even with prompting the user the end result will be the same.

gabro: 👍 alright then

gabro (approved):

ckipp01:
> @ckipp01 how is this handled in the vim DAP extension?

I actually haven't checked before, but it seems that if you have unsaved changes in the buffer, it just runs with the last known state. However, that may be quite tricky to change there. Mainly because `nvim-metals` isn't in charge of the actual running at all, just the setup. It basically hands over a couple functions as configuration to `nvim-dap` and then the user triggers the entire thing via `nvim-dap`, not `nvim-metals`. So I'll need to actually poke around `nvim-dap` a bit to see if maybe that could be handled there.

tgodzik:
> > @ckipp01 how is this handled in the vim DAP extension?
> 
> I actually haven't checked before this, but it seems that if you have unsaved changes in the buffer, it just runs with the last known state. However, that may be quite tricky to change there. Mainly because `nvim-metals` isn't in charge of the actual running at all, just the setup. It basically hands over a couple functions as configuration to `nvim-dap` and then the user triggers the entire thing via `nvim-dap`, not `nvim-metals`. So I'll need to actually poke around `nvim-dap` a bit to see if maybe that could be handled there.

I think this is perfectly fine. The main issue would be if it automatically saved the buffers like VS Code, but otherwise if it's totally in the users control then I then we shouldn't change anything.

PR #541, opened 2021-04-19
Update dependencies `metals-languageclient` to v.0.5.0 and `vscode-languageclient` to v.7.0.0

Giggiux:
This PR is needed to close scalameta/metals-feature-requests#202 
It will update `metals-languageclient` to the latest released version
It will update `vscode-languageclient` to support LSP 3.16.0

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -35,7 +35,8 @@ import {
   Location,
   TextDocumentPositionParams,
   TextDocument,
-} from "vscode-languageclient";
+  ExecuteCommandParams,
+} from "vscode-languageclient/node";
 import { LazyProgress } from "./lazy-progress";
 import * as fs from "fs";
 import {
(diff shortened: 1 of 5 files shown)


tgodzik (approved):

src/extension.ts
```diff
@@ -664,7 +669,7 @@ function launchMetals(
                 command: ServerCommands.CopyWorksheetOutput,
                 arguments: [uri.toString()],
               })
-              .then((result) => {
+              .then((result: any) => {
```
gabro, line 672: `any` is dangerous because it basically turns off the typechecker.

Why do we need a cast here? What's the type of `result` if we don't cast it?

gabro (changes requested):
One minor comment

src/extension.ts
Giggiux: Wait, did I change that? 
If I did, it was due to trying some bug fixing and forgetting to remove it, so I believe it can go back to no `any` 😅

tgodzik:
Thanks @gabro for taking a look! I would have missed that one :sweat:

gabro (approved):
LGTM 👍

PR #547, opened 2021-04-23
Update dependencies `metals-languageclient` to v.0.5.0 and `vscode-la…

Giggiux:
…nguageclient` to v.7.0.0

After fixing the issue in scalameta/metals with string's notification support in scalameta/metals#2735, this should now work without errors. Also closes #545.

### DIFF
--- a/package.json
+++ b/package.json
@@ -608,9 +608,9 @@
     "vsce": "1.87.1"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


tgodzik (approved):

tgodzik:
Thanks for digging into this!

PR #572, opened 2021-05-24
Scala3 increase/decrease indentation and keyword completion

Giggiux:
This PR merges [smarter/vscode-dotty-syntax](https://github.com/smarter/vscode-dotty-syntax)'s changes into metals, allowing for scala3's optional braces syntax support with auto increase and decrease in indentation and new keywords completion.

This is still a WIP, but it's fairly stable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -591,7 +591,7 @@
     "vscode:prepublish": "yarn compile",
     "compile": "tsc -p ./",
     "watch": "tsc -watch -p ./",
-    "test": "yarn compile",
+    "test": "jest",
     "build": "vsce package --yarn",
     "vscode:publish": "vsce publish --yarn",
     "ovsx:publish": "ovsx publish",
@@ -599,12 +599,15 @@
     "format-check": "prettier --check '**/*.{ts,js,json,yml}'"
   },
   "devDependencies": {
+    "@types/jest": "^26.0.23",
     "@types/node": "15.12.2",
     "@types/shell-quote": "1.7.0",
     "@types/vscode": "1.43.0",
+    "jest": "^27.0.4",
     "ovsx": "0.1.0",
     "prettier": "2.3.1",
-    "typescript": "4.3.3",
+    "ts-jest": "^27.0.3",
+    "typescript": "4.3.2",
     "vsce": "1.93.0"
   },
   "dependencies": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -72,6 +72,7 @@ import {
   DecorationsRangesDidChange,
 } from "./decoration-protocol";
 import { clearTimeout } from "timers";
+import { decreaseIndentPattern, increaseIndentPattern } from "./indentPattern";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
@@ -1002,10 +1003,9 @@ function enableScaladocIndentation() {
   languages.setLanguageConfiguration("scala", {
     indentationRules: {
       // ^(.*\*/)?\s*\}.*$
-      decreaseIndentPattern: /(^\s*end\b\s*)\b(if|while|for|match|try|\w+)$/,
+      decreaseIndentPattern: decreaseIndentPattern(),
       // ^.*\{[^}"']*$
-      increaseIndentPattern:
-        /(((<!\bend\b\s*?)\b(if|while|for|match|try))|(\bif\s+(?!.*?\bthen\b.*?$)[^\s]*?)|(\b(then|else|do|catch|finally|yield|case))|=|=>|<-|=>>|:)\s*?$/,
+      increaseIndentPattern: increaseIndentPattern(),
     },
     wordPattern:
       /(-?\d*\.\d\w*)|([^\`\~\!\@\#\%\^\&\*\(\)\-\=\+\[\{\]\}\\\|\;\:\'\"\,\.\<\>\/\?\s]+)/g,
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1002,9 +1002,10 @@ function enableScaladocIndentation() {
   languages.setLanguageConfiguration("scala", {
     indentationRules: {
       // ^(.*\*/)?\s*\}.*$
-      decreaseIndentPattern: /^(.*\*\/)?\s*\}.*$/,
+      decreaseIndentPattern: /(^\s*end\b\s*)\b(if|while|for|match|try|\w+)$/,
       // ^.*\{[^}"']*$
-      increaseIndentPattern: /^.*\{[^}"']*$/,
+      increaseIndentPattern:
+        /(((<!\bend\b\s*?)\b(if|while|for|match|try))|(\bif\s+(?!.*?\bthen\b.*?$)[^\s]*?)|(\b(then|else|do|catch|finally|yield|case))|=|=>|<-|=>>|:)\s*?$/,
     },
     wordPattern:
       /(-?\d*\.\d\w*)|([^\`\~\!\@\#\%\^\&\*\(\)\-\=\+\[\{\]\}\\\|\;\:\'\"\,\.\<\>\/\?\s]+)/g,
--- a/src/indentPattern.ts
+++ b/src/indentPattern.ts
@@ -1,0 +1,15 @@
+export function decreaseIndentPattern() {
+  return /((^\s*end\b\s*)\b(if|while|for|match|try|\w+)$|(^(.*\*\/)?\s*\}.*)$)/;
+}
+export function increaseIndentPattern() {
+  const old_if = /\b(if|while)\s+\([^\)]*?\)/;
+  const keywords_not_ending = /((?<!\bend\b\s*?)\b(if|while|for|match|try))/;
+
+  const keywords =
+    /(\b(then|else|do|catch|finally|yield|return|throw))|=|=>|<-|=>>|:/;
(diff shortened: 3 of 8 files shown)


src/extension.ts
```diff
@@ -122,6 +135,80 @@ export function deactivate(): Thenable<void> | undefined {
   return currentClient?.stop();
 }
 
+class KeywordCompletionItemProvider implements CompletionItemProvider {
```
tgodzik, line 138: We already do it in Metals for Scala 2, we need to also implement it for Scala 3, but let's not do it in the vscode extension.

Giggiux: Oks! 
I literally just almost blindly implemented the original repo changes! :D

src/extension.ts
```diff
@@ -79,16 +79,18 @@ const openSettingsCommand = "workbench.action.openSettings";
 let treeViews: MetalsTreeViews | undefined;
 let currentClient: LanguageClient | undefined;
 
-let worksheetDecorationType: TextEditorDecorationType =
-  window.createTextEditorDecorationType({
+let worksheetDecorationType: TextEditorDecorationType = window.createTextEditorDecorationType(
```
tgodzik, line 82: You should be able to run `yarn format ` to fix it (maybe also `yarn install` before to update the prettifier)

src/extension.ts
```diff
@@ -1002,9 +1002,10 @@ function enableScaladocIndentation() {
   languages.setLanguageConfiguration("scala", {
     indentationRules: {
       // ^(.*\*/)?\s*\}.*$
-      decreaseIndentPattern: /^(.*\*\/)?\s*\}.*$/,
+      decreaseIndentPattern: /(^\s*end\b\s*)\b(if|while|for|match|try|\w+)$/,
       // ^.*\{[^}"']*$
-      increaseIndentPattern: /^.*\{[^}"']*$/,
+      increaseIndentPattern:
+        /(((<!\bend\b\s*?)\b(if|while|for|match|try))|(\bif\s+(?!.*?\bthen\b.*?$)[^\s]*?)|(\b(then|else|do|catch|finally|yield|case))|=|=>|<-|=>>|:)\s*?$/,
```
tgodzik, line 1008: Does it cover the previous pattern?

tgodzik (changes requested):

src/extension.ts
```diff
@@ -1002,9 +1002,10 @@ function enableScaladocIndentation() {
   languages.setLanguageConfiguration("scala", {
     indentationRules: {
       // ^(.*\*/)?\s*\}.*$
-      decreaseIndentPattern: /^(.*\*\/)?\s*\}.*$/,
+      decreaseIndentPattern: /(^\s*end\b\s*)\b(if|while|for|match|try|\w+)$/,
```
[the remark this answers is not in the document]
Giggiux: Concatenating them should do the trick, and it *seems* after a quick check that the syntax is ok.

I will implement test in next commits!

Giggiux:
With the latest commits I've implemented the suggested changes, test are implemented, and Tests' CI integration was added.

src/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
+import * as path from "path";
+import Mocha from "mocha";
+import glob from "glob";
+
+export function run(): Promise<void> {
```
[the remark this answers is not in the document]
tgodzik: I don't see any need for this in https://mochajs.org/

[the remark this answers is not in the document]
Giggiux: I believe it's a good compromise because it uses the same "structure" as the [suggested way](https://code.visualstudio.com/api/working-with-extensions/testing-extension) to test vscode extensions.
In the future will be easier to reuse if we want to ever use vscode testing api.
If you don't like this idea, I can change it to use the "not-implementing" mode.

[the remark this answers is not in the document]
tgodzik: I think we can add it when needed. We don't have any plans currently to test the entire extension and it should be easy to add it at that time.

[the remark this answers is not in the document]
Giggiux: Simplified in the latest commit and merged with the latest main version!

src/indentPattern.ts
```diff
@@ -0,0 +1,18 @@
+export function decreaseIndentPattern() {
+  return /((^\s*end\b\s*)\b(if|while|for|match|try|\w+)$|(^(.*\*\/)?\s*\}.*)$)/;
+}
+export function increaseIndentPattern() {
+  // substituted the following "old if syntax" regex from smarter's repo
```
tgodzik, line 5: No need for the comment, it's something related to this PR, but might be confusing in the future, especially someone might not know who is `smarter`

src/test/indentPattern.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  return [checkIndent(indentPattern, true), checkIndent(indentPattern, false)];
+}
+
+suite("IncreaseIndentPattern Test Suite", () => {
+  const [checkIncrease, checkNotIncrease] = checkFunctions(
+    increaseIndentPattern()
+  );
+
+  test("Scala2", () => {
+    checkIncrease("object Main extends App {");
```
tgodzik, line 18: Nice tests! Pretty clear what they are actually testing

src/test/indentPattern.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    checkNotIncrease("if (x < 0) (not possible code)");
+    checkNotIncrease("while (x < 0) (not possible code)");
+
+    // keywords that can end
+    checkIncrease("if");
+    checkIncrease("while");
+    checkIncrease("for");
+    checkIncrease("match");
+    checkIncrease("try");
+    checkIncrease("{some code} try");
```
tgodzik, line 44: Can there be any code before a try ? I think this is not possible, I would remove the two cases.

src/test/indentPattern.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    checkIncrease("    {some code} try");
+    checkNotIncrease("end if");
+
+    // if-then
+    checkIncrease("if (x < 0) then");
+    checkNotIncrease("if (x < 0) then {code}");
+
+    checkIncrease("if (x < 0):");
```
tgodzik, line 52: `if` cannot start with `:`, this might be useful for type ascriptions, but this would not parse anyway.

src/test/indentPattern.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // keywords
+    checkIncrease("{some code} =>");
```
tgodzik, line 55: Do we need `{some code}` everywhere? Especially `if`, `for`, `try` will not compile this way.

src/test/indentPattern.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  test("Scala3", () => {
+    checkDecrease("end if");
+    checkDecrease("end while");
+    checkDecrease("end for");
+    checkDecrease("end match");
+    checkDecrease("end try");
+    checkDecrease("end Class");
+    checkNotDecrease("end Something Else");
```
tgodzik, line 94:
```suggestion
    checkNotDecrease("end Something");
```
can't have two words for a name.

tgodzik (changes requested):
Some minor comments, but great work! I would remove some test cases that are not a parseable Scala code.

src/test/indentPattern.test.ts
Giggiux: There are some examples where the code will compile in those three cases:

```scala
  val b = for
    el <- list
  yield el
  ```
  
  ```scala
  val c: Int = if 
    a.nonEmpty 
  then 
    2
  else 
    1
```

```scala
val d = try
  println(a)
  catch
    case e =>
      println(e)
```

tgodzik (approved):
LGTM from my side! @gabro anything typescript wise that we could be doing better here?

package.json
```diff
@@ -591,17 +591,22 @@
...
     "format": "prettier --write '**/*.{ts,js,json,yml}'",
     "format-check": "prettier --check '**/*.{ts,js,json,yml}'"
   },
   "devDependencies": {
+    "@types/glob": "^7.1.3",
+    "@types/mocha": "^8.2.2",
     "@types/node": "15.12.2",
     "@types/shell-quote": "1.7.0",
     "@types/vscode": "1.43.0",
+    "glob": "^7.1.7",
```
gabro, line 608: is this used? I don't find an import

package.json
```diff
@@ -591,17 +591,22 @@
...
+    "mocha": "^9.0.0",
```
gabro, line 609: minor, but I would prefer using `jest` since we already use it in metals-languageclient.

gabro (changes requested):
Nice work, just a couple of minor things

package.json
Giggiux: In one of the changes it was removed, I forgot to clean the package.json after! 
Removed it! :D

package.json
Giggiux: I implemented using jest, can you double check everything looks ok? 
It's the first time I configured it!

gabro: Looks great, thanks!

gabro (approved):
👍 LGTM, but @Giggiux I think you need to rebase to resolve the conflicts

gabro (approved):
👍 LGTM, but @Giggiux I think you need to rebase to resolve the conflicts

PR #587, opened 2021-06-09
Update building Node version to v14

Giggiux:
Node.js v10 Active support ended the 19th of May 2020, the Security Support ended the 30th of April 2021.
the current latest LTS is Node 14 with an end of Security Support in Apr 2023.

The changes in this PR should just update the various Github's Actions to use the latest Node.js 14 version (currently 14.17.0). I tried to build the extension locally with this Node.js version and everything looked ok.

Updating to this version also allows for simpler updates to dependencies, that will slowly losing support for Node.js versions <12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          - node-version: "14"
+          node-version: "14"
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM although I am not sure what that would imply. @gabro could you confirm this is ok :sweat_smile: ?

gabro (approved):
LGTM

PR #603, opened 2021-07-01
serverProperties: Add default xmx value

dos65:
I'm not sure if it should be 512m or 1G (512 is enough for metals and scalameta) but I think it anyway should be specified.

Having multiply metals processes without `Xmx` setting  sometimes causes problems  because they use default value (25% of MaxRam) and they consume more memory than required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,9 @@
           "items": {
             "type": "string"
           },
+          "default": [
+            "-Xmx512m"
+          ],
           "markdownDescription": "Optional list of properties to pass along to the Metals server. By default, the environment variable `JAVA_OPTS` and `.jvmopts` file are respected. Each property needs to be a separate item.\n\nExample: `-Dhttps.proxyHost=…`, `-Dhttps.proxyPort=…` or `-Dmetals.statistics=all`"
         },
         "metals.ammoniteJvmProperties": {


package.json
```diff
@@ -109,6 +109,9 @@
           "items": {
             "type": "string"
           },
+          "default": [
+            "-Xmx512m"
```
tgodzik, line 113:
```suggestion
            "-Xmx1024m"
```
is what I found optimal, so let's maybe not set it too low?

tgodzik (changes requested):
@olafurpg opinions? Should we set `Xmx` by default?

I think this might be useful as people do not usually know it's useful to set it and might think that Metals is always using way to much memory.

package.json
dos65: Tried `512m` additionally on akka/akka and apache/spark.
The real memory usage is about `200m` so I think `512` fits better for default.

olafurpg: 512 should be enough for even large repos unless there have been added new indexes recently that hold onto a lot of memory.

olafurpg:
This is a good idea!

olafurpg:
It might be worth considering adding logic that regularly checks memory usage and reports instructions on how to increase memory when necessary. I think both IntelliJ and eclipse have something like that.

gabro (approved):

tgodzik (approved):
> It might be worth considering adding logic that regularly checks memory usage and reports instructions on how to increase memory when necessary. I think both IntelliJ and eclipse have something like that.

Let's add a feature request to track it, but otherwise LGTM!

PR #629, opened 2021-07-20
Offer to install Java in case of issues

tgodzik:
Also, made a fix to only update the workspace config if that one was in fact invalid. This might be helpful in case of remote workspaces, where globally updating the config will break the local settings.

It would be ebst in those cases to update the remote setting, but it doesn't seem to be possible currently.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -76,7 +76,10 @@ import { decreaseIndentPattern, increaseIndentPattern } from "./indentPattern";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
+const downloadJava = "Download Java";
 const openSettingsCommand = "workbench.action.openSettings";
+const installJava8Action = "Install Java (JDK 8)";
+const installJava11Action = "Install Java (JDK 11)";
 let treeViews: MetalsTreeViews | undefined;
 let currentClient: LanguageClient | undefined;
 
@@ -124,9 +127,6 @@ export function deactivate(): Thenable<void> | undefined {
 }
 
 function showMissingJavaMessage(): Thenable<void> {
-  const installJava8Action = "Install Java (JDK 8)";
-  const installJava11Action = "Install Java (JDK 11)";
-
   const message =
     "Unable to find a Java 8 or Java 11 installation on this computer. " +
     "To fix this problem, update the 'Java Home' setting to point to a Java 8 or Java 11 home directory " +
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -266,17 +285,25 @@ function fetchAndLaunchMetals(context: ExtensionContext, javaHome: string) {
...
-        if (choice === openSettingsAction) {
-          commands.executeCommand(openSettingsCommand);
-        }
-      });
+      window
+        .showErrorMessage(msg, openSettingsAction, downloadJava)
+        .then((choice) => {
+          if (choice === openSettingsAction) {
+            commands.executeCommand(openSettingsCommand);
+          } else if (choice == downloadJava) {
```
gabro, line 293:
```suggestion
          } else if (choice === downloadJava) {
```

gabro (approved):
One super minor thing, but otherwise LGTM

src/extension.ts
tgodzik: Didn't notice that one, thanks!

dos65 (commented):
I've seen that there is a usage o `locate-java-home` library inside the `getJavaConfig` and we use it to check the version of JDK. I'm wondering why it doesn't fail if there is no JDK installation.
https://github.com/scalameta/metals-vscode/blob/3c4e460ef20ff6504e1d8a4e3cbcae8e20a0abcc/src/extension.ts#L231-L236

tgodzik:
> I've seen that there is a usage o `locate-java-home` library inside the `getJavaConfig` and we use it to check the version of JDK. I'm wondering why it doesn't fail if there is no JDK installation.
> https://github.com/scalameta/metals-vscode/blob/3c4e460ef20ff6504e1d8a4e3cbcae8e20a0abcc/src/extension.ts#L231-L236

It does fail, but not if we manually configured java-home or java has a wrong version. I this case it might fail while trying to start coursier or Metals. In that case it's nice to give the users a possibility to download it automatically I think.

dos65 (approved):
> It does fail, but not if we manually configured java-home or java has a wrong version. I this case it might fail while trying to start coursier or Metals. In that case it's nice to give the users a possibility to download it automatically I think.

Yeah, I think it's good fallback option.

Btw, I found the original issue where user said that he has JDK 16 but it looks like extension shouldn't accept it (according [to this line](https://github.com/scalameta/metals-languageclient/blob/main/src/getJavaHome.ts#L51)). We either use this library improperly or there is a bug.

Also, I wasn't clear that it's just question.
LGTM!

PR #644, opened 2021-08-06
Point the message about java home to to exact Metals setting

tgodzik:
Following up from https://github.com/scalameta/metals/issues/2020#issuecomment-891360215


dos65 (approved):
LGTM

PR #649, opened 2021-08-17
Set super method code lenses to true by default

tgodzik:
Thanks to https://github.com/scalameta/metals/pull/3045 we no longer need to calculate the super methods and just use the existing semanticdb information.

I don't think we need to confirm it since we already use a lot of different info from semanticdb and it doesn't cause any additional CPU usage.

CC @tanishiking

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,7 +183,7 @@
         },
         "metals.superMethodLensesEnabled": {
           "type": "boolean",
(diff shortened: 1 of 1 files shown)


dos65 (approved):

tanishiking (approved):
🚀

PR #660, opened 2021-08-30
display build target info

Arthurm1:
see https://github.com/scalameta/metals/pull/3380

requires https://github.com/scalameta/metals-languageclient/pull/331

### DIFF
--- a/src/goToLocation.ts
+++ b/src/goToLocation.ts
@@ -1,5 +1,6 @@
 import { workspace, window, ViewColumn, Uri, Range } from "vscode";
 import { DocumentUri, Range as LspRange } from "vscode-languageclient/node";
+import { MetalsFileProvider } from "./metalsContentProvider";
 
 export interface WindowLocation {
   uri: DocumentUri;
@@ -7,7 +8,10 @@ export interface WindowLocation {
   otherWindow: boolean;
 }
 
-export function gotoLocation(location: WindowLocation): void {
+export function gotoLocation(
+  location: WindowLocation,
+  metalsFileProvider: MetalsFileProvider | undefined = undefined
+): void {
   const range = new Range(
     location.range.start.line,
     location.range.start.character,
@@ -25,7 +29,11 @@ export function gotoLocation(location: WindowLocation): void {
         )
         .pop()?.viewColumn || ViewColumn.Beside;
   }
-  workspace.openTextDocument(Uri.parse(location.uri)).then((textDocument) =>
+  let uri = Uri.parse(location.uri);
+  // vscode will cache the virtual documents even after closing unless onDidChange is fired
+  if (uri.scheme == "metalsDecode" && metalsFileProvider)
+    metalsFileProvider.onDidChangeEmitter.fire(uri);
+  workspace.openTextDocument(uri).then((textDocument) =>
     window.showTextDocument(textDocument, {
       selection: range,
       viewColumn: vs,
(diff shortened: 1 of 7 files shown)


src/treeview.ts
```diff
@@ -191,6 +191,7 @@ class MetalsTreeDataProvider implements TreeDataProvider<string> {
           ? Uri.parse(item.nodeUri)
           : undefined,
       collapsibleState: toTreeItemCollapsibleState(item.collapseState),
+      contextValue: item.contextValue,
```
tgodzik, line 194: Is this needed in the protocol or can be handle it on the vs code side?

tgodzik, line 194: MAybe:

```suggestion
      contextValue: if (item.nodeUri.startsWith(projects:) && item.nodeUri.contains("_root_")) "projects" else undefined,
```
Something along these lines? (though that is Scala code :sweat_smile: )

Arthurm1: You suggestion worked.  I've made the change.  Adding contextValue to the API is no longer needed.

tgodzik:
Languageclient has been released https://github.com/scalameta/metals-languageclient/releases/tag/v0.5.8 :tada:

src/goToLocation.ts
```diff
@@ -1,13 +1,17 @@
...
 export interface WindowLocation {
   uri: DocumentUri;
   range: LspRange;
   otherWindow: boolean;
 }
 
-export function gotoLocation(location: WindowLocation): void {
+export function gotoLocation(
+  location: WindowLocation,
+  metalsFileProvider: MetalsFileProvider | undefined = undefined
```
kpodsiad, line 13:
```suggestion
  metalsFileProvider?: MetalsFileProvider
```
should have the same effect

kpodsiad, line 13:
```suggestion
  metalsFileProvider?: MetalsFileProvider
```
should have the same effect

src/goToLocation.ts
```diff
@@ -25,7 +29,12 @@ export function gotoLocation(location: WindowLocation): void {
         )
         .pop()?.viewColumn || ViewColumn.Beside;
   }
-  workspace.openTextDocument(Uri.parse(location.uri)).then((textDocument) =>
+  const uri = Uri.parse(location.uri);
+  // vscode will cache the virtual documents even after closing unless onDidChange is fired
+  if (uri.scheme == "metalsDecode" && metalsFileProvider) {
+    metalsFileProvider.onDidChangeEmitter.fire(uri);
+  }
+  workspace.openTextDocument(uri).then((textDocument) =>
```
kpodsiad, line 37: I wonder why is that a bad thing?

Arthurm1: Because the output of the virtual docs can change.  If you run javap and then change the source and re-run it should refresh the output.  Also - if you run `Display Target` as Metals is indexing then it won't know the list of source jars yet so it will say that all jar dependencies are missing sources.  If you re-run after it's got the source jars then it will now show updated build info.

kpodsiad: got it, thanks

tgodzik (changes requested):
Let's bump to newest snapshot once https://github.com/scalameta/metals/actions/runs/1806111245 is released and resolve conflicts.

kpodsiad:
I bumped snapshot version in https://github.com/scalameta/metals-vscode/pull/836 - there is no need to do this here anymore.

kpodsiad (approved):
Awesome stuff @Arthurm1!
I wanted this feature so badly that I've resolved conflict myself, hope you don't mind ;)

Arthurm1:
@kpodsiad No problem - thanks

PR #667, opened 2021-09-02
Increase default Xmx to 1G

dos65:
Unfortunately, `512m` might be not enough for now because Scala3 PC requires much more memory that Scala 2.


tgodzik (approved):
LGTM

PR #670, opened 2021-09-06
Add Java Formatting and New File command

Arthurm1:
Depends on: https://github.com/scalameta/metals/pull/2520

Adds 3 user config settings for Java file formatting

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,6 +279,19 @@
             "3.0.0"
           ],
           "markdownDescription": "The Scala compiler version that is used as the default or fallback in case a file doesn't belong to any build target or the specified Scala version isn't supported by Metals.\n\nThis applies to standalone Scala files, worksheets, and Ammonite scripts.\n\n The `automatic` value means that the Scala version for these files will be inferred from the highest supported Scala version in your projects build definition"
+        },
+        "metals.eclipseFormatConfigPath": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -279,6 +279,19 @@
             "3.0.0"
           ],
           "markdownDescription": "The Scala compiler version that is used as the default or fallback in case a file doesn't belong to any build target or the specified Scala version isn't supported by Metals.\n\nThis applies to standalone Scala files, worksheets, and Ammonite scripts.\n\n The `automatic` value means that the Scala version for these files will be inferred from the highest supported Scala version in your projects build definition"
+        },
+        "metals.eclipseFormatConfigPath": {
```
tgodzik, line 283: Could we somehow squash those settings into one object maybe ? Or perhaps could the other two settings be included in the actual config?

It seems weird that there are settings that people need to set manually.

package.json
```diff
@@ -279,6 +279,14 @@
...
+          "type": "string",
+          "markdownDescription": "Optional custom path to the eclipse-formatter.xml file.\n\nShould be an absolute path and use forward slashes `/` for file separators (even on Windows)."
+        },
+        "metals.eclipseFormatProfile": {
```
tgodzik, line 287: If we still need `eclipseFormatProfile` could we keep it together in an object? It will be easier to find for the users if the two are together.

PR #675, opened 2021-09-08
add javap/semanticdb file viewer

Arthurm1:
See https://github.com/scalameta/metals/pull/3107

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,32 @@
     "workspaceContains:*/*/src/main/scala"
   ],
   "contributes": {
+    "languages": [
+      {
+        "id": "javap-verbose",
+        "extensions": [
+          ".javap-verbose"
+        ]
+      },
+      {
+        "id": "javap",
+        "extensions": [
+          ".javap"
+        ]
+      }
+    ],
+    "grammars": [
+      {
+        "language": "javap-verbose",
+        "scopeName": "source.jvm-bytecode-verbose",
+        "path": "./syntaxes/jvm-bytecode-verbose.json"
+      },
+      {
+        "language": "javap",
+        "scopeName": "source.jvm-bytecode",
+        "path": "./syntaxes/jvm-bytecode.json"
+      }
+    ],
     "configurationDefaults": {
       "[scala]": {
         "editor.suggestSelection": "first",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -470,6 +495,16 @@
           "command": "metals.new-scala-file",
           "when": "metals:enabled",
           "group": "navigation@4"
+        },
+        {
+          "command": "metals.show-javap-verbose",
```
tgodzik, line 500: We could probably have a submenu called `Analyse Scala file` or something along those lines? 

https://code.visualstudio.com/api/references/contribution-points#contributes.menus

Arthurm1:
@tgodzik 
I can either shift **all** Metals commands to a sub-menu...

image

Or just the analyze...

image
Do you have a preference?

tgodzik:
> Do you have a preference?

I would say the second one.

Arthurm1:
1. Added submenu
2. Reinstated verbose options
3. User can now call decode commands on the output document so it can be refreshed when data has changed

Arthurm1:
What's the situation with licenses?

I want to include the jvm-bytecode syntaxes to make the javap output much more readable.  They are distributed under the MIT license and metals-vscode under the Apache license.  Are they compatible?  Do I just commit the MIT license along with the syntaxes?

Conversation on license here: https://github.com/mnxn/vscode-jvm-bytecode-viewer/issues/2

mnxn:
(I wrote the syntaxes)

Apache and MIT are compatible provided you follow the license terms.
I would be satisfied with a NOTICE.md like in the metals repository: https://github.com/scalameta/metals/blob/main/NOTICE.md

tgodzik (changes requested):
Great work! We should also add the license as suggested.

Arthurm1:
I think I've addressed everything

I gave `registerTextEditorCommand` a go but it would mean creating a duplicate set of commands to work off the file explorer context menu so I left then as `registerCommand`

The enabling/disabling is a lot cleaner now.

syntaxes/jvm-bytecode.json
```diff
@@ -0,0 +1,395 @@
+{
+  "$schema": "./schema/tmlanguage.json",
```
mnxn, line 2: The schema this references is from https://github.com/martinring/tmlanguage.
You will have to add its license to NOTICE.md too if you want to include it in this repository.

Alternatively, you can just remove the schema file and the `$schema` field from this and the other syntax and it would work fine. I only included it in my original project because it makes it easier to edit the syntax.

Arthurm1: Thanks for the heads-up @mnxn

Arthurm1:
Included `tmLanguage Schema` license

tgodzik (changes requested):
Looks good! We just need to update the languageclient version

Arthurm1:
Done

tgodzik (approved):
LGTM!

Arthurm1:
Changed to pass `DecoderResponse` instead of JSON
```
export interface DecoderResponse {
  requestedUri?: string;
  value?: string;
  error?: string;
}
```

tgodzik (approved):
Let's wait for the next release!

tgodzik:
I will go ahead and merge this one since we need this PR in two different ones, so my plan is to merge the current changes and I will release from a custom tag if changes are needed.

PR #678, opened 2021-09-09
Revert automatic indentation for optional braces

tgodzik:
It turns out that we need to have the same number of indents as deindents, otherwise when we write `end m...` it will increase indentation n - 1, where n was the last indentation that was increased. In order to make it work properly we would also need to deindent without an end marker.

This changes the behaviour to only indent for onEnterRules, which does not work the same and does not require deindent.

I also added an additional rule so that no automatic indentation is made when we press enter before an expression.

The problem this is fixing is:
scala3-example-project-1631218361758

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -72,7 +72,7 @@ import {
   DecorationsRangesDidChange,
 } from "./decoration-protocol";
 import { clearTimeout } from "timers";
-import { decreaseIndentPattern, increaseIndentPattern } from "./indentPattern";
+import { increaseIndentPattern } from "./indentPattern";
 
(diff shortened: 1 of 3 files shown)


dos65 (approved):

tgodzik:
We could add something similar to:

https://github.com/microsoft/vscode-python/blob/3698950c97982f31bb9dbfc19c4cd8308acda284/src/client/language/languageConfiguration.ts#L37

But it doesn't seem super needed and I might follow up later on.

tgodzik:
@dos65 I changed the behaviour to indent still but only in the `onEnterRules`, which doesn't have the same problem as the previous approach.

dos65 (approved):

PR #681, opened 2021-09-12
Make Java Home setting machine-scoped

aldum:
Recently, I started using VSCode(ium) on Windows, which necessitated manually specifying the Java home setting in the extension settings. This was all fine and dandy, until I turned back to my Linux box, only to find it being broken by the settings synchronization. To my great satisfaction, it turned out to be already a supported to restrict the effect of a setting to machine-local: https://code.visualstudio.com/updates/v1_34#_machinespecific-settings
I tested this with Codium 1.60 across 2 Linux (Artix) and 1 Windows (10) hosts, having satisfying results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,7 @@
         },
         "metals.javaHome": {
           "type": "string",
+          "scope": "machine",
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Thanks for finding this!

PR #689, opened 2021-09-16
Add Symbol-search command

dos65:
Fixes scalameta/metals#2346.

The build-in symbols search in vscode doesn't work well as it filters out items which label doesn't match query. Beucase of that queries with `;` (include dependencies) and short forms like `j.n.f.Path` haven't worked.

output

### DIFF
--- a/package.json
+++ b/package.json
@@ -381,6 +381,11 @@
         "command": "metals.test-current-target",
         "category": "Metals",
         "title": "Run all test in the current project"
+      },
+      {
+        "command": "metals.symbol-search",
+        "category": "Metals",
+        "title": "Search symbol in workspace"
       }
     ],
     "menus": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -74,6 +74,8 @@ import {
 import { clearTimeout } from "timers";
 import { increaseIndentPattern } from "./indentPattern";
 import { TastyResponse } from "./executeCommand";
+import { gotoLocation } from "./goToLocation";
+import { openSymbolSearch } from "./openSymbolSearch";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
@@ -867,6 +869,9 @@ function launchMetals(
         });
       });
 
+      // NOTE: we offer a custom symbol search command to work around the limitations of the built-in one, see https://github.com/microsoft/vscode/issues/98125 for more details.
+      registerCommand(`metals.symbol-search`, () => openSymbolSearch(client));
+
       window.onDidChangeActiveTextEditor((editor) => {
         if (editor && isSupportedLanguage(editor.document.languageId)) {
           client.sendNotification(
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -1198,3 +1204,114 @@ function toggleBooleanWorkspaceSetting(setting: string) {
   const currentValues = configProperty?.workspaceValue ?? false;
   config.update(setting, !currentValues, ConfigurationTarget.Workspace);
 }
+
+class SymbolItem implements QuickPickItem {
```
gabro, line 1208: can we move all this custom symbol search stuff into another file? It would easier to remove if/when the issue is fixed upstream

gabro (changes requested):
I wish we didn't need to do this, but this is a fair workaround.

Before we merge, I've realized this was never reported properly upstream, so I've just done it here https://github.com/microsoft/vscode/issues/133313

Can you link to that issue in a comment, with a quick explanation of why we have a custom symbol search?

dos65:
@gabro Actually there is an issue about that. Sorry that I didn't mentioned it - https://github.com/microsoft/vscode/issues/98125 . 
According do discussion, it would require changes in protocol and I don't think that it will be fixed soon

gabro:
ah, thanks for the pointer, I've marked that issue as duplicate

src/extension.ts
dos65: Done!

src/extension.ts
```diff
@@ -828,6 +830,8 @@ function launchMetals(
         });
       });
 
+      registerCommand(`metals.symbol-search`, () => openSymbolSearch(client));
```
gabro, line 833: I would still add a comment here referencing the upstream issue, so that we remember why we need this. Something like

```suggestion
      // NOTE: we offer a custom symbol search command to work around the limitations of the built-in one, see https://github.com/microsoft/vscode/issues/98125 for more details.
      registerCommand(`metals.symbol-search`, () => openSymbolSearch(client));
```

gabro (changes requested):
Just a minor thing, then 👍

package.json
```diff
@@ -370,6 +370,11 @@
         "command": "metals.test-current-target",
         "category": "Metals",
         "title": "Run all test in the current project"
+      },
+      {
+        "command": "metals.symbol-search",
+        "category": "Metals",
+        "title": "Search symbol"
```
gabro, line 377:
```suggestion
        "title": "Search symbol in workspace"
```

?

gabro (approved):
:shipit:

gabro:
@dos65 you may need to rebase to fix the conflict

tgodzik:
What do you think about releasing it today? I wanted to add the fix for `{}` and we can just quickly mention that on scalameta.

gabro:
👍 sounds good, vscode releases are cheap

dos65:
@tgodzik :+1:

PR #692, opened 2021-09-17
Allow braces to be split with an added indentation

tgodzik:
Before the change:

after

after:

before


gabro (approved):
I'm a simple person, I see a regex with a nice GIF, I click approve :shipit:

dos65 (approved):

PR #693, opened 2021-09-17
Prefill symbol search with the word under cursor, if any

gabro:
This mimics a behavior of the default symbol search which I've always found immensely useful.

When there's an active text editor, the quick pick is prefilled with the word under cursor (provided there's an active text editor and there's actually a word under the cursor).

Demo:


tgodzik (approved):
LGTM!

gabro:
With this in place, I feel more comfortable in replacing it as the default search command. I've added some docs on how to do it here: https://github.com/scalameta/metals/pull/3139

PR #694, opened 2021-09-18
Override default symbol search keybinding

dos65:
It seems that priority rule for binding selection in case of clash is: `User defined > Extension > Default`.
So this override works if `ctrl+t` is default and it doesn't touch it if user set something else on this shortcut


tgodzik:
Not 100% sure about this one, what do other people thing about it?

PR #707, opened 2021-09-24
Add position to show tasty command

kpodsiad:
Changes needed by https://github.com/scalameta/metals/pull/3148
Also contains changes from https://github.com/scalameta/metals-vscode/pull/675

PR #710, opened 2021-09-27
Tasty quick pick

kpodsiad:
Contains https://github.com/scalameta/metals-vscode/pull/707


src/extension.ts
```diff
@@ -797,12 +797,40 @@ function launchMetals(
...
+            textDocument: { uri: editor.document.uri.toString() },
+            position: null,
+          });
+        } else {
+          const { positions, items } = await executeCommand<AvailableClasses>(
+            client,
+            `metals.show-available-classes`,
+            { uri: editor.document.uri.toString() }
+          );
+          if (items.length === 1) {
```
gabro, line 817: Would it be possible to keep this logic on the server?
We have `metals/quickPick` which the server can use to open a quickPick prompt on the editor

kpodsiad: Although it's possible at the current moment, in the near future I want to reimplement show tasty using `decompile file` command and virtual documents instead of webview. AFAIK then it won't be possible to do what you propose, I'll try to explain why.
Let `Main.scala` be:
```scala
package a.b.c
class Foo 
class Bar 
```
Currently, we will create the following URI: `metalsDecode:file:///somePath/Main.scala.tasty?decoder=tasty` but this URI is ambiguous because it doesn't contain any information on which class from this file should be shown. And you'' have opened the new document with URI like this `file:///somePath/Main.scala.tasty` and it won't be possible to open the second document for second class without interrupting the first one. That's why we need some knowledge a priori to be able to extend URI with some information about the class so the URIs could look like e.g. `file:///somePath/Main.scala.Foo.tasty` and `file:///somePath/Main.scala.Bar.tasty`

But I'm not an expert at virtual documents and the way they work, maybe I'm missing something.
@Arthurm1 maybe you could help us. Is there any possibility to solve the above issue without asking the server before?

Arthurm1: I may be mistaken but I think @gabro is just asking you to move the quickpick functionality from VSCode to the server so that interaction with non-vscode clients is easier.  So instead of `metals.show-available-classes` you could have something like `metals.choose-class`?  You can see how this is done in the `new-scala-file` command.

One thing I would say about the decode file stuff is that I setup the javap and semanticdb so that you could re-run the same command (e.g. decode javap verbose) on the javap verbose output file.  So the user could keep switching between source and output files while fixing whatever issue they have. So it would be handy to be able to run `show tasty` on a URI with the class name already in it e.g. run it on `file:///somePath/Main.scala.Foo.tasty` so that it can be refreshed.  Not sure if you handle this?

I like that you've added the class name to the URI - looks like the class name finding stuff was a bit painful.

kpodsiad: Thanks to you, @Arthurm1, I was able to understand what @gabro meant (clearly my brain has some problems with thinking after midnight 😃 )

gabro: Hey, yes, thanks @Arthurm1 for clarifying, I just meant that the logic of showing the quick pick based on the number of results could be managed by the server, so that clients only need to implement the `metals/quickPick` endpoint but they don't have to worry to replicate additional logic specifically for this command

PR #715, opened 2021-09-30
Add eslint to project

kpodsiad:


.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    {
+      files: ["*.ts"],
+      rules: {
+        "@typescript-eslint/no-explicit-any": "off",
+        "no-useless-escape": "off",
+        "no-case-declarations": "off",
+        "guard-for-in": "error",
+        "no-var": "error",
+        "curly": "error",
+        // quotes: ["error", "single"]
```
kpodsiad, line 21: The question is `single`, `double` or `backtick`?

kpodsiad: @gabro what do you think?

gabro: I don't think this must be managed by eslint at all. For all formatting rules let's use prettier

gabro (commented):
Thanks @kpodsiad! I have many comments to this, but I won't be able to look at it for a few days, unless we're in a rush can we please hold on to it for a bit? Two general comments I can leave now:
- let's not use eslint for formatting rules (quotes, braces, etc). Let's stick to prettier for it, which is more appropriate, and to with whatever default it has
- as suggested on discord, I would start smaller, I already see many rules with questionable benefits that require us to turn them off selectively. It's a small codebase with relatively few contributors, I think we can afford going little by little

gabro (commented):
Another couple of suggestions for improving dx:
- setting up recommended extensions for the workspace, so that new contributors opening the project will get a prompt suggesting to install them
- setting up lint-staged to run prettier and eslint on changed files as a pre commit hook
- enforce eslint on the CI

kpodsiad:
Thanks for your feedback @gabro 
I'll add todo list containing your suggestions, but one of them needs further discussion. We aren't in rush so don't worry and take your time ;)

> * let's not use eslint for formatting rules (quotes, braces, etc). Let's stick to prettier for it, which is more appropriate, with whatever default it has

Yes, you are right with quotes, it could be managed by prettier. However, braces should be managed by eslint, [see this comment](https://github.com/prettier/prettier/issues/7659#issuecomment-590004436).

> * as suggested on discord, I would start smaller, I already see many rules with questionable benefits that require us to turn them off selectively. It's a small codebase with relatively few contributors, I think we can afford going little by little

I started with just recommended rules, I disabled only 3 rules so it's not that many. I think it's better to stick with the recommended ones and disable the rules which don't suit us than another way, it's shorter for sure. 

> * setting up lint-staged to run prettier and eslint on changed files as a pre commit hook

Pre-commit hook could very annoying, especially when developing sth locally and frequently committing local changes. For now, I wouldn't add it.

kpodsiad:
> Ok, I'm still not super sold on enforcing formatting rules with eslint because it's trickier to setup a "format on save" workflow. In order to fix those errors on save you would need to enable code actions on save in VS Code, but that setting also fixes a bunch of other non-formatting stuff, so it's not something we would necessarily want to recommend.

@gabro Do you think that lack of "fix on save" action is problematic? `ESLint` extension shows errors immediately and offers the possibility to quick-fix them. I think, in the terms of this particular rule which is `curly`, what ESLint does is good enough. Especially when we'll take into account amount of people who will be developing metals extension and fact that inserting braces is a little bit more than just formatting ;)

PR #719, opened 2021-10-08
Implement VSCode side of 'find text in dependency JARs'

Z1kkurat:
metals-find-in-files-1

This is the VSCode-side implementation of https://github.com/scalameta/metals/pull/3093.  
The treeview is created separately from the base Metals container because it's updated manually in a different way than other metals' treeviews. It also uses native vscode's 'Search' icon as proposed by @tgodzik (I would paint something like metals logo + magnifying glass, but I'm too bad at all artist' things, sorry 😞).

### DIFF
--- a/src/findinfiles.ts
+++ b/src/findinfiles.ts
@@ -1,5 +1,4 @@
 import {
-  Event,
   EventEmitter,
   ExtensionContext,
   OutputChannel,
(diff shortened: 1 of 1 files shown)


Z1kkurat:
@ckipp01 AFAIK, vscode's user API for that is very limited (or absent). Last time I checked, you can't just invoke that references popup from anywhere, meaning that you need to be a some file with a cursor at some point of the text.  
I'm talking about `editor.action.peekLocations` and `editor.action.goToLocations` commands from [here](https://code.visualstudio.com/api/references/commands), their usage is also [advised](https://github.com/microsoft/vscode/issues/100904) by VSCode devs. Maybe I'm wrong or missing something, but I failed to find the API to invoke 'peek references' or 'find references' window with some custom content from anywhere in the editor and not from a specific point in file.

PR #751, opened 2021-11-08
Fix revert

tgodzik:


tgodzik:
The previous PR did not actually pass, since the condition for mergify was wrong.

PR #755, opened 2021-11-09
Implement basics of test API

kpodsiad:
~There are some compilation errors which will be fixed in https://github.com/scalameta/metals-vscode/pull/754~ - done and merged

After the brainstorm, together with @tgodzik we concluded that using DAP Event to pass some structured data about test results (well-defined JSON) will be the best and the fastest approach. 
For now, mentioned data is passed through an OutputEvent, more specifically as a `testResult` event which contains all necessary information about the whole suite and at the same time about every single test.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -36,6 +36,7 @@ import {
   ExecuteCommandRequest,
   Location,
   CancellationToken,
+  CodeLensRefreshRequest,
 } from "vscode-languageclient/node";
 import { LazyProgress } from "./lazy-progress";
 import * as fs from "fs";
@@ -64,6 +65,7 @@ import {
   MetalsQuickPick,
   DebugDiscoveryParams,
   RunType,
+  TestUIKind,
 } from "metals-languageclient";
 import * as metalsLanguageClient from "metals-languageclient";
 import { startTreeView } from "./treeview";
@@ -83,7 +85,8 @@ import {
 } from "./findinfiles";
 import * as ext from "./hoverExtension";
 import { decodeAndShowFile, MetalsFileProvider } from "./metalsContentProvider";
-import { getTextDocumentPositionParams } from "./util";
+import { getTextDocumentPositionParams, getValueFromConfig } from "./util";
+import { createTestManager } from "./test-explorer/test-manager";
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
 const downloadJava = "Download Java";
@@ -359,6 +362,7 @@ function launchMetals(
     slowTaskProvider: true,
     statusBarProvider: "on",
     treeViewProvider: true,
+    testExplorerProvider: true,
   };
 
   const clientOptions: LanguageClientOptions = {
@@ -610,6 +614,33 @@ function launchMetals(
         { scheme: "file", language: "scala" },
         codeLensRefresher
       );
+
+      const getTestUI = () =>
+        getValueFromConfig<TestUIKind>(
+          config,
+          "testUserInterface",
+          "Test Explorer"
+        );
+
+      const istTestManagerDisabled = getTestUI() === "Code Lenses";
+      const testManager = createTestManager(client, istTestManagerDisabled);
+
+      const disableTestExplorer = workspace.onDidChangeConfiguration(() => {
+        const testUI = getTestUI();
+        if (testUI === "Code Lenses") {
+          testManager.disable();
+        } else {
+          testManager.enable();
+        }
+      });
+
+      const refreshTests = client.onRequest(CodeLensRefreshRequest.type, () => {
+        testManager.discoverTestSuites();
+      });
+
+      context.subscriptions.push(disableTestExplorer);
+      context.subscriptions.push(refreshTests);
+      context.subscriptions.push(testManager.testController);
 
       // Handle the metals/executeClientCommand extension notification.
       client.onNotification(ExecuteClientCommand.type, (params) => {
--- a/src/test-explorer/test-manager.ts
+++ b/src/test-explorer/test-manager.ts
@@ -1,0 +1,175 @@
+import * as vscode from "vscode";
+import { TestRunProfileKind, tests } from "vscode";
+import {
+  ExecuteCommandRequest,
+  LanguageClient,
+} from "vscode-languageclient/node";
+import { testCache } from "./test-cache";
+import { runHandler } from "./test-run-handler";
+import {
+  SuiteDiscovery,
+  TargetName,
+  TargetUri,
+  TestDiscoveryResult,
+  TestItemMetadata,
+} from "./types";
+import { toVscodeRange } from "./util";
+
+export function createTestManager(
+  client: LanguageClient,
+  isDisabled: boolean
+): TestManager {
+  return new TestManager(client, isDisabled);
+}
+
+class TestManager {
+  readonly testController = tests.createTestController(
+    "metalsTestController",
+    "Metals Test Explorer"
+  );
+
+  private isDisabled = false;
+  private isRunning = false;
+
+  constructor(private readonly client: LanguageClient, isDisabled: boolean) {
+    if (isDisabled) {
+      this.disable();
+    }
+    this.testController.resolveHandler = async (item?: vscode.TestItem) => {
+      if (item != null) {
+        return;
+      } else {
+        await this.discoverTestSuites();
+      }
+    };
+
+    const callback = () => (this.isRunning = false);
+
+    this.testController.createRunProfile(
+      "Run",
+      TestRunProfileKind.Run,
+      (request, token) => {
+        if (!this.isRunning) {
+          this.isRunning = true;
+          runHandler(this.testController, true, callback, request, token);
+        }
+      },
+      true
+    );
+
+    this.testController.createRunProfile(
+      "Debug",
+      TestRunProfileKind.Debug,
+      (request, token) => {
+        if (!this.isRunning) {
+          this.isRunning = true;
+          runHandler(this.testController, false, callback, request, token);
+        }
+      },
+      false
+    );
+  }
+
+  async enable(): Promise<void> {
+    this.isDisabled = false;
+    await this.discoverTestSuites();
+  }
+
+  /**
+   * Disables test manager, also it deletes all discovered test items in order to hide gutter icons.
+   */
+  disable(): void {
+    this.testController.items.forEach((item) =>
+      this.testController.items.delete(item.id)
+    );
+    this.isDisabled = true;
+  }
+
+  discoverTestSuites(): Promise<void> {
+    if (this.isDisabled) {
+      return Promise.resolve();
+    }
+
+    return this.client
+      .sendRequest(ExecuteCommandRequest.type, {
+        command: "discover-test-suites",
+      })
+      .then(
+        (value: SuiteDiscovery[]) => {
+          for (const { targetName, targetUri, discovered } of value) {
+            const rootNode = this.testController.createTestItem(
+              targetName,
+              targetName
+            );
+            createTestItems(
+              this.testController,
+              discovered,
+              rootNode,
+              targetName,
+              targetUri
+            );
+            const data: TestItemMetadata = {
+              kind: "project",
+              targetName,
+              targetUri,
+            };
+            testCache.setMetadata(rootNode, data);
+            this.testController.items.add(rootNode);
+          }
+        },
+        (err) => console.error(err)
+      );
+  }
+}
+
+/**
+ * Create TestItems from the given @param discoveredArray and add them to the @param parent
+ */
+function createTestItems(
+  testController: vscode.TestController,
+  discoveredArray: TestDiscoveryResult[],
+  parent: vscode.TestItem,
+  targetName: TargetName,
+  targetUri: TargetUri
+) {
+  for (const discovered of discoveredArray) {
+    if (discovered.kind === "suite") {
+      const { className, location, fullyQualifiedName } = discovered;
+      const parsedUri = vscode.Uri.parse(location.uri);
+      const parsedRange = toVscodeRange(location.range);
+      const testItem = testController.createTestItem(
+        fullyQualifiedName,
+        className,
+        parsedUri
+      );
+      testItem.range = parsedRange;
+      const data: TestItemMetadata = {
+        kind: "suite",
+        targetName,
+        targetUri,
+      };
+      testCache.setMetadata(testItem, data);
+      parent.children.add(testItem);
+    } else {
+      const data: TestItemMetadata = {
+        kind: "package",
+        targetName,
+        targetUri,
+      };
+
+      const packageNode = testController.createTestItem(
+        `${parent.id}.${discovered.prefix}`,
+        discovered.prefix
+      );
+      parent.children.add(packageNode);
+      testCache.setMetadata(packageNode, data);
+      createTestItems(
+        testController,
+        discovered.children,
+        packageNode,
+        targetName,
+        targetUri
+      );
+    }
+  }
+}
--- a/src/test-explorer/test-run-handler.ts
+++ b/src/test-explorer/test-run-handler.ts
@@ -1,0 +1,174 @@
+import { ServerCommands } from "metals-languageclient";
+import * as vscode from "vscode";
+import {
+  CancellationToken,
+  commands,
+  TestController,
+  TestRunRequest,
+} from "vscode";
+import { debugServerFromUri, DebugSession } from "../scalaDebugger";
+import { analyzeTestRun } from "./analyze-test-run";
+import { testCache } from "./test-cache";
+import { DapEvent, TargetUri, TestItemMetadata } from "./types";
+
+// this id is used to mark DAP sessions created by TestController
+// thanks to that debug tracker knows which requests it should track and gather results
+export const testRunnerId = "scala-dap-test-runner";
+
+/**
+ * Register tracker which tracks all DAP sessions which are started with @constant {testRunnerId} kind.
+ * Dap sends execution result for every suite included in TestRun in a special event of kind 'testResult'.
+ * Tracker has to capture these events and store them all in map under debug session id as a key.
+ */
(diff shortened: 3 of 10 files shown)


src/test-provider/output-analyzer.ts
```diff
@@ -0,0 +1,72 @@
+export interface TestRunActions {
+  failed?(testId: string, time: number, message: string): void;
+  passed?(testId: string, time: number): void;
+  skipped?(testId: string): void;
+}
+
+// e.g. Execution took (90ms|0.1s|etc.)
+const timeRegex = /Execution took (\d?\.?\d+)(ms|s)/;
```
olafurpg, line 8: Did you consider getting this information via BSP instead of parsing it from stdout? The stdout format is unstable and could change easily.

kpodsiad: Are you referring to the [test notifications](https://build-server-protocol.github.io/docs/specification.html#test-notifications-1)? I thought I could use `buildTarget/test` to run tests but AFAIK it runs all tests from the given projects or at least the whole test suites, depending on implementation. Also `The server's implementation decides the granularity at which tests are reported` so it seemed to me like too many cases have to be handled.
If you're not referring to the test notifications please elaborate more, because maybe I missed something. Can I obtain this kind of information from BSP when I'm using DAP to run tests? Does BSP take part in communication after the DAP session is started? I thought that both answers are no, but if they're affirmative I will happily get rid off regexes :D

tgodzik: The question is whether we can actually get that data from the test runners :thinking: I quickly checked and what we are currently doing is just sending back line by line via the output notification. @adpi2 do you know if it would be possible to have some kind of test harness to report success/failure/elapsed time ?

olafurpg: This information is part of the sbt testing framework so it should be possible to forward it via BSP. We don’t need to use buildTarget/test since it doesn’t support debugging. Instead, the build server can just push notifications for the test events.

adpi2: Yes, I agree with @olafurpg, we don't need to use the `buildTarget/test` and we can push test notifications during the `debugSession\start` session but it is not yet implemented. In sbt test-agent we have something called [`ForkEnvent`](https://github.com/sbt/sbt/blob/1f4517d5cc5f6417c24fc3478819d41b548e1ba6/testing/agent/src/main/java/sbt/ForkMain.java#L446-L482) that could be forwarded as [BSP test notifications](https://build-server-protocol.github.io/docs/specification.html#test-notifications-1) to Metals. (It looks like the `TestStart` notification is missing in `ForkEvent` currently, but it could be added).

tgodzik: BSP test notifications are a bit problematic, since they are not part of DAP and it would require Metals forward them as LSP notifications that would update a mutable state for the test session to use. Not sure if we would not get any issue from that approach, especially if anyone runs a couple of tests at the same time (we wouldn't be able to tell which session should be updated).

I wonder whether it would be ok to just to use output event with the `data` field containing json with the information from the TestEvent. This would all be done in the same session and would be simple to handle on all possible frontends.

https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output

src/test-provider/analyze-test-run.ts
```diff
@@ -0,0 +1,43 @@
...
+  const results = new Map(resultsTuples);
+  for (const test of tests) {
+    const suiteName = test.id;
+    const result = results.get(suiteName);
+    if (result != null) {
+      const duration = result.duration;
+      const failed = result.tests.filter(failedGuard);
+      if (failed.length > 0) {
+        const msg = failed
+          .map((t) => ansicolor.strip(t.error))
```
kpodsiad, line 32: `ansicolor.strip` because error msg doesn't suppor ansicolor

src/test-provider/analyze-test-run.ts
```diff
@@ -0,0 +1,43 @@
+import * as vscode from "vscode";
+import {
+  Failed,
+  SingleTestResult,
+  TestRunActions,
+  TestSuiteSummary,
+} from "./types";
+import { ansicolor } from "ansicolor";
+
+function failedGuard(result: SingleTestResult): result is Failed {
```
tgodzik, line 10: I am not familiar with this syntax, what does it do?

src/test-provider/types.ts
```diff
@@ -0,0 +1,94 @@
...
+  kind: "failed";
+  testName: string;
+  duration: number;
+  error: string;
+}
+
+/**
+ * Describes actions from vscode.TestRun
+ */
+export interface TestRunActions {
```
tgodzik, line 86: Why do we need an additional interface here?

tgodzik (commented):
Great work! I still have some questions, especially that I am not an expert in TS.

tgodzik:
@gabro would you be able to take a look? Ignore me if you are busy. :sweat_smile:

gabro:
@tgodzik hey, I'm on vacation until tomorrow, but I can try taking a look on Monday!
In case it's blocking a release don't mind waiting for me :)

tgodzik:
It's not blocking at all. I am just starting to think about the new release. Thanks!

src/test-provider/analyze-test-run.ts
kpodsiad: It's a filter function that also narrows the type of the kept elements, `result is Failed` is a `type predicate` and can be treated as a special form of a boolean. [See more about type rpedicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

src/test-provider/types.ts
kpodsiad: It'll allow testing of `analyzeTestRun` function without the need of `vscode.TestRun` instance. I want to add them later when test explorer will have single test granularity. 

I could remove it until it'll be really needed, but I was too lazy :P

src/util.ts
```diff
@@ -1,10 +1,18 @@
...
 import { LanguageClient } from "vscode-languageclient/node";
 
+declare const sym: unique symbol;
+/**
+ * Creates a newtype without any runtime overhead. It's important for ID to be both unique and descriptive.
+ */
+export type newtype<A, ID extends string> = A & {
+  readonly [sym]: ID;
+};
+
```
kpodsiad, line 15: I know it's a little bit magical to the person who is not familiar with the typescript type system & magic. However, these types created by `newtype` produce compile errors and have meaningful names instead of just `string`.

src/test-provider/util.ts
```diff
@@ -0,0 +1,11 @@
+import { Range } from "vscode-languageclient/node";
+import * as vscode from "vscode";
+
+export function toVscodeRange(range: Range): vscode.Range {
```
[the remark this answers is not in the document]
kpodsiad: I don't know :(
One of them is just an interface (you can think about it as a typed JSON without any runtime overhead) and the second one is a class (classes have runtime overhead). VS code testing API expects the second.

It could be dangerous if you mix those two. For example, some internals might expect class and try to invoke a function on it. However, because you used wrong type in you interface Typescript didn't produce compiler error and bum, you got runtime error :(

src/util.ts
kpodsiad: It seems like it's too much for poor Tomek :D

src/test-provider/test-manager.ts
```diff
@@ -0,0 +1,327 @@
...
+  const queue = createRunQueue(request);
+
+  for (const { test, data } of queue) {
+    if (token.isCancellationRequested) {
+      run.skipped(test);
+    } else {
+      run.started(test);
+      const children = getTestItemChildren(test);
+      children.forEach((c) => run.enqueued(c));
+      await commands
```
gabro, line 106: I would prefer not mixing async/await and the `then` api of promises if possible.
Also in general this expression is very long and I think it would benefit in readability if broken down in multiple smaller expressions

gabro (commented):
Really nice job, I've left a first pass of stylistic comments

src/extension.ts
```diff
@@ -608,6 +611,37 @@ function launchMetals(
...
+        async () => {
+          const testUI = getValueFromConfig<TestUIKind>(
+            config,
+            "testUserInterface",
+            "Code Lenses"
+          );
+          if (testUI === "Code Lenses") {
+            testManager.disable();
+          } else {
+            await testManager.enable();
```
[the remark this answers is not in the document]
kpodsiad: I think it's a habit to await every function which returns a promise, but here and below it is indeed not necessary

src/test-provider/test-manager.ts
kpodsiad: Ok, so followng is a good way to go?
```typescript
try {
  const debugSession = await commands...
  const wasStarted = await ...
  await analyzeResults()
} catch(error) {
}
```

src/extension.ts
```diff
@@ -608,6 +611,35 @@ function launchMetals(
...
+      const istTestManagerDisabled =
+        getValueFromConfig<TestUIKind>(
+          config,
+          "testUserInterface",
+          "Code Lenses"
+        ) === "Code Lenses";
+      const testManager = createTestManager(client, istTestManagerDisabled);
+
+      const disableTestExplorer = workspace.onDidChangeConfiguration(() => {
+        const testUI = getValueFromConfig<TestUIKind>(
```
tgodzik, line 623: Could this be a function and reused in istTestManagerDisabled ?

src/extension.ts
```diff
@@ -608,6 +611,35 @@ function launchMetals(
         codeLensRefresher
       );
 
+      const istTestManagerDisabled =
+        getValueFromConfig<TestUIKind>(
+          config,
+          "testUserInterface",
+          "Code Lenses"
```
tgodzik, line 618: Should we have the default value set to Code Lenses?

src/test-explorer/test-run-handler.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Register tracker which tracks all DAP sessions which are started with @constant {testRunnerId} kind.
+ * Dap sends execution result for every suite included in TestRun in a special event of kind 'testResult'.
+ * Tracker has to capture these events and store them all in map under debug session id as a key.
+ */
+vscode.debug.registerDebugAdapterTrackerFactory("scala", {
+  createDebugAdapterTracker(session) {
+    if (session.configuration.kind === testRunnerId) {
+      return {
+        onWillStartSession: () => testCache.setEmptySuiteResultsFor(session.id),
```
tgodzik, line 27: Can we run multiple tests at the same time?

src/test-explorer/test-manager.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.isDisabled = false;
+    await this.discoverTestSuites();
+  }
+
+  /**
+   * Disables test manager, also it deletes all discovered test items in order to hide gutter icons.
+   */
+  disable(): void {
+    this.testController.items.forEach((item) =>
+      this.testController.items.delete(item.id)
```
tgodzik, line 71: Should we also empty the TestCache?

src/test-explorer/test-manager.ts
```diff
@@ -0,0 +1,163 @@
...
+function createTestItems(
+  testController: vscode.TestController,
+  discoveredArray: TestDiscoveryResult[],
+  parent: vscode.TestItem,
+  targetName: TargetName,
+  targetUri: TargetUri
+) {
+  for (const discovered of discoveredArray) {
+    if (discovered.kind === "suite") {
+      const { className, location, fullyQualifiedName } = discovered;
```
tgodzik, line 125: Could this be typed ?

src/test-explorer/test-manager.ts
```diff
@@ -0,0 +1,163 @@
...
+        className,
+        parsedUri
+      );
+      testItem.range = parsedRange;
+      const data: TestItemMetadata = {
+        kind: "suite",
+        targetName,
+        targetUri,
+      };
+      testCache.setMetadata(testItem, data);
```
tgodzik, line 139: Why do we need the cache actually?

tgodzik (commented):
Some questions from me, excuse me if they don't make sense :sweat_smile:  I am a typescript amateur.

src/extension.ts
kpodsiad: We could default to the `Test Explorer`, no problem.

src/test-explorer/test-manager.ts
kpodsiad: `vscode.TestItem` doesn't contain any additional field for metadata. It's basically just an id, label, location and children. It's better to keep additional & necessary data elsewhere. Maybe the `cache` is not the best word for it, but I cannot think of something better.

src/test-explorer/test-manager.ts
kpodsiad: Actually, there is no point in doing that. `this.testController.items.forEach` iterates only over top-level test items which are projects. In order to delete everything we will have to extract all the children from projects and it's not worth it.
Old test items will be eventually GC'ed or even replaced by the newest ones.

src/util.ts
gabro: I'm quite familiar with this encoding of newtypes, but I think the unique symbol should be the value, otherwise it's quite useless, the definition above is equivalent to:

```ts
export type newtype<A, ID extends string> = A & {
  readonly theType: ID;
};
```

This way the uniqueness is only guaranteed by the string passed when defining the type, which can work ok, but you are not using the unique symbol.

An alternative encoding can just be:

```ts
export type Newtype<A, O> = A & O

type UserId = Newtype<number, { readonly UserId: unique symbol }>
type ThingId = Newtype<number, { readonly ThingId: unique symbol}>
```

Full example here: 

https://www.typescriptlang.org/play?ssl=1&ssc=13&pln=1&pc=20#code/KYDwDg9gTgLgBDAnmYcBywDuSUB4CCANHAPIB8cAvHPnAGSkBQjOqAqgM7BQCSAJlXRZWuAHYBXALYAjbsQDecKMACGfCKIA2iOJ278AXHHGiAlgEdxqDohkRNcAL5kWyVABUAFqdEBzfoIY2G5iUrJQCkqq6lo6Xj7+fEYmFlZwNnaazsx8wADGmirKcABmJnkwphqlABTiXLxJug38AJRGAG4QpnwA3Iy5BUWoeRoc8PX6TXqN-YOFxaOi4wjefoZw8et9zCU1MGuJrYx7k42tQA

(this is loosely based on https://github.com/gcanti/newtype-ts/blob/master/src/index.ts)

Also (suuuper tiny detail) this is not strictly a newtype, since a `TargetName` can be passed where a `string` is expected. The encoding in newtype-ts is a bit stricter in this sense, that's why I've relaxed it a bit

kpodsiad: >I'm quite familiar with this encoding of newtypes, but I think the unique symbol should be the value, otherwise it's quite useless, the definition above is equivalent to:

Actually, I used that in this way intentionally. The symbol is declared as a unique one and is used as a key for unnamed property. Moreover, the symbol is not exported so only a newtype can use it. Thanks to that you cannot access this property or even fake it. What do I mean by "fake it"? Let's consider the following example:
```typescript
type newtype<A, ID extends string> = A & {
  readonly theType: ID;
};
type SuiteName = newtype<string, "suiteName">;
type Type2 = string & {
  theType: "suiteName";
};

const x: Type2 = "aezakmi" as Type2;
const y: Type1 = x; // works
```
I know that this is an almost never gonna happen scenario, but I try to be careful with structural typing.

>Also (suuuper tiny detail) this is not strictly a newtype, since a TargetName can be passed where a string is expected. The encoding in newtype-ts is a bit stricter in this sense, that's why I've relaxed it a bit

Yeah, you're right. But this is more than fine in this case since I wanted to use "newtype" to distinguish between string types and for us, the name `newtype` shouldn't be misleading.

src/extension.ts
```diff
@@ -611,6 +615,33 @@ function launchMetals(
...
+        if (testUI === "Code Lenses") {
+          testManager.disable();
+        } else {
+          testManager.enable();
+        }
+      });
+
+      const refreshTests = client.onRequest(CodeLensRefreshRequest.type, () => {
+        testManager.discoverTestSuites();
+      });
```
kpodsiad, line 639: Test explorer should be refreshed just like code lense. This change is possible due to https://github.com/scalameta/metals/pull/3355

src/test-explorer/test-manager.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+    this.testController.resolveHandler = async (item?: vscode.TestItem) => {
+      if (item != null) {
+        return;
+      } else {
+        await this.discoverTestSuites();
+      }
+    };
+
+    const callback = () => (this.isRunning = false);
```
kpodsiad, line 46: I experimented a little bit and it seems like this + vscode built-in cancelation mechanism seems to work (more or less). When you are in the middle of test run and you'll click to run another test site `this.isRunning` will prevent you from running the next suite. However, the first request is cancelled and is marked as a skipped/cancelled by vscode even though test-analyzer sets particular test items status. I think it's good enough for start and if this will be a problem we can spend more time on this mechanic. Ok @tgodzik ?

src/util.ts
gabro: Thanks for the explanation, it make sense then! One minor thing is that not using a unique symbol in the value too means that you can accidentally do:

type Type1 = newtype<string, "type1">;
type Type2 = newtype<string, "type1">; // oops

and those two will be silently interchangeable.

It's probably ok if we don't have many cases, it's a mistake that can be caught in code review

kpodsiad:
I squashed commits to a single one with a little bit more description.
@tgodzik @gabro have you got any comments left? I would like to put the "waiting for a metals release" label on this PR as fast as possible :D

src/test-explorer/test-run-handler.ts
tgodzik: Actually, could we not allow it? Sorry for not wording it clearer. Is there anything that could break on multiple sessions?

tgodzik (approved):
This will be awesome!

PR #767, opened 2021-11-24
Add Debuggers label

tgodzik:
The label is basis for https://code.visualstudio.com/docs/editor/debugging where debuggers are dynamically queried.

It's probably never going to pop up there (I think it's the msot popular extensions), but it's better to add the label.


gabro (approved):

dos65 (approved):
:+1:

PR #770, opened 2021-11-26
Replace config generation with automatic run of the current file

tgodzik:
Previously, when no launch.json was configured we would show a chain of quick picks so that the user would guided through configuring their entry. This however, would not be obvious to the users and would show up only once at the start. This was also not too obvious for the bginners and prompted https://github.com/scalameta/metals/issues/3272

Now, we remove the config generation and replace it by mechanism to detect if no launch.json is available and then run the command to run anything in the current file.

We might return partly to config generation when we have a mechanism to discover the the mains and tests, which can then be used to suggest new entries to launch.json.

Depends on https://github.com/scalameta/metals/pull/3311

### DIFF
--- a/src/scalaDebugger.ts
+++ b/src/scalaDebugger.ts
@@ -6,19 +6,23 @@ import {
   ProviderResult,
   WorkspaceFolder,
   DebugAdapterDescriptor,
+  DebugConfigurationProviderTriggerKind,
 } from "vscode";
-import { ServerCommands } from "metals-languageclient";
+import {
+  DebugDiscoveryParams,
+  RunType,
+  ServerCommands,
+} from "metals-languageclient";
 
 const configurationType = "scala";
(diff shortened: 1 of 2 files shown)


src/scalaDebugger.ts
```diff
@@ -6,19 +6,23 @@ import {
   ProviderResult,
   WorkspaceFolder,
   DebugAdapterDescriptor,
+  DebugConfigurationProviderTriggerKind,
 } from "vscode";
-import { ServerCommands } from "metals-languageclient";
+import {
+  // DebugDiscoveryParams,
+  // RunType,
```
tgodzik, line 13: If https://github.com/scalameta/metals/pull/3311 gets merged I will update `RunType` enum

gabro (approved):

kpodsiad (approved):
LGTM!

Issue #784, opened 2021-12-12
Worksheet evaluations only appear in a single view of worksheet


stewSquared:
**Describe the bug**

When splitting an editor and viewing the same worksheet in different editor groups, worksheet evaluations only appear in one editor. This makes it difficult to work with two views of the same worksheet, as one will have intellisense, but no evaluations.

If the editor showing evaluations is closed, no evaluations will be shown for the worksheet until restarting altogether, and the option to cancel an ongoing evaluation stops popping up (write an infinite loop, and you have to restart vscode).

Note that this also happens when an editor is split "in group" (single editor instance of a single group, but only view split rather than editor reproduced into another group.

**To Reproduce** Steps to reproduce the behavior:

1. open a worksheet
2. execute `> View: Split Editor Right` or `> View: Split Editor in Group`
3. allow worksheet evaluation to complete
4. evaluations are only shown in one view

**Expected behavior**

Evaluations are show in both views

**Screenshots**

![two editors of same worksheet in different groups](https://user-images.githubusercontent.com/4022813/145733084-4ba701ee-2c43-403f-a764-bf005d16f146.png)

![editor split view in same group](https://user-images.githubusercontent.com/4022813/145733058-be3f6f74-3e71-4339-acbb-a308703509ad.png)

**Installation:**

- Operating system: Windows/Linux
- VSCode version: 1.63.0
- VSCode extension version: 1.10.15
- Metals version: 0.10.9
- Scala version: 3.1.0

**Additional context**

Discussion in discord during advent of code 2021:
https://discord.com/channels/632642981228314653/632841682211045396/919515055354179624

More broadly, I also mention a few features that might be nice to have:
https://discord.com/channels/632642981228314653/632841682211045396/919363255577767966

>  - an output panel that continually watches worksheet as it updates
>  - ability to convert a worksheet quickly into a scala file with a main method
>  - ability to sbt runMain a worksheet
>  - evaluations appear in every open editor of a worksheet

stewSquared:
Hooray! Thanks, @tgodzik 
I'll let you know if I find follow-up issues (in particular, around the "evaluating" modal that doesn't pop up to let you cancel).

tgodzik:
> Hooray! Thanks, @tgodzik I'll let you know if I find follow-up issues (in particular, around the "evaluating" modal that doesn't pop up to let you cancel).

Does it also happen when you have a single worksheet?

stewSquared:
[...]

PR #788, opened 2021-12-13
Fix decorations not showing up if files are duplicated

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/784


kpodsiad (approved):
LGTM

PR #809, opened 2022-01-05
Make go to definition work with the newest languageclient

tgodzik:
`metals.goto-position` is needed for locally defined implicit decorations, but we can skip the step of sending it to the server.

I also updated the client configuration to declare the link format for command links.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -363,6 +363,7 @@ function launchMetals(
     statusBarProvider: "on",
     treeViewProvider: true,
     testExplorerProvider: true,
+    commandInHtmlFormat: "vscode",
   };
 
   const clientOptions: LanguageClientOptions = {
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -836,16 +837,6 @@ function launchMetals(
         }
       );
 
-      registerCommand("metals.goto-path-uri", (...args) => {
```
tgodzik, line 839: This is not used anymore not invoked by metals

src/extension.ts
```diff
@@ -860,6 +851,11 @@ function launchMetals(
         });
       });
 
+      registerCommand(`metals.${ServerCommands.GotoPosition}`, (args) => {
+        const location = args && (args as Location);
+        if (location) gotoLocation(location, false);
```
kpodsiad, line 856:
```suggestion
      registerCommand(`metals.${ServerCommands.GotoPosition}`, (location: Location) => {
        if (location) {
          gotoLocation(location, false);
        }
```

kpodsiad (approved):

src/extension.ts
tgodzik: Good idea! Fixed.

PR #813, opened 2022-01-06
[FindInJarFiles] Show tree view only after first call

### LINKED ISSUE #812: Find In files view always shows
**Describe the bug**

Even if suer is not using `find in files` it will show up on the left. 

Look on the tabs view

**Expected behavior**

Tab is shown only when in use.

**Screenshots**

image

**Installation:**

- Operating system: Linux
- VSCode version: 1.61
- VSCode extension version: 1.10.15
- Metals version: (found in VSCode settings, under `metals.serverVersion`) 010.9

**Search terms**

search in file tab

dos65:
Fixes #812

### DIFF
--- a/src/findinfiles.ts
+++ b/src/findinfiles.ts
@@ -1,4 +1,5 @@
 import {
+  commands,
   EventEmitter,
   ExtensionContext,
   OutputChannel,
@@ -53,6 +54,7 @@ export function createFindInFilesTreeView(
   provider: FindInFilesProvider,
   context: ExtensionContext
 ): TreeView<unknown> {
+  commands.executeCommand("setContext", "metals.showFindInFiles", false);
   const treeView = window.createTreeView("metalsFindInFiles", {
     treeDataProvider: provider,
     showCollapseAll: true,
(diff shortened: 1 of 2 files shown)


src/findinfiles.ts
```diff
@@ -53,6 +54,7 @@ export function createFindInFilesTreeView(
   provider: FindInFilesProvider,
   context: ExtensionContext
 ): TreeView<unknown> {
+  commands.executeCommand("setContext", "metals.showFindInFiles", false);
```
tgodzik, line 57: It doesn't seem to cause it to disappear when we move to a separate window.

We should also do:
```
const onDidChangeVisibility = treeView.onDidChangeVisibility(async (e) => {
    if (!e.visible)
      commands.executeCommand("setContext", "metals.showFindInFiles", false);
  });
 context.subscriptions.push(onDidChangeVisibility);
```

tgodzik (changes requested):

src/findinfiles.ts
tgodzik: OR actually, maybe we want it to show still :thinking:

tgodzik (approved):
Actually, ignore me, this is exactly how references work.

dos65:
> Actually, ignore me, this is exactly how references work.

Yep, I remember that I had some version of what you proposed but then compared with how default search and references and decided not to do that.

PR #826, opened 2022-01-17
ci: publish pre-release version to VS Code Marketplace

kpodsiad:
From docs:
_VS Code Marketplace only support major.minor.patch for extension versions and semver pre-release tags are not supported. Support for this will arrive in the future. Because of this we recommend that extensions use major.EVEN_NUMBER.patch for release versions and major.ODD_NUMBER.patch for pre-release versions. For example: 0.2.* for release and 0.3.* for pre-release. VS Code will auto update extensions to the highest version available, so even if a user opted into a pre-release version and there is an extension release with a higher version, that user will be updated to the released version._

Since `semver pre-release tags are not supported` some workaround has to be created, docs advise using `ODD_NUMBER` and `EVEN_NUMBER` for minors to differentiate between release and pre-release versions. It seems a little bit counterintuitive, hence together with @tgodzik we concluded that this approach will be easier. 

Release versions will use `major.MINOR.0` and pre-release ones will use `major.minor.PATCH`.

The `pre-release` job was added which:
- `needs: [release]` because it should run sequentially (in case there is the release, we want to publish pre-release anyway with proper version - bumped both minor and patch)
- has `always()` - `Causes the step to always execute` - even when release was skipped
- doesn't publish pre-release for trivial actions such as bumping dependencies

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,4 +1,4 @@
-name: Release
+name: Pre-release
 
 on:
   push:
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -16,7 +16,7 @@ jobs:
   pre-release:
     if: |
       !contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
-      !contains(github.event.head_commit.message, 'Signed-off-by: dependabot[bot]') &&
+      !contains(github.event.author.email, 'dependabot[bot]@users.noreply.github.com') &&
       github.event.author.email != 'bot@scalameta.org'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -6,13 +6,6 @@ on:
       - main
 
 jobs:
-  logging:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Print event metadata
-        run: |
-          echo "${{toJSON(github.event.commits[0])}}"
-          echo "${{toJSON(github.event.head_commit)}}"
   pre-release:
     if: |
       !contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,44 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  logging:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+        echo "${{toJSON(github.event.commits[0])}}"
+        echo "${{toJSON(github.event.head_commit)}}"
+  pre-release:
+    if: |
+      !contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
+      !contains(github.event.head_commit.message, 'Signed-off-by: dependabot[bot]') &&
+      github.event.author.email != 'bot@scalameta.org'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "16"
+      - uses: actions/checkout@v2
+      - name: Setup git env
+        run: |
+          git config --global user.email "bot@scalameta.org"
+          git config --global user.name "Scalameta bot"
+      - run: |
+          yarn install
+          yarn version --no-git-tag-version --patch
+      - name: Commit changes
+        run: |
+          git add .
+          git commit -m "Increase patch version for pre-release"
+      - name: Publish pre-release version to Visual Studio Marketplace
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -59,3 +61,36 @@ jobs:
           OVSX_PAT: ${{ secrets.OVSX_TOKEN }}
         run: |
           yarn ovsx:publish --pat $OVSX_PAT ./metals-${{ steps.get_version.outputs.NEW_VERSION }}.vsix
+
+  pre-release:
+    if: |
+      always() &&
+      github.event_name == 'push' && 
+      !contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
+      !startsWith(github.event.head_commit.message, 'Bump')
```
tgodzik, line 70: Could we isntead check if the author is scalameta bot or dependabot?

.github/workflows/release.yml
```diff
@@ -3,15 +3,18 @@ name: Release
 on:
   release:
     types: [published]
+  push:
```
tgodzik, line 6: Could we put it into separate workflow?

tgodzik (changes requested):
Looks good! Just some minor comments.

.github/workflows/release.yml
kpodsiad: It depends on what we want to do after release. If we want to publish only the release version then it can be put into a separate workflow. 
If we want to publish release and then pre-release then I'm afraid that race condition may occur - release bumps minor version and commits it, pre-release bumps patch version - there is need for synchronization between them. Doing it via `need` was the easiest way I found to do it

.github/workflows/release.yml
kpodsiad: Seems like `github.event_name == 'push' &&`, depending on what we want to do after release, is or isn't needed. See your other comment

kpodsiad: It is possible, but I'm not sure what that check should look like. Can I add a logging step, which will log all necessary information needed for this check? After the first PR's from that bots workflow can be tweaked

gabro (approved):
Nice idea!

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Print event metadata
+        run: |
+          echo "${{toJSON(github.event.commits[0])}}"
+          echo "${{toJSON(github.event.head_commit)}}"
+  pre-release:
+    if: |
+      !contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
+      !contains(github.event.head_commit.message, 'Signed-off-by: dependabot[bot]') &&
+      github.event.author.email != 'bot@scalameta.org'
```
kpodsiad, line 20:
```shell
!contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
 !contains(github.event.head_commit.message, 'Signed-off-by: dependabot[bot]') &&
 ```
is not nice but it should be good enough for start. I'm not sure that dependabot has some constant email that ca be used. 
`logging` workflow should show what properties can be used to filter out dependabot PR's more gracefully. 

Alternatively, I can leave only `!contains(github.event.head_commit.message, 'scalameta/dependabot/')` which should be sufficient and don't bother with possible better alternatives.

@tgodzik it's your call

File: .github/workflows/pre-release.yml [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+          git config --global user.name "Scalameta bot"
+      - run: |
+          yarn install
+          yarn version --no-git-tag-version --patch
+      - name: Commit changes
+        run: |
+          git add .
+          git commit -m "Increase patch version for pre-release"
+      - name: Publish pre-release version to Visual Studio Marketplace
+        uses: HaaLeo/publish-vscode-extension@v1
```
tgodzik, line 39: Do we need the additional action? Can't we use vsce? This isn't an official action so I would be a bit hesitant to use it.

tgodzik (changes requested):

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,45 @@
...
+    branches:
+      - main
+
+jobs:
+  logging:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Print event metadata
+        run: |
+          echo "${{toJSON(github.event.commits[0])}}"
```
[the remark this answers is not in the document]
kpodsiad: Actually, it was a question for you https://github.com/scalameta/metals-vscode/pull/826#discussion_r786237833. I modified the condition and you don't have any objections so I can remove it.

.github/workflows/pre-release.yml
kpodsiad: It's actively maintained and it's very easy to use hence I don't see any problem with using it

tgodzik (approved):
LGTM! Thanks for driving this!

PR #830, opened 2022-01-18
ci, tests: add extension tests

kpodsiad:
* add extension tests which allow to use vscode API https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions.
I want to use `vscode` API to test Test Explorer behavior - creating test items via `TestController` and etc.

* split `test` dir into `unit` and `extension` - although I think `extension` could be placed elsewhere

* ~change `extensionDependencies` to `extensionPack` in `package.json`. Metals do not depend on Scala syntax, but these extensions should be bundled - see https://code.visualstudio.com/api/references/extension-manifest#extension-packs~
Actually, it was done in https://github.com/scalameta/metals-vscode/pull/835. There's also a explanation why.


package.json
```diff
@@ -824,7 +824,8 @@
     "vscode:prepublish": "yarn compile",
     "compile": "tsc -p ./",
     "watch": "tsc -watch -p ./",
-    "test": "jest",
+    "test": "jest unit/",
```
kpodsiad, line 827: `"test": "jest unit/",` is not ideal, I think [`jest` config](https://jestjs.io/docs/configuration) may help with that

package.json
```diff
@@ -854,7 +860,7 @@
     "promisify-child-process": "4.1.1",
     "vscode-languageclient": "7.0.0"
   },
-  "extensionDependencies": [
```
tgodzik, line 857: Maybe this should be a separate PR?

tgodzik (changes requested):
Looks good, though I would move the extension pack change to a different PR otherwise it's kind of hidden insid ehere.

gabro (approved):
nice! It looks good, just one non-blocking question

package.json
```diff
@@ -854,7 +860,7 @@
...
+  "extensionPack": [
```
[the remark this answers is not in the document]
kpodsiad: I'm really glad you asked because I could have forgotten about one crucial thing. 

`extensionDependencies` means that it will be installed if the main extension is installed (since it depends on its API), I think `extensionPack` doesn't have this kind of guarantee. In this particular case, I changed it because the extension's tests were failing that there is no `scala-lang.scala` extension.

@tgodzik has a right, I'll do it in separate PR and make sure after that if installing Metals installs Scala Syntax too

package.json
kpodsiad: Yeah, good point

package.json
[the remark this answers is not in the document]
kpodsiad: Actually I was wrong
Screenshot 2022-01-20 at 13 06 34

kpodsiad:
I changed jest configuration, running `yarn test` will only run unit tests

PR #832, opened 2022-01-19
fix: update vscode version to support prelease

kpodsiad:
I missed that prelease requires vscode `^1.63` but we don't want to force other users to update. Maybe what I propose is dirty, but it keeps status quo for release version and enables us to publish pre-release ones

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,47 @@
+name: Pre-release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  pre-release:
+    if: |
+      !contains(github.event.head_commit.message, 'scalameta/dependabot/') &&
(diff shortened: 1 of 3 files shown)


tgodzik (changes requested):
I think this is too soon to update.

tgodzik (approved):
Not the most fun thing to do, but it's better than updating it for all users. Thanks!

.github/workflows/pre-release.yml
```diff
@@ -36,3 +42,6 @@ jobs:
           preRelease: true
           yarn: true
       # Don't publish for VSX, it doesn't support pre-release versions
+      # Push changes - increased patch version
+      - name: Push changes
+        run: git push origin HEAD:main
```
tgodzik, line 47:
```suggestion
        run: git push origin HEAD
```

kpodsiad: As we speak, I'll leave this as it is.

PR #834, opened 2022-01-20
ci: allow to skip prerelease workflow

kpodsiad:
Sometimes changes don't bring any value to users and prerelase version doesn't need to be published. To avoid unnecessary versions and commits (scalameta bot bumping patch version) one can add `[skip pre]` to the commit message - prerelease workflow will be skipped. 

I think it's nice to have such a mechanism.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -8,9 +8,10 @@ on:
 jobs:
   pre-release:
     if: |
+      github.event.author.email != 'bot@scalameta.org' &&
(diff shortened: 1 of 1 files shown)


gabro (approved):
makes sense!

kpodsiad:
I almost forget to include `[skip pre]` in the commit message 😅 Fortunately, I added description with this phrase, but it wasn't on purpose :D

tgodzik (approved):

PR #835, opened 2022-01-20
chore: switch from `extensionDependencies` to `extensionPack` [skip pre]

kpodsiad:
It is a very subtle change. However, it's plays a huge role in extension tests where, for some reason, `scala-lang.scala` extension can't be found if it's specified in `extensionDependencies`. Because Metals doesn't depend on Scala Syntax API it's safe to make this change.

Screenshot 2022-01-20 at 13 06 34

### DIFF
--- a/package.json
+++ b/package.json
@@ -854,7 +854,7 @@
     "promisify-child-process": "4.1.1",
     "vscode-languageclient": "7.0.0"
   },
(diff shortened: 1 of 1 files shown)


tgodzik (approved):

gabro (approved):
👍 makes sense

PR #836, opened 2022-01-20
refactor: Test Explorer - depends on server notifications

kpodsiad:
Needs https://github.com/scalameta/metals/pull/3554

* do not query the Metals server for test suites for every refresh model event.
* handle test events from metals `addSuite`, `removeSuite`, `addTestCases`. The logic for each event is located in a separate file. Traversing the TestExplorer (it's basically a tree) could be extracted to a common function but I decided that it will be too much hassle with passing callback for every branch. There is some duplication but the code should be straightforward to understand.
* I added tests which checks if events are handled correctly and results are interpreted correctly

@tgodzik don't be scare seeing `+804` - half of it are tests ;)


src/extension.ts
```diff
@@ -699,6 +694,11 @@ function launchMetals(
           case ClientCommands.FocusDiagnostics:
             commands.executeCommand(ClientCommands.FocusDiagnostics);
             break;
+          case "metals-update-test-explorer": {
```
tgodzik, line 697: We should add it to `ClientCommands` in `metals-languageclient`

src/test-explorer/add-test-suite.ts
```diff
@@ -0,0 +1,104 @@
...
+  targetUri: TargetUri,
+  event: AddTestSuite
+): void {
+  const buildTargetItem = getOrCreateBuildTargetItem(
+    testController,
+    targetName,
+    targetUri
+  );
+
+  function addTestSuiteLoop(parent: vscode.TestItem, testIds: string[]) {
```
tgodzik, line 27:
```suggestion
  function addTestSuiteLoop(parent: vscode.TestItem, testPath: string[]) {
```
Isn't this a test path actually?

src/test-explorer/add-test-cases.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Assumes that path from root to the leaf exists.
+ */
+export function addTestCases(
+  testController: vscode.TestController,
+  targetName: TargetName,
+  targetUri: TargetUri,
+  event: AddTestCases
+): void {
+  //
+  function addTestCasesLoop(parent: vscode.TestItem, testIds: string[]): void {
```
tgodzik, line 21:
```suggestion
  function addTestCasesLoop(parent: vscode.TestItem, testPath: string[]): void {
```

src/test-explorer/add-test-suite.ts
```diff
@@ -0,0 +1,104 @@
...
+          targetName,
+          targetUri,
+        };
+        testCache.setMetadata(packageNode, data);
+        addTestSuiteLoop(packageNode, restOfIds);
+      }
+    } else {
+      const { className, location, fullyQualifiedClassName } = event;
+
+      if (parent.children.get(fullyQualifiedClassName)) {
```
tgodzik, line 48: Shouldn't we replace it if the suite already exists?

src/test-explorer/add-test-suite.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      const parsedUri = vscode.Uri.parse(location.uri);
+      const parsedRange = toVscodeRange(location.range);
+      const testItem = testController.createTestItem(
+        fullyQualifiedClassName,
+        className,
+        parsedUri
+      );
+      testItem.canResolveChildren = event.canResolveChildren;
```
tgodzik, line 59: What does `canResolveChildren` mean? Could you comment here?

src/test-explorer/test-manager.ts
```diff
@@ -68,13 +56,37 @@ class TestManager {
...
+      }
+    };
   }
 
   async enable(): Promise<void> {
     this.isDisabled = false;
     await this.discoverTestSuites();
   }
 
+  debug() {
```
tgodzik, line 73: Should this be removed?

src/test-explorer/test-run-handler.ts
```diff
@@ -55,35 +57,50 @@ export async function runHandler(
...
-    } else {
-      run.started(test);
-      const children = testCache.getTestItemChildren(test);
-      children.forEach((c) => run.enqueued(c));
-
-      try {
-        await commands.executeCommand("workbench.action.files.save");
-        const testsIds = children.map((t) => t.id);
-        const session = await createDebugSession(data.targetUri, testsIds);
+      continue;
```
tgodzik, line 60: Could we do `if else` instead of continue ?

tgodzik (changes requested):
Looks good as far as my typescript knowledge goes :sweat_smile: 

Left a few questions.

src/test-explorer/add-test-suite.ts
kpodsiad: I think we should

tgodzik (approved):
Looks good! Let's focus on the Metals PR now

gabro (approved):
I'm fairly busy at work, so I won't reasonably be able to review this anytime soon.

I trust @kpodsiad's work and @tgodzik's review :-)

PR #850, opened 2022-01-26
Fix invoking focus on problems tab

tgodzik:
Previously, each time an error was detected we would try to register a command indicated by `command` parameter if it was already registred. This would obviously fail as we explicitely checked for that. Now, whenever there is a problem in the workspace, the problem tab will be focused. Though the question is "should it?", wouldn't it be a bit annoying?

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -729,9 +729,7 @@ function launchMetals(
           item.command = params.command;
           commands.getCommands().then((values) => {
             if (values.includes(command)) {
-              registerCommand(command, () => {
-                client.sendRequest(ExecuteCommandRequest.type, { command });
-              });
+              commands.executeCommand(command);
             }
           });
         } else {


kpodsiad (approved):
> Though the question is "should it?", wouldn't it be a bit annoying?

We can add an option to customize this behavior, purely on vscode side.

tgodzik:
> > Though the question is "should it?", wouldn't it be a bit annoying?
> 
> We can add an option to customize this behavior, purely on vscode side.

Ok, we can test on ourselves first and add an option later if needed

DavidWeatherford:
I'm using Metals Server Version 0.11.1 and I'm seeing the same behavior as @peteslater-ee. I had to downgrade my metals extension in VS Code back to v1.12.0 just to be able to work. When using v1.13.0 my cursor keeps focusing on the Problems panel when a compilation error occurs. This is especially annoying when auto-save is enabled in VS Code and then every few key strokes results in my cursor being moved to the Problems panel.

tgodzik:
Hi @DavidWeatherford, `0.11.1` is not the current Metals server version. We recently released `0.11.2`, which works best with the newest Visual Studio Code extension. `0.11.1` will work correctly with the previous extension version `1.12.0`

What is the reason you use `0.11.1` version? Did you at any point modify the version?

DavidWeatherford:
@tgodzik My team uses sbt version `1.2.8` and scala `2.12.11`. Code navigation (specifically finding references) isn't supported when using sbt version `1.2.8` and the latest metals/bloop versions. I downgraded bloop to `1.4.12` and metals to `0.11.1` to get back the "find references" functionality. However, that's when I encountered the new behavior in the VS Code metals extension `v1.13.0` where my cursor focus is moved to the Problems panel after each compilation.

tgodzik:
What is the problem with sbt `1.2.8` with Bloop? Does the build import not work? You can downgrade Bloop itself without downgrading anything else, but there should not be any issues between `1.4.12` and `1.4.13`

DavidWeatherford:
@tgodzik I've done some more testing based on your comment:

1. With sbt `1.2.8`, metals `0.11.2`, bloop `1.4.13`, and metals VS Code extension `1.13.0` I lose the ability to navigate code using "find all references".
2. When I downgrade metals to `0.11.1` and leave all others at their latest version, I get back the ability to navigate code using "find all references" however compilation results in my cursor focus being moved to the Problems panel.

Bloop version does not play a role in this. Is losing the ability to navigate code with "find all references" from metals `0.11.1` to `0.11.2` a metals VS Code extension issue or a metals issue? If it is a metals issue, then I'll have to remain at `0.11.1` for now and then it would be helpful to put this pull request's feature behind a setting in VS code.

tgodzik:
> ss losing the ability to navigate code with "find all references" from metals 0.11.1 to 0.11.2 a metals VS Code extension issue or a metals issue?

This looks like a Metals issue then, is there anything in the logs that pops up? If you remove `.metals` directory and reload the window does it help?

> then it would be helpful to put this pull request's feature behind a setting in VS code.

This wasn't really a feature, but a bugfix of something that was never properly working, once it turned out to be working it became annoying.

Probably it's best to just set the VS Code version to 1.12.0 (can be done in the extension tab) and use the default Metals version for that version which is `0.11.1` (best to remove the version from the settings json). VS Code extension versions are only guaranteed to work with the default version, otherwise we cannot promise anything.

In the meantime if you could report the issue in the Metals repo with all the details we can take a look at that.

DavidWeatherford:
> This looks like a Metals issue then, is there anything in the logs that pops up? If you remove .metals directory and reload the window does it help?
>
Unfortunately no logs and removing the directory doesn't help. Thank you for the advice!

DavidWeatherford:
@kpodsiad This behavior appears to be limited to a specific project. I've checked others using the same configuration and they do not run into this same issue. I'll continue investigating on my side.

I just checked the `metals.log` file and saw the below error. I've removed some details in the file paths as they are private.
`2022.03.25 12:52:50 ERROR code navigation does not work for the file '<file I was doing find all references in>' because the SemanticDB file '<project root>\.bloop\root\bloop-bsp-clients-classes\classes-Metals-JjdLy-jLRAKNr-gmuqfPfA==\META-INF\semanticdb\<path to file I was doing find all references in>.semanticdb' doesn't exist. There can be many reasons for this error.`

kpodsiad:
I had seen a lot of  `because the SemanticDB doesn't exists` warnings and after https://github.com/scalameta/metals/pull/3405 I see them very rarely. 

@DavidWeatherford Could I ask you to do more testing? Usually, provided that the project was compiling, these steps helped me with the mentioned problem.
- update vscode extension to the `v1.13.0`
- remove manually typed `server version` (if there's any)
- check if there are no errors in the problems tab
- try to run find all references in your file
- if semanticDB is not found execute `connect to the build server` command. It should create new folder in `bloop-bsp-clients-classes`
- if any of these helped try to remove all of the `bloop-bsp-clients-classes` directories


May I ask if there is some blocker that prevents you and your team from migrating sbt version to the >1.4.0? Starting from this version there's a BSP support and you can use sbt as a build server. From my experience, I've never observed SemanticDB in in sbt.

PR #851, opened 2022-01-26
fix: debug isn't working after compilation error

kpodsiad:
fixes https://github.com/scalameta/metals-vscode/issues/849

I was thinking about signaling the user what's going on. However, vscode notification mechanism, which is a surprise, doesn't provide an option to display a true "notification" - message which will disappear on its own after some time.

I was thinking about `vscode.window.showWarningMessage("Debug didn't start because of compiler errors");` but this message stays until explicitly dismissed.

### DIFF
--- a/src/test-explorer/test-run-handler.ts
+++ b/src/test-explorer/test-run-handler.ts
@@ -45,7 +45,7 @@ vscode.debug.registerDebugAdapterTrackerFactory("scala", {
 export async function runHandler(
   testController: TestController,
   noDebug: boolean,
-  callback: () => void,
+  afterFinished: () => void,
   request: TestRunRequest,
   token: CancellationToken
 ): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/test-explorer/test-run-handler.ts
```diff
@@ -77,6 +77,7 @@ export async function runHandler(
 
         await analyzeResults(run, children, callback);
       } catch (error) {
+        callback();
```
tgodzik, line 80: No callback seems to be sent after  `if (!wasStarted) {` also?

tgodzik (changes requested):

src/test-explorer/test-run-handler.ts
kpodsiad: God catch. I added `finally` callback

tgodzik (approved):

PR #856, opened 2022-01-27
Manually add loadash to resolve issue with node-fetch

tgodzik:


tgodzik:
Should also be fixed by https://github.com/scalameta/metals-vscode/pull/857

PR #857, opened 2022-01-27
refactor: remove jest and use mocha

kpodsiad:
Since mocha is already used in extension tests there is no point in keeping jest.

`yarn-lock` is now 2x smaller than before, I hope it's a good sign 🧐


tgodzik (approved):
LGTM!

kpodsiad:
It also should help with our previous dependencies problems because jest was bringing that unlucky dep which was conflicting.

PR #861, opened 2022-01-28
Make Java Home config scope machine-overridable

dlobsinger:
The previous change to the scope of `metals.javaHome` from #681 makes it so that users cannot override the configuration at a workspace level. For users that have multiple projects with specific Java versions it is useful to be able to set the path per project.
The VSCode configuration scope can be set as `machine-overridable` to allow for the setting to be explicitly configured for a workspace but still defaults to being configured on a machine level: https://code.visualstudio.com/api/references/contribution-points#Configuration-schema

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
         },
         "metals.javaHome": {
           "type": "string",
-          "scope": "machine",
+          "scope": "machine-overridable",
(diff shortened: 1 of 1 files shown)


kpodsiad (approved):
Thanks for fixing it! After it's merged (add a small delay for vcode marketplace) you can use `pre-release` version of extension to have that changes ;)

PR #862, opened 2022-01-30
Support synthetic decorations in worksheets

tgodzik:
Previously, we would have a separate type of decorations for worksheets. Now, we have a separate type for worksheets evaluations and synthetic decorations.

Needed for https://github.com/scalameta/metals/pull/3582

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -69,10 +69,7 @@ import {
 import * as metalsLanguageClient from "metals-languageclient";
 import { startTreeView } from "./treeview";
 import * as scalaDebugger from "./scalaDebugger";
-import {
-  DecorationTypeDidChange,
-  DecorationsRangesDidChange,
-} from "./decoration-protocol";
+import { DecorationsRangesDidChange } from "./decoration-protocol";
 import { clearTimeout } from "timers";
 import { increaseIndentPattern } from "./indentPattern";
 import { gotoLocation, WindowLocation } from "./goToLocation";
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -1068,9 +1066,6 @@ function launchMetals(
       scalaDebugger.initialize(outputChannel).forEach((disposable) => {
         context.subscriptions.push(disposable);
       });
-      client.onNotification(DecorationTypeDidChange.type, (options) => {
```
tgodzik, line 1071: This is never used and not sure if it actually should.

src/extension.ts
```diff
@@ -1068,9 +1066,6 @@ function launchMetals(
...
-        decorationType = window.createTextEditorDecorationType(options);
-      });
       client.onNotification(DecorationsRangesDidChange.type, (params) => {
         const editors = window.visibleTextEditors;
         const path = Uri.parse(params.uri).toString();
```
kpodsiad, line 1071: I've noticed that `client.onNotification` return `Disposable` but we're discarding returned value. Maybe it should be added to `context.subscriptions` (yeah, it's a broader problem)

`extension.ts` is a very large file, I wonder if we can e.g. create `notifications` dir and extract all `client.onNotification` to different files

kpodsiad (approved):
I'd love to see some refactor which will reduce `extension.ts` size, but it isn't obligatory ;)

src/extension.ts
tgodzik: I added all the disposables from `onNotification` to subscriptions. Good idea!

kpodsiad: Thanks, that's a step into good direction ;)

kpodsiad (approved):

PR #863, opened 2022-01-31
Add play framework configuration snippet

tgodzik:


package.json
```diff
@@ -809,6 +809,19 @@
...
+            "label": "Scala: Play framework prod server",
+            "description": "A new configuration for running a production Play framework production server",
+            "body": {
+              "type": "scala",
+              "request": "launch",
+              "name": "Play main",
+              "mainClass": "play.core.server.ProdServerStart",
+              "args": [],
+              "jvmOptions": [],
+              "buildTarget": "???"
```
kpodsiad, line 823: Is `???` okay?

kpodsiad (commented):
Just one question

package.json
tgodzik: That's the default value that we use in the snippets. It would be better to allow the users choose when creating the snippet, but it's a bit more complicated.

kpodsiad: If only knowledge about build targets wasn't necessary.

tgodzik: In this case it's necessary, since we don't know which target is a play server target :/ 

We could try to improve it for sure, but we don't keep the main classes from dependencies currently.

kpodsiad: Could we replace `???` with detailed information on how to fill this field? Like "Type build target name which has a dependency on `play.core`" or sth like this?

package.json
```diff
@@ -726,7 +726,7 @@
               },
               "buildTarget": {
                 "type": "string",
-                "description": "Name of the build target of the class to run"
+                "description": "Name of the build target of the class to run, can be found in the projects section of the Metals tab"
```
tgodzik, line 729: I added a bit more info here, which gets displayed when you mouse over the build target. I feel like including the description in the value might be misleading.

kpodsiad: Good idea!

kpodsiad (approved):

PR #864, opened 2022-01-31
Allow for relative paths in java home config

dlobsinger:
This change is an implementation to allow for the path provided in `metals.javaHome` to be a relative path from the root directory of the workspace. 

The current behaviour when calling `getJavaHome` is that the path must be an absolute path or is resolved to a path relative to the system root `/`. If a user has a JDK located in their project directory, for example by using a nix profile to install a JDK in the same directory, it would be convenient to allow for the Metals configuration to use a path relative to the workspace.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -125,7 +125,6 @@ export async function activate(context: ExtensionContext): Promise<void> {
     },
     async () => {
       commands.executeCommand("setContext", "metals:enabled", true);
-
       try {
         const javaHome = await getJavaHome(getJavaHomeFromConfig());
         return fetchAndLaunchMetals(context, javaHome);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -121,10 +121,20 @@ export async function activate(context: ExtensionContext): Promise<void> {
     },
     async () => {
       commands.executeCommand("setContext", "metals:enabled", true);
+
+      const javaHomeConfig = ((jhPath) =>
+        jhPath?.trim() && !path.isAbsolute(jhPath)
+          ? path.resolve(
+              ...[workspace.workspaceFolders?.[0]?.uri.fsPath, jhPath].filter(
+                (s): s is string => !!s
+              )
+            )
+          : jhPath)(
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -121,10 +121,20 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
+      const javaHomeConfig = ((jhPath) =>
+        jhPath?.trim() && !path.isAbsolute(jhPath)
+          ? path.resolve(
+              ...[workspace.workspaceFolders?.[0]?.uri.fsPath, jhPath].filter(
+                (s): s is string => !!s
+              )
+            )
+          : jhPath)(
+        workspace.getConfiguration("metals").get<string>("javaHome")
+      );
```
kpodsiad, line 134: I was thinking about little bit more verbose
```typescript
export function getJavaHomeFromConfig(): string | undefined {
  const javaHomePath = workspace
    .getConfiguration("metals")
    .get<string>("javaHome");
  if (javaHomePath?.trim() && !path.isAbsolute(javaHomePath)) {
    const pathSegments = [
      workspace.workspaceFolders?.[0]?.uri.fsPath,
      javaHomePath,
    ].filter((s): s is string => s != null);
    return path.resolve(...pathSegments);
  } else {
    return javaHomePath;
  }
}
```
It's not as concise as what you're proposing but I find it easier to read ;)

Also, `extension.ts` is an enormous file, moving this function to e.g. `util.ts` might be a good idea.

kpodsiad (changes requested):
Thanks for your another contribution ;) I think we could simplify it a little bit, that immediate execution part wasn't obvious for me :c

src/extension.ts
dlobsinger: Good point about moving this into `util.ts` - I wasn't sure whether it would be better to have as little impact in the scope of the code, or to have a more verbose implementation. I'll make those changes.

kpodsiad (approved):
LGTM!

PR #874, opened 2022-02-02
refactor: extract commands from `extension.ts` to separate files

kpodsiad:
`extension.ts` is an enormous file and I always struggle to find anything in it. As a first step, I decided to extract some commands registration to separate files. This is a quick poc, let me know what do you think.

cc: @tgodzik @gabro

### DIFF
--- a/src/commands/client-commands.ts
+++ b/src/commands/client-commands.ts
@@ -1,0 +1,61 @@
+import { ClientCommands } from "metals-languageclient";
+import { commands, ExtensionContext, window } from "vscode";
+import { LanguageClient } from "vscode-languageclient/node";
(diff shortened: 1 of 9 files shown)


tgodzik (commented):
Looks like a good approach! Minus side, we no longer will be able to brag on how big the extension file is :sweat_smile:

kpodsiad:
> Looks like a good approach! Minus side, we no longer will be able to brag on how big the extension file is :sweat_smile: 

I wasn't aware that it's a reason to brag 😅

kpodsiad:
`extension.ts` size was reduced from 1330 to 916, I think It's good enough for this PR

PR #886, opened 2022-02-10
fix: do not use prefixes as tests labels

kpodsiad:


kpodsiad:
Really simple fix and since I already have an opened PR that makes some changes in this code I'll merge this PR without review.

PR #888, opened 2022-02-11
Fix/target test item uri

kpodsiad:
I also added a super basic eslint for mocha to not leave `.only` tests in CI ;)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,8 @@
     "out": true // set this to false to include "out" folder in search results
   },
   "editor.formatOnSave": true,
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "files.watcherExclude": {
+    "**/target": true
+  }
 }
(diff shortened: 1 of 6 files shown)


.vscode/settings.json
```diff
@@ -7,5 +7,8 @@
     "out": true // set this to false to include "out" folder in search results
   },
   "editor.formatOnSave": true,
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "files.watcherExclude": {
```
tgodzik, line 11: Wasn't this added by metals by msitake?

tgodzik (approved):
LGTM, just one question

.vscode/settings.json
kpodsiad: i forgot to mention it. Running extension tests has to be done from vs code insiders and there musn't be any normal vscode instance opened. So when your workflow looks like:
- implement feature in vscode
- write some extension tests in vscode
- kill all vscode instances, open vscode insiders
- run extension tests from insiders
- e.g. code with `metals-vscode` opens, Metals are modifying settings json

I was tired of removing it each time before commit, so I commited it once and for all.

Issue #880, opened 2022-02-07
Goto definition is very slow


anton-zen:
**Describe the bug**

Goto definition takes from 7 to 15 seconds.

**To Reproduce** Steps to reproduce the behavior:

Find a val or def reference and go to its definition via context menu.

**Expected behavior**

200ms

**Installation:**

- Operating system: macOS
- VSCode version: 1.64.0
- VSCode extension version: v1.12.0
- Metals version: 0.11.1

**Additional context**

Macbook pro 2019, 2.6 GHz 6-Core Intel Core i7, 32 GB of RAM.

kpodsiad:
Is this a Scala 3 project?

tgodzik:
Thanks for reporting! Unfortunately we can't do anything without a reproduction, since normally go to definition works fine. I would be great if you could provide more information.

anton-zen:
@kpodsiad yes, it's a Scala 3 project.

@tgodzik I'll be happy to assist as much as I can. Is there some way to diagnose this? There is nothing in Output tab when I use this feature.

tgodzik:
We would need information whether this happens on every source code in your project or maybe a particular symbol. Is it in the same file? Does the code compile? The best would be a small snippet that we can test it out on or a link to the project if it's possible.

anton-zen:
Yes, code compiles, sbt shell runs fine, and metals import finishes successfully.

It seems to happen with any symbol but I am using a reference to a sibling method and it's always slow, not just at first attempt. Opening an additional vscode window seems to be triggering this issue but I am not 100% sure. Maybe it's just happening after a couple of minutes.

I could reproduce this with this Scala 3 project https://github.com/scala/scala3-example-project with `runExample` symbol.
Also noticed that java process is taking between 100-300% of CPU when this happens.

Also happens in a tiny hello world Scala 2 project but takes just 2-3 seconds (`sbt new scala/hello-world.g8`) with this code:
```
object Main extends App {
  println(hello)

  def hello = {
    "hello!"
  }
}
```

anton-zen:
There is something interesting in metals.log though:
```
[...]

PR #890, opened 2022-02-14
Add a question about disabling unrelated extensions

tgodzik:
There is a similar question in VS Code issue template so that might be helpful to diagnose the problem.

Inspired by https://github.com/scalameta/metals-vscode/issues/880

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -10,6 +10,10 @@ assignees: ""
 
 <!-- A clear and concise description of what the bug is. -->
 
+**Does this happen with only Scala Syntax and Metals enabled?**
+
+Yes/No
+
 **To Reproduce** Steps to reproduce the behavior:
 
 1. Go to '...'


kpodsiad:
Alternatively, we can warn about known incompatible extensions (dotty, that recent sbt one). Isn't it better to prevent?

tgodzik:
> Alternatively, we can warn about known incompatible extensions (dotty, that recent sbt one). Isn't it better to prevent? (I think there is already one for dotty)

I don't think it's an alternative, it would be good to do both. There might be other extensions that we don't know of.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -10,6 +10,10 @@ assignees: ""
 
 <!-- A clear and concise description of what the bug is. -->
 
+**Does this happen with only Scala Syntax and Metals enabled?**
```
kpodsiad, line 13: I would rephrase that to sth like: `Do you have other Scala-related extensions like dotty, lightbend sbt, etc. ?` I'm afraid that some users might have an impression that non-scala extensions may affect metals (sometimes they do like errors from java server, but it's a different case I think)

tgodzik: But non-scala extension actually may also cause issues, there is a bunch of interactions that happened in the past and this might help out to rule them out.

kpodsiad: Do you expect that users will list all/some of their extensions when filling the form? I doubt it. On the contrary, it may be easier for them to list scala-related extensions. 

I'm afraid that such a general question won't have ay benefits and it's better to compile more narrowed one (like about scala-related extensions or java LSP) which will target 70% of cases than a very generic one which will require too much hassle from the users ("should I list all of them?", "I don't know if someone can interfere with Metals"). Don't you think?

>  there is a bunch of interactions that happened in the past

For example? I'm familiar with issues about java LSP interop, but I don't recall others.

tgodzik:
Since it's more controvertial than initially thoguht, I will close it.

.github/ISSUE_TEMPLATE/bug_report.md
kpodsiad: > If we don't disable all other extensions then we cannot be sure that there is an interaction.

It's true but in reality, it can be very impractical. 

Sorry for being very unconvinced about this change, but I'm really worried that it can be very misleading to normal users (not these few pro-users). 

Maybe we can refactor the issue template to be divided into a few parts like the one in metals is? Then we can make a section about other extensions non-mandatory or prefill it with some defaults or leave a suggestion through placeholders? What do you think?
It's not my intention to do nothing, but I want to do something which will have a minimal impact on creating an issue. Assuming that 60-70% of issues may come with java or scala related extensions it can be a good middle ground?

PR #892, opened 2022-02-17
fix: use `||` instead of `??`

kpodsiad:
For some reason an empty string was being taken as a Metals version.

`"" || 1` yields 1 where `"" ?? 1` yields "" and it's a reason why. `??` was my suggestion at review, guilty as charged.


dos65 (approved):
Thx for the fix!

PR #893, opened 2022-02-18
Change displayBuildTarget to listBuildTargets

Arthurm1:
see https://github.com/scalameta/metals/pull/3649

This shifts the burden of quickpicking from Metals onto the client

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -567,6 +567,40 @@ function launchMetals(
         );
       });
 
+      function displayBuildTarget(target: string): void {
+        const workspaceRoots = workspace.workspaceFolders;
+        if (workspaceRoots && workspaceRoots.length > 0) {
+          const uriStr = `metalsDecode:file:///${workspaceRoots[0].name}/${target}.metals-buildtarget`;
+          const uri = Uri.parse(uriStr);
+          workspace
+            .openTextDocument(uri)
+            .then((textDocument) => window.showTextDocument(textDocument));
+        }
+      }
+
+      registerCommand(`metals.target-info-display`, async (...args: any[]) => {
+        if (args.length > 0) {
+          // get build target name from treeview uri of the form "projects:file:/root/metals/.mtags/?id=mtags3!/_root_/"
+          const treeViewUri = args[0] as string;
+          const query = treeViewUri.split("/?id=");
+          if (query.length > 1) {
+            const buildTarget = query[1].split("!");
+            if (buildTarget.length > 0) {
+              displayBuildTarget(buildTarget[0]);
+            }
+          }
+        } else {
+          // pick from list of targets
+          const targets = await client.sendRequest(ExecuteCommandRequest.type, {
+            command: ServerCommands.ListBuildTargets,
+          });
+          const picked = await window.showQuickPick(targets);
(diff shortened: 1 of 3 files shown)


Arthurm1:
Needs https://github.com/scalameta/metals-languageclient/pull/419 to pass checks

src/extension.ts
```diff
@@ -567,6 +567,43 @@ function launchMetals(
...
+            .openTextDocument(uri)
+            .then((textDocument) => window.showTextDocument(textDocument));
+        }
+      }
+
+      registerCommand(`metals.target-info-display`, async (...args: any[]) => {
+        if (args.length > 0) {
+          // get build target name from treeview uri of the form "projects:file:/root/metals/.mtags/?id=mtags3!/_root_/,"
+          const treeViewUri = args[0] as string;
+          const buildTargetRegEx = new RegExp("projects:.*?id=(.*)!.*", "g");
```
tgodzik, line 585: Could we instead parse it as an URI and get the query parameters?

tgodzik (changes requested):
Looks good! Just one small suggestion in regards to parsing, but it can be ignored.

Otherwise, let's bump the language client and default metals server version once the PR in metals is merged (which I will do in a sec.)

src/extension.ts
Arthurm1: It would still only get us halfway there.

A URI example from the treeview would be...
`projects:file:/root/metals/.mtags/?id=mtags3!/_root_/`

Using `URLSearchParams(uri).get("id")` would give...
`mtags3!/_root_/`

So I'd still need to parse for `!` to get `mtags3`

I can do that if you want?

tgodzik: We could then just do `split("!")` so maybe it's a bit clrearer than a regex?

kpodsiad: This splitting part seems too much for me.
IMO in this case regex is simpler and with the proper example in the comment (exists) isn't also that hard to understand.

Arthurm1: I'm easy.  Ideally the treeviewuri would be in the format we already want but it's used to work out how to drill down into classes/methods so I wasn't going to change it.  I tried to extract the target name with `URLSearchParams` but it doesn't like our URI so converts the whole thing to a single query param which is no help. I guess URI != URL

tgodzik: ok then, for me whichever is fine

src/extension.ts
```diff
@@ -567,6 +567,45 @@ function launchMetals(
...
+              command: ServerCommands.ListBuildTargets,
+            })
+            .then(async (...args: any[]) => {
+              const targets = args?.[0] as string[];
+              window.showQuickPick(targets).then((target) => {
+                if (target) {
+                  displayBuildTarget(target);
+                }
+              });
+            });
```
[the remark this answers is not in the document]
Arthurm1: I've made the change.  Works for me.  I don't know typescript well enough to know the implications.

src/extension.ts
Arthurm1: I've changed to `split` + `split`.  Does the job. 🤷

tgodzik (approved):
LGTM!

PR #899, opened 2022-02-24
Add mirror setting to help with coursier set up

tgodzik:
Tested it locally and it seems to work correctly replacing the URL used by coursier boostrap

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,7 +188,7 @@
         },
         "metals.coursierMirror": {
           "type": "string",
-          "markdownDescription": "Optional alternative repository to maven central if maven central is not available."
+          "markdownDescription": "Optional alternative repository to maven central if maven central is not available. For example: `https://jcenter.bintray.com`, which will be used to set to property described [here](https://get-coursier.io/blog/#mirrors)"
         },
         "metals.ammoniteJvmProperties": {
           "type": "array",
@@ -898,7 +898,7 @@
   },
   "dependencies": {
     "ansicolor": "^1.1.100",
-    "metals-languageclient": "0.5.9",
+    "metals-languageclient": "0.5.10",
     "promisify-child-process": "4.1.1",
     "vscode-languageclient": "7.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -186,6 +186,10 @@
           ],
           "markdownDescription": "Optional list of properties to pass along to the Metals server. By default, the environment variable `JAVA_OPTS` and `.jvmopts` file are respected. Each property needs to be a separate item.\n\nExample: `-Dhttps.proxyHost=…`, `-Dhttps.proxyPort=…` or `-Dmetals.statistics=all`"
         },
+        "metals.coursierMirror": {
+          "type": "string",
+          "markdownDescription": "Optional alternative repository to maven central if maven central is not available."
+        },
         "metals.ammoniteJvmProperties": {
           "type": "array",
           "items": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -89,6 +89,7 @@ import { createTestManager } from "./test-explorer/test-manager";
 import { BuildTargetUpdate } from "./test-explorer/types";
 import * as workbenchCommands from "./workbenchCommands";
 import { getServerVersion } from "./getServerVersion";
+import { getMirrorPath } from "./mirrors";
 
 const outputChannel = window.createOutputChannel("Metals");
 const openSettingsAction = "Open settings";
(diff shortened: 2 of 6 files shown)


src/extension.ts
```diff
@@ -577,15 +580,15 @@ function launchMetals(
...
-      registerCommand(
-        `metals.${ServerCommands.DisplayTargetInfo}`,
-        (...args) => {
-          client.sendRequest(ExecuteCommandRequest.type, {
-            command: ServerCommands.DisplayTargetInfo,
-            arguments: args,
-          });
-        }
-      );
+      // registerCommand(
```
tgodzik, line 583: Until https://github.com/scalameta/metals-vscode/pull/893 is merged

package.json
```diff
@@ -186,6 +186,10 @@
           ],
           "markdownDescription": "Optional list of properties to pass along to the Metals server. By default, the environment variable `JAVA_OPTS` and `.jvmopts` file are respected. Each property needs to be a separate item.\n\nExample: `-Dhttps.proxyHost=…`, `-Dhttps.proxyPort=…` or `-Dmetals.statistics=all`"
         },
+        "metals.coursierMirror": {
+          "type": "string",
+          "markdownDescription": "Optional alternative repository to maven central if maven central is not available."
+        },
```
kpodsiad, line 192: What is expected to be set here? It's an url or path to file which is described [here](https://get-coursier.io/blog/#mirrors)? Looking at the the coursier [docs](https://github.com/coursier/coursier/blob/master/modules/paths/src/main/java/coursier/paths/Mirror.java#L26) it's the second

kpodsiad: or no, I have to look at this closer

kpodsiad: Anyway, I think the description lacks information on what do we mean by `alternative repository`

src/mirrors.ts
```diff
@@ -0,0 +1,36 @@
...
+import { getConfigValue, metalsDir } from "./util";
+
+const mirrorProperty = "coursierMirror";
+
+export function getMirrorPath(
+  config: WorkspaceConfiguration
+): string | undefined {
+  const mirrorConfig = getConfigValue<string>(config, mirrorProperty);
+
+  const value = mirrorConfig?.value.trim();
```
kpodsiad, line 13: I would propose:
```typescript
/**
 * @returns path to the mirror config file described in https://get-coursier.io/blog/#mirrors
 */
export function getCoursierMirrorPath(
  config: WorkspaceConfiguration
): string | undefined {
  const mirrorConfig = getConfigValue<string>(config, mirrorProperty);

  const value = mirrorConfig?.value.trim();
  if (mirrorConfig && value && value.length > 0) {
    return writeMirrorFile(value, mirrorConfig.target);
  }
}

function writeMirrorFile(mirrorString: string, target: ConfigurationTarget) {
  const dotMetalsDir = metalsDir(target);
  const file = path.join(dotMetalsDir, "mirror.properties");
  const mirrorContents = [
    "metals.from=https://repo1.maven.org/maven2",
    `metals.to=${mirrorString}`,
  ].join("\n");
  fs.writeFileSync(file, mirrorContents);
  return file;
}
```
There is no any gain in checking if file already exists imo

package.json
kpodsiad: I would add link to the `https://get-coursier.io/blog/#mirrors` and mention that the user has to provide for example only `https://jcenter.bintray.com`

kpodsiad (changes requested):

package.json
tgodzik: :+1: sure!

src/mirrors.ts
tgodzik: Changed!

src/util.ts
```diff
@@ -14,6 +20,39 @@ export type newtype<A, ID extends string> = A & {
...
+      value: defaultValue,
+      target: ConfigurationTarget.Global,
+    };
+  }
+}
+
+export function metalsDir(target: ConfigurationTarget): string {
+  if (target == ConfigurationTarget.Workspace && workspace.workspaceFolders) {
+    const wsDir = workspace.workspaceFolders[0]?.uri.fsPath;
+    return path.join(wsDir, ".metals");
```
tgodzik, line 50: @dos65 I changed the directory for the server version data to `.metals` since this would be a more consistent naming. We already have things like `.bloop` etc.

package.json
kpodsiad: I would add also `https://get-coursier.io/blog/#mirrors`, wouldn't it be helpful?

kpodsiad (approved):

package.json
kpodsiad: What if we can add this information "provided link will be used to set `to` property described [here](https://get-coursier.io/blog/#mirrors)"?

src/util.ts
dos65: Hmm. Technically, it isn't it a problem.
I intentionally named is differently for a case when somebody creates a project inside home dir. I understand that it's weird but who knows, maybe that might be a case for a temporal usage(remote/inside container)

PR #908, opened 2022-03-08
chore: unify filenames

kpodsiad:
We have different file names conventions, I unified them to one - pacscalCase.

Although I wonder if kebab-case isn't more popular approach in frontend ecosystem.


gabro (approved):
👍 makes sense, thanks!

PR #913, opened 2022-03-11
Show file path for workspace symbols

tgodzik:
Previously, we would not show the path to file containing the symbol, which would be problematic if we had the same name for different symbols. Now, we additionally show the path.

image

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1004,10 +1004,7 @@ function launchMetals(
         });
       });
 
-      let workspaceUri: undefined | Uri;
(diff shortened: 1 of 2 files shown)


kpodsiad (approved):
I took the liberty of pushing suggestions directly.
When you're dealing with possible undefined properties you could use optional chaining syntax `?.` - `const workspaceUri = workspace.workspaceFolders?.[0]?.uri;`

tgodzik:
Perfect! Thanks!

PR #918, opened 2022-03-16
Fix search text in jars and add highlighting

Arthurm1:
Quick fix for searching text in jars when using virtual documents.

Adds highlighting to search text.  E.g. here is a search for **Cipher**...
image

@dos65 I felt guilty I broke your stuff

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -966,6 +966,7 @@ function launchMetals(
           client,
           findInFilesProvider,
           findInFilesView,
+          metalsFileProvider,
           outputChannel
         )
       );
--- a/src/findInFiles.ts
+++ b/src/findInFiles.ts
@@ -5,16 +5,19 @@ import {
   OutputChannel,
   Position,
   Range,
+  Selection,
   TextEditorRevealType,
   TreeDataProvider,
   TreeItem,
   TreeItemCollapsibleState,
+  TreeItemLabel,
   TreeView,
   Uri,
   window,
   workspace,
 } from "vscode";
 import { LanguageClient, Location } from "vscode-languageclient/node";
+import { MetalsFileProvider } from "./metalsContentProvider";
 
 class TopLevel {
   constructor(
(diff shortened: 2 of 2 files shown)


dos65:
@Arthurm1 oh, yep. I remembered that we already had [this issue with didOpen](https://github.com/scalameta/metals-vscode/pull/719#discussion_r728276249). So, I think the best solution would be to extend the Location structure in response from the server.

Arthurm1:
@kpodsiad You're a typescript expert - how do I solve this issue with ESLint - it wants `fileContent` to be a `const` but that doesn't compile...

```typescript
var fileContent;
if (uri.scheme == "jar") {
   fileContent = await metalsFileProvider.provideTextDocumentContent(uri);
} else {
   const readData = await workspace.fs.readFile(uri);
   fileContent = Buffer.from(readData).toString("utf8");
}
```

kpodsiad:
You're too kind @Arthurm1! :D
Rule of thumb in js/ts - never use `var`. Use `const` for immutable data and `let` for mutable. So in this case `let` should be enough. 
You can also do some refactor to get rid of `let` like:
```typescipt
      const getFileContent = async () => {
        if (uri.scheme == "jar") {
          return await metalsFileProvider.provideTextDocumentContent(uri);
        } else {
          const readData = await workspace.fs.readFile(uri);
          return Buffer.from(readData).toString("utf8");
        }
      };
      const fileContent = await getFileContent();
```

Arthurm1:
@dos65 I agree - extending the `Location` structure with preview text would halve the work time and cut down on client code but I think that should be saved for another PR since it changes API.  This should at least fix the issue - although it's slow.

I saw that there is a VSCode API proposal for [find text in files](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.findTextInFiles.d.ts) API with what looks like a workable [result class with preview](https://github.com/microsoft/vscode/blob/0f12306b19bd46f1876c4416906d150eae6dc10d/src/vscode-dts/vscode.proposed.textSearchProvider.d.ts#L253) but I've no idea how long these things stay as a proposal of if there is any timeline for them.  The [issue](https://github.com/microsoft/vscode/issues/59924) hasn't changed for a while.

This might also be solved by implementing the [FileSystemProvider API](https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider) for jars - which should put all the searching in VSCode's hands.

PR #926, opened 2022-03-24
added bloop jvm properties to package.json

zmerr:


zmerr:
@tgodzik Now how can I test the Metals code if this does not get merged first?

tgodzik:
You can start the extension in VS Code itself, there is more data about it in https://github.com/scalameta/metals-vscode/blob/main/CONTRIBUTING.md

tgodzik:
Let's update to the newest SNAPSHOT after https://github.com/scalameta/metals/actions/runs/2194854046 is finished

package.json
```diff
@@ -247,6 +247,13 @@
           "type": "string",
           "markdownDescription": "This version will be used for the Bloop build tool plugin, for any supported build tool,while importing in Metals as well as for running the embedded server"
         },
+        "metals.bloopJvmProperties": {
+          "type": "array",
+          "items": {
```
kpodsiad, line 252: @zmerr
We need to add `"scope"` here. Question is which scope should we pick. On the [docs](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema) there are:
```

A configuration setting can have one of the following possible scopes:

application - Settings that apply to all instances of VS Code and can only be configured in user settings.
machine - Machine specific settings that can be set only in user settings or only in remote settings. For example, an installation path which shouldn't be shared across machines.
machine-overridable - Machine specific settings that can be overridden by workspace or folder settings.
window - Windows (instance) specific settings which can be configured in user, workspace, or remote settings.
resource - Resource settings, which apply to files and folders, and can be configured in all settings levels, even folder settings.
language-overridable - Resource settings that can be overridable at a language level.
```

Given that I will pick `application` - bloop settings should be global, defining it per workspace doesn't make any sense. wdyt @tgodzik

zmerr: yes! I was also wondering that the current functionality applies it universally by writing it to the Bloop Global json file, even if the jvm settings are only set for the workspace!

zmerr: so it should be 
`”scope”: “machine-overridable”`?@tgodzik

tgodzik: I think so

kpodsiad:
@zmerr could you update the default Metals version to the newest SNAPSHOT one from https://scalameta.org/metals/docs/?

zmerr:
@kpodsiad I just created a new PR for handling the removal of jvm properties which in some cases were getting ignored!

zmerr:
@kpodsiad I have configured Metals to automatically use the latest SNAPSHOT! I mean each time, when opening VSCode, if there is a new SNAPSHOT, it downloads and uses that!

tgodzik (changes requested):
If you could change the Metals server version to the latest SNAPSHOT version we could merge the PR

tgodzik (approved):
I resolved the conflicts, hopefully didn't mess something up :sweat_smile:

PR #936, opened 2022-04-07
chore: exclude unnecessary files when publish the package

yoshinorin:
Hello.
Thank you for awesome extension :)

## Description

The current published extension includes some unnecessary files.
I updated `.vscodeignore` and exclude them.

## Before

```sh
This extension consists of 1295 files, out of which 603 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore
DONE  Packaged: C:\Users\username\development\metals-vscode-dev\metals-vscode\metals-1.13.10.vsix (1295 files, 2.92MB)
```

## After

```sh
DONE  Packaged: C:\Users\username\development\metals-vscode-dev\metals-vscode\metals-1.13.10.vsix (73 files, 1.27MB)
```

Thank you :)

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,14 +1,12 @@
 .vscode/**
 .vscode-test/**
 out/test/**
-out/testExplorer/**
 out/**/*.map
 src/**
 .gitignore
 tsconfig.json
 vsc-extension-quickstart.md
 .github
-node_modules
 .mergify.yml
 .eslintrc.js
 .prettierignore
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,8 +1,14 @@
 .vscode/**
 .vscode-test/**
 out/test/**
+out/testExplorer/**
 out/**/*.map
 src/**
 .gitignore
 tsconfig.json
 vsc-extension-quickstart.md
+.github
+node_modules
+.mergify.yml
+.eslintrc.js
+.prettierignore
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -10,3 +10,4 @@ vsc-extension-quickstart.md
 .mergify.yml
 .eslintrc.js
 .prettierignore
+.update-version.js


.vscodeignore
```diff
@@ -1,8 +1,14 @@
...
 .vscode-test/**
 out/test/**
+out/testExplorer/**
 out/**/*.map
 src/**
 .gitignore
 tsconfig.json
 vsc-extension-quickstart.md
+.github
+node_modules
```
kpodsiad, line 11: Isn't `node_modules` needed at all?

kpodsiad (commented):
Hi @yoshinorin and thanks for your contribution! I have one question about your PR.

.vscodeignore
yoshinorin: > Isn't node_modules needed at all?

Maybe but not sure.
For example the `omnisharp-vscode` (Made by Microsoft) excludes `node_modules`.

> https://github.com/OmniSharp/omnisharp-vscode/blob/4f7084d2dea4f16b4bc4282c53467e3ecf185575/release.vscodeignore#L17

And as far as I read the official document.

> https://code.visualstudio.com/api/working-with-extensions/bundling-extension#publishing

But not sure, Would you please give me a time?
I will try to check if it works well on my local machine in this weekend.

kpodsiad: According to the https://code.visualstudio.com/api/working-with-extensions/publishing-extension#using-.vscodeignore: `You should ignore all files not needed at runtime`. If I understand this correctly, `node_modules` contains packages that are required during runtime via `require` clause.

Moreover, `Note: Development dependencies listed in devDependencies will be automatically ignored, you don't need to add them to the .vscodeignore file.` suggests that this will automatically exclude some packages from `node-modules` like test dependencies.

For sure `out/testExplorer/**` shouldn't be ignored, it contains code responsible for https://scalameta.org/metals/docs/editors/vscode#test-explorer ;)

I'll also conduct some tests next week, the most urging question is whether `node_modules` should or shouldn't be excluded. I'm certain that after some tests we will achieve optimal subset of ignored files ;)

yoshinorin: @kpodsiad 
I tried to create a package (VSIX) without `node_modules` and install it to my local machine. It didn't work. 
I update `.vscodeignore`.

> For sure out/testExplorer/** shouldn't be ignored, it contains code responsible for https://scalameta.org/metals/docs/editors/vscode#test-explorer ;)

Sorry, I was misunderstood that is test code.

kpodsiad: Tested it locally and currently it works! I also added `.update-version.js` to ignored files, I'll push this change and merge this PR. 
Thanks @yoshinorin for your contribution!

yoshinorin: Thank you for your approval and sorry for the confusion about `node_modules` 😓

PR #944, opened 2022-04-17
Add Interactive semantic info display on jar files

Arthurm1:
See https://github.com/scalameta/metals/pull/3843

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -79,7 +79,11 @@ import {
   startFindInFilesProvider,
 } from "./findInFiles";
 import * as ext from "./hoverExtension";
-import { decodeAndShowFile, MetalsFileProvider } from "./metalsContentProvider";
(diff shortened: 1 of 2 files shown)


Arthurm1:
Just a code tidy - no change in functionality now that the semanticdb file extensions haven't changed

kpodsiad (approved):

kpodsiad:
Although, maybe one thing. I've noticed that `show javap` isn't working on java dependency files. Is it intended? 
It's a stupid action to perform, but can we restrict this behavior? There is `when` condition for command but I don't know if there is a `scheme` property context available there.

Arthurm1:
This tidy is likely covered by https://github.com/scalameta/metals-vscode/pull/963 including a lot of changes to the `when` conditions so I'm closing in favour of that

PR #951, opened 2022-04-23
feat: add install Java17 option when choose JavaDownload

yoshinorin:
As is the title.

## Problem

If we do not apply `modal: true` to `showInformationMessage`, users can not decide which button should select that they want to download. Because, `showInformationMessage` width is narrow and can not contain 4 buttons.

narrow

## Solution (idea)

We can apply `modal: true` to `showInformationMessage`. This option `showInformationMessage` with a dialog. It can contain 4 buttons.

metals-vscode2

How do you think?

Thank you :)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -97,6 +97,7 @@ const openSettingsAction = "Open settings";
 const downloadJava = "Download Java";
 const installJava8Action = "Install Java (JDK 8)";
 const installJava11Action = "Install Java (JDK 11)";
+const installJava17Action = "Install Java (JDK 17)";
 let treeViews: MetalsTreeViews | undefined;
 let currentClient: LanguageClient | undefined;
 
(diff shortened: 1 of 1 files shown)


kpodsiad:
I'm ok with the modal, because having a JDK is crucial for Metals and any problem related to its absence is serious, thus modal makes sense - it forces you to make some decisions.

wdyt @tgodzik

tgodzik (approved):
Looks good, but I would even remove the option for JDK 8 to prepare for a possible deprecation.

dos65:
@tgodzik @kpodsiad 
There is also an option to put these choices into quick pick menu. wdyt?

tgodzik:
> @tgodzik @dos65 what is the consensus? If I understand correctly it is:
> 
> * removing JDK8
> * using modal is fine since having JDK is crucial

That's my opinion here, yes.

kpodsiad:
I want to include this in the current release thus I'll commit changes that we concluded.

dos65 (approved):
:+1:

kpodsiad:
Thanks @yoshinorin for your contribution!

PR #959, opened 2022-04-26
added Bloop Java Home settings

zmerr:


package.json
```diff
@@ -247,6 +247,11 @@
           "type": "string",
           "markdownDescription": "This version will be used for the Bloop build tool plugin, for any supported build tool,while importing in Metals as well as for running the embedded server"
         },
+        "metals.bloopJavaHome": {
```
tgodzik, line 250: I think it would be ok to just use the current javaHome

zmerr: This PR is created in relation with https://github.com/scalameta/metals/pull/3871 and https://github.com/scalameta/metals/pull/3746#discussion_r836537568

PR #962, opened 2022-04-27
chore: move some methods out from `extension.ts`

kpodsiad:


kpodsiad:
maybe I'll do it in other PR

PR #963, opened 2022-04-27
Add Filesystem API to jars

Arthurm1:
Linked to https://github.com/scalameta/metals/pull/3750
and https://github.com/scalameta/metals-languageclient/pull/461

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,8 @@
       {
         "id": "java",
         "extensions": [
-          ".cfr"
+          ".cfr",
+          ".class"
         ]
       },
       {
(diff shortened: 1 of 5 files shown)


Arthurm1:
@kpodsiad Do you know how do fix the [ESLint](https://github.com/scalameta/metals-vscode/runs/6842886771?check_suite_focus=true) errors here?  I'm trying to implement an interface method but I don't need the parameters - hence they're unused.  I thought prefixing with `_` caused the errors to be ignored but that only works with local compilation, not the github actions ESLint stage.

I've copied Microsoft's example method from [here](https://github.com/microsoft/vscode-extension-samples/blob/52679f289a1ba34c27fd1504ae7b0a9275505586/fsprovider-sample/src/fileSystemProvider.ts#L210)

kpodsiad:
@Arthurm1 you try different options which disabled ESLint for unused params - https://eslint.org/docs/rules/no-unused-vars#varsignorepattern. `varsIgnorePattern` might be a could candidate - I used this to suppress errors for `_` [here](https://github.com/scalameta/metals-vscode/pull/1009/files)

Arthurm1:
@kpodsiad That works thanks!

Issue #966, opened 2022-04-29
vs code always Warning "Unable to find a java 11 or java 17 "


yinshurman:
### Describe the bug

when open a sbt project in vscode , there is a warning:
` Unable to find a java 11 or java 17 on this computer. To fix this problem ,update metals.javahome ...`
but I already set that value to my java 11 home (graalvm version):
`/Library/java/JavaVirtualMachines/graalvm/Contents/Home`
Until now  everything works except the warning information always show up ,it is a little annoying 😅
here is my java version:
```
java -version  
openjdk version "11.0.15" 2022-04-19   
OpenJDK Runtime Environment GraalVM CE 22.1.0 (build 11.0.15+10-jvmci-22.1-b06)  
OpenJDK 64-Bit Server VM GraalVM CE 22.1.0 (build 11.0.15+10-jvmci-22.1-b06, mixed mode, sharing)
```

here is my java home 
``` 
echo $JAVA_HOME
/Library/java/JavaVirtualMachines/graalvm/Contents/Home
```

### Expected behavior

No warning !

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

vscode extension: scala(metals) v1.16.0

tgodzik:
Transferred it to metals-vscode, where the issue is actually happening. Thanks for helping out with the investigation!

PR #973, opened 2022-05-06
Create .metals directory if it doesn't exist

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/966


kpodsiad:
It seems like eslint on CI always passes :P I'll take a look at that

tgodzik:
> It seems like eslint on CI always passes :P I'll take a look at that

I think works correctly though? It did fail previously, I just fixed it quickly :D

Issue #971, opened 2022-05-05
Broken dependency on mkdirp


fo5for:
**Describe the bug**

"Install Java" fails due to broken dependency on mkdirp. metals-languageclient requires mkdirp 1.0.3 but the bundled version is 0.5.5.

**To Reproduce** Steps to reproduce the behavior:

1. Install the extension on an environment without Java installed
2. Click on "Download Java" in the "Failed to download Metals" popup
3. Click on any one of the "Install Java" options

**Expected behavior**

Downloads Java and installs it.

**Screenshots**

![](https://user-images.githubusercontent.com/90050474/166903535-b124490b-3a94-48cf-b343-b764cca27749.png)

**Installation:**

- Operating system: Linux
- VSCode version: 1.65.2 (code-server 4.3.0)
- VSCode extension version: 1.16.0
- Metals version: 0.11.5

**Additional context**

It had been working on 1.12.0 and earlier.

PR #975, opened 2022-05-06
Force a later version of mkdirp

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/971

### DIFF
--- a/package.json
+++ b/package.json
@@ -894,6 +894,7 @@
     "@typescript-eslint/eslint-plugin": "^5.22.0",
     "@typescript-eslint/parser": "^5.22.0",
     "@vscode/test-electron": "^2.1.3",
+    "mkdirp": "1.0.3",
     "eslint": "^8.14.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -894,6 +894,7 @@
     "@typescript-eslint/eslint-plugin": "^5.22.0",
     "@typescript-eslint/parser": "^5.22.0",
     "@vscode/test-electron": "^2.1.3",
+    "mkdirp": "1.0.3",
```
kpodsiad, line 897: I don't understand one thing, if some dependency requires mkdirp in `1.x` then why on earth we need to specify this manually. I mean, come js, why are that dumb :(

Or maybe I'm too used to JVM dependency mechanics..

kpodsiad (approved):

package.json
tgodzik: I don't really understand it either, I would think they would use the latest version, but it doesn't :man_shrugging:

PR #1003, opened 2022-06-03
release: Bump versions for v0.11.6

tanishiking:
following https://scalameta.org/metals/docs/contributors/releasing

> check or update enum values of fallbackScalaVersion property in package.json. They should be the same as V.supportedScalaVersions in build.sbt

Should we do this? I just left it as is, since we didn't touch in the previous release?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ The following table shows the status of various features.
 
 ## Requirements
 
-**Java 11 or 17 provided by OpenJDK or Oracle**. Eclipse OpenJ9 is not
+**Java 8, 11, 17 provided by OpenJDK or Oracle**. Eclipse OpenJ9 is not
 supported, please make sure the `JAVA_HOME` environment variable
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,11 +27,11 @@ The following table shows the status of various features.
 
 ## Requirements
 
-**Java 11 or 17 provided by OpenJDK or Oracle**. Eclipse OpenJ9 is not
+**Java 8, 11, 17 provided by OpenJDK or Oracle**. Eclipse OpenJ9 is not
```
tgodzik, line 30: Let's discourage Java 8, we might have forgot to remove it in the docs and only removed here.

tgodzik (approved):
One small thing, but it's totally fine to fix later.

README.md
tanishiking: Fixed!

PR #1009, opened 2022-06-06
feat: show Metals' release notes if server version is updated

kpodsiad:
### Motivation

Currently Metals updates are completely silent and I suspect that most of the users doesn't even know that update just happened. Not everyone has twitter account or is checking [Metals' blog](https://scalameta.org/metals/blog/) on regular basis and this kind of people (I suspect that most of the beginners?) can omit all of those juicy new features which are shipped with the new release. 

It is a goal of this PR to provide some mechanism to inform user about latest changes.

---

### Other solutions

Currently, as far as I know, there is no mechanism for showing extension updates in vscode - see https://github.com/microsoft/vscode/issues/102139. There are some ways to do it, for example I found https://github.com/alefragnani/vscode-whats-new but it seems like its use case is a bit different than what this PR needs (metals-vscode doesn't have any useful changes on its own).  

---

### Goals

Show some indicator that update happened, maybe show embedded release notes in the webview panel?  

No matter what indicator will be chosen it mechanism should:
- not show too often. Ideally only once, after first vscode window start with updated Metals extension
- not show when running metals e.g. on gitpod or other containers which install extension every time
- be able to be disabled in options

---

### Current state

sbt-playground-1654627142733

---

### DIFF
--- a/src/releaseNotesProvider.ts
+++ b/src/releaseNotesProvider.ts
@@ -88,8 +88,7 @@ async function showReleaseNotes(
    * TODO: what about wsl?
    */
   function isRemote(): Either<string, void> {
-    // const isWsl = env.remoteName === "wsl";
-    return env.remoteName == null
+    return env.remoteName == null || env.remoteName !== "wsl"
       ? makeRight(undefined)
       : makeLeft(`is a remote environment ${env.remoteName}`);
   }
--- a/src/releaseNotesProvider.ts
+++ b/src/releaseNotesProvider.ts
@@ -23,7 +23,6 @@ export async function showReleaseNotes(
   outputChannel: vscode.OutputChannel
 ) {
   try {
-    // context.globalState.update(versionKey, "0.11.5");
     const result = await showReleaseNotesImpl(calledOn, context, serverVersion);
     if (result.kind === "left") {
       const msg = `Release notes was not shown: ${result.value}`;
--- a/src/releaseNotesProvider.ts
+++ b/src/releaseNotesProvider.ts
@@ -6,19 +6,25 @@ import { fetchFrom } from "./util";
 import { Either, makeLeft, makeRight } from "./types";
 
 const versionKey = "metals-server-version";
+type CalledOn = "onExtensionStart" | "onUserDemand";
 
 /**
  * Show release notes if possible, swallow errors since its not a crucial feature.
  * Treats snapshot versions like 0.11.6+67-926ec9a3-SNAPSHOT as a 0.11.6.
+ *
+ * @param calledOn determines when this function was called.
+ * For 'onExtensionStart' case show release notes only once (first time).
+ * For 'onUserDemand' show extension notes no matter if it's another time.
  */
-export async function showReleaseNotesIfNeeded(
+export async function showReleaseNotes(
+  calledOn: CalledOn,
   context: ExtensionContext,
   serverVersion: string,
   outputChannel: vscode.OutputChannel
 ) {
   try {
     // context.globalState.update(versionKey, "0.11.5");
-    const result = await showReleaseNotes(context, serverVersion);
+    const result = await showReleaseNotesImpl(calledOn, context, serverVersion);
     if (result.kind === "left") {
       const msg = `Release notes was not shown: ${result.value}`;
       outputChannel.appendLine(msg);
@@ -31,19 +37,19 @@ export async function showReleaseNotesIfNeeded(
   }
 }
 
-async function showReleaseNotes(
+async function showReleaseNotesImpl(
+  calledOn: CalledOn,
   context: ExtensionContext,
   currentVersion: string
 ): Promise<Either<string, void>> {
   const state = context.globalState;
-  state.update(versionKey, "0.11.5");
 
   const remote = isRemote();
   if (remote.kind === "left") {
     return remote;
   }
 
-  const version = getVersion();
+  const version = getVersion(calledOn);
   if (version.kind === "left") {
     return version;
   }
@@ -96,7 +102,7 @@ async function showReleaseNotes(
   /**
    *  Return version for which release notes should be displayed
    */
-  function getVersion(): Either<string, string> {
+  function getVersion(calledOn: CalledOn): Either<string, string> {
     const previousVersion: string | undefined = state.get(versionKey);
     // strip version to
     // in theory semver.clean can return null, but we're almost sure that currentVersion is well defined
@@ -107,8 +113,9 @@ async function showReleaseNotes(
       return makeLeft(msg);
     }
 
-    if (!previousVersion) {
-      // if there was no previous version then show release notes for current cleaned version
+    // if there was no previous version or user explicitly wants to read release notes
+    // show release notes for current cleaned version
+    if (!previousVersion || calledOn === "onUserDemand") {
       return makeRight(currentVersion);
     }
 
--- a/src/releaseNotesProvider.ts
+++ b/src/releaseNotesProvider.ts
@@ -1,0 +1,180 @@
+import { env, ExtensionContext } from "vscode";
+import * as vscode from "vscode";
+import * as semver from "semver";
+import { Remarkable } from "remarkable";
+import http from "https";
+
+const versionKey = "metals-server-version";
+
+// prettier-ignore
+const releaseNotesMarkdownUrl: Record<string, string> = {
+  "0.11.5": "https://raw.githubusercontent.com/scalameta/metals/main/website/blog/2022-04-28-aluminium.md",
+  "0.11.6": "https://raw.githubusercontent.com/scalameta/metals/main/website/blog/2022-06-03-aluminium.md",
+};
+
+/**
+ * Show release notes if possible, swallow errors since its not a crucial feature.
+ * Treats snapshot versions like 0.11.6+67-926ec9a3-SNAPSHOT as a 0.11.6.
+ */
+export async function showReleaseNotesIfNeeded(
+  context: ExtensionContext,
+  serverVersion: string,
+  outputChannel: vscode.OutputChannel
+) {
+  try {
+    await showReleaseNotes(context, serverVersion);
+  } catch (error) {
+    outputChannel.appendLine(
+      `Couldn't show release notes for Metals ${serverVersion}`
+    );
+    outputChannel.appendLine(`${error}`);
+  }
+}
+
+async function showReleaseNotes(
+  context: ExtensionContext,
+  currentVersion: string
+): Promise<void> {
+  const state = context.globalState;
+
+  const version = getVersion();
+  const releaseNotesUrl = version
+    ? releaseNotesMarkdownUrl[version]
+    : undefined;
+  if (isNotRemote() && version && releaseNotesUrl) {
+    await showPanel(version, releaseNotesUrl);
+  }
+
+  async function showPanel(version: string, releaseNotesUrl: string) {
+    const releaseNotes = await getReleaseNotesMarkdown(releaseNotesUrl);
+
+    if (typeof releaseNotes === "string") {
+      const panel = vscode.window.createWebviewPanel(
+        `scalameta.metals.whatsNew`,
+        `Metals ${version} release notes`,
+        vscode.ViewColumn.One
+      );
+
+      panel.webview.html = releaseNotes;
+      panel.reveal();
+
+      // set wrong value for local development
+      // const newVersion = "0.11.5";
+
+      const newVersion = version;
+
+      // Update current device's latest server version when there's no value or it was a older one.
+      // Then sync this value across other devices.
+      state.update(versionKey, newVersion);
+      state.setKeysForSync([versionKey]);
+
+      context.subscriptions.push(panel);
+    }
+  }
+
+  /**
+   * Don't show panel for remote environment because it installs extension on every time.
+   * TODO: what about wsl?
+   */
+  function isNotRemote(): boolean {
+    const isNotRemote = env.remoteName == null;
+    // const isWsl = env.remoteName === "wsl";
+    return isNotRemote;
+  }
+
+  /**
+   *  Return version for which release notes should be displayed
+   *  or
+   *  undefined if notes shouldn't be displayed
+   */
+  function getVersion(): string | undefined {
+    const previousVersion: string | undefined = state.get(versionKey);
+    // strip version to 'major.minor.patch'
+    const cleanVersion = semver.clean(currentVersion);
+    if (cleanVersion) {
+      if (previousVersion) {
+        const compare = semver.compare(cleanVersion, previousVersion);
+        const diff = semver.diff(cleanVersion, previousVersion);
+
+        // take into account only major, minor and patch, ignore snapshot releases
+        const isNewerVersion =
+          compare === 1 &&
+          (diff === "major" || diff === "minor" || diff === "patch");
+
+        return isNewerVersion ? cleanVersion : undefined;
+      }
+
+      // if there was no previous version then show release notes for current cleaned version
+      return currentVersion;
+    }
+  }
+}
+
+async function getReleaseNotesMarkdown(
+  releaseNotesUrl: string
+): Promise<string | { error: unknown }> {
+  const ps = new Promise<string>((resolve, reject) => {
+    http.get(releaseNotesUrl, (resp) => {
+      let body = "";
+      resp.on("data", (chunk) => (body += chunk));
+      resp.on("end", () => resolve(body));
+      resp.on("error", (e) => reject(e));
+    });
+  });
+
+  const text = await ps;
+  // every release notes starts with that
+  const beginning = "We're happy to announce the release of";
+  const notesStartIdx = text.indexOf(beginning);
+  const releaseNotes = text.substring(notesStartIdx);
+
+  // cut metadata yaml from release notes, it start with --- and ends with ---
+  const metadata = text
+    .substring(0, notesStartIdx - 1)
+    .replace("---", "")
+    .replace("---", "")
+    .trim()
+    .split("\n");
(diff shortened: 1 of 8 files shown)


src/releaseProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as semver from "semver";
+
+const versionKey = "metals-server-version";
+
+export class ReleaseProvider {
+  private readonly state: typeof this.context.globalState;
+
+  constructor(
+    private readonly context: ExtensionContext,
+    // TODO switch to extension version maybe?
```
tgodzik, line 12: I think server version is better in this case, but we should limit it to full versions, not snapshots.

src/releaseProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+    const isNewerVersion = this.isNewerVersion(previousValue);
+
+    if (isNewerVersion) {
+      const panel = vscode.window.createWebviewPanel(
+        `scalameta.metals.whatsNew`,
+        `What's New in Metals`,
+        vscode.ViewColumn.One
+      );
+
+      // TODO better content
```
tgodzik, line 37: Maybe we could just embed the full release notes? We could create aliases like `https://scalameta.org/metals/0.11.6` etc.

src/releaseProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+  showWhatsNewIfNeeded() {
+    if (this.shouldShowPanel) {
+      this.showPanel();
+    }
+  }
+
+  private showPanel() {
+    const previousValueOpt: string | undefined = this.state.get(versionKey);
+
+    const previousValue = previousValueOpt || this.currentServerVersion;
```
tgodzik, line 27: So in case we didn't save the version (mechanism was never invoked) we would show the panel for the user? If that is the case I think it's a good idea.

tgodzik (commented):
Great idea! :tada: It would be similar to the VS Code updates that show up, which are really nice to check out. The main difficulty will be to make sure that we don't show it too often.

src/releaseProvider.ts
kpodsiad: Currently if there was no version we didn't show anything. But showing current release notes makes sense, I'll change that

src/releaseProvider.ts
tgodzik: You could also get the `body` property from `https://api.github.com/repos/scalameta/metals/releases/tags/v0.11.6`

This wouldn't require a manual step with the alias (unless we can create them automatically? :thinking: )

kpodsiad: I did the request to the `https://scalameta.org/metals/blog/2022/06/03/aluminium` similar to [this](https://github.com/scalameta/metals-vscode/blob/75fbfe51fbe248c4c13fef9ba8d66201dd10f24b/src/getServerVersion.ts#L75-L85) but I got "301 Moved Permanently". Not sure why though

tgodzik: Fun fact, if you use `https://scalameta.org/metals/blog/2022/06/03/aluminium/` it works correctly with the added `/`

src/releaseProvider.ts
kpodsiad: update: if there was no version before, show release notes for current version

src/releaseProvider.ts
kpodsiad: At least for now I stayed with markdown and parsing it to html - it was already done and I'm not very keen of cutting fragment of html from whole site. But I'm don't have strong opinion on that.

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,180 @@
...
+import { Remarkable } from "remarkable";
+import http from "https";
+
+const versionKey = "metals-server-version";
+
+// prettier-ignore
+const releaseNotesMarkdownUrl: Record<string, string> = {
+  "0.11.5": "https://raw.githubusercontent.com/scalameta/metals/main/website/blog/2022-04-28-aluminium.md",
+  "0.11.6": "https://raw.githubusercontent.com/scalameta/metals/main/website/blog/2022-06-03-aluminium.md",
+};
```
kpodsiad, line 13: The only downside of current approach is a need for adding link to release notes here. I'll try to think of sth better, but maybe this won't be a blocker?

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   *  Return version for which release notes should be displayed
+   *  or
+   *  undefined if notes shouldn't be displayed
+   */
+  function getVersion(): string | undefined {
+    const previousVersion: string | undefined = state.get(versionKey);
+    // strip version to 'major.minor.patch'
+    const cleanVersion = semver.clean(currentVersion);
+    if (cleanVersion) {
```
kpodsiad, line 94: `semver.clean(currentVersion)` will return 'major.minor.patch'. Technically it can return null/undefined if input is invalid but `currentVersion` should be valid Metals version, otherwise downloading it will fail in `extension.ts`

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,180 @@
...
+async function showReleaseNotes(
+  context: ExtensionContext,
+  currentVersion: string
+): Promise<void> {
+  const state = context.globalState;
+
+  const version = getVersion();
+  const releaseNotesUrl = version
+    ? releaseNotesMarkdownUrl[version]
+    : undefined;
```
kpodsiad, line 43: There's a tricky part here, locally published snapshot will be trimmed to 'major.minor.patch' but there won't be released notes for such version -> `state.update(versionKey, newVersion);` won't trigger until there will be.

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,180 @@
...
+        vscode.ViewColumn.One
+      );
+
+      panel.webview.html = releaseNotes;
+      panel.reveal();
+
+      // set wrong value for local development
+      // const newVersion = "0.11.5";
+
+      const newVersion = version;
```
kpodsiad, line 64: I left this on purpose for local development/testing reasons, just swap `//` betwen lines to always get release notes or to overwrite saved version.

kpodsiad (commented):
I think PR is ready for review & testing.
Tests:
- [x] MacOS - done by me
- [ ] Linux
- [ ] Windows (not sure how it'll work with wsl) - I read that wsl is treated as remote(?) - needs verification
- [ ] remote containers (gitpod, codespaces) - I read that they start from scratch every time, better to not show release notes at all than show them every time
- [x] synchronization between editors on local computer - release notes are showed only in first vscode instance, subsequent windows don't get them
- [ ] synchronization between devices - not necessary but nice to have & should work

src/releaseProvider.ts
kpodsiad: Crazy idea:
- obtain `body` from `https://api.github.com/repos/scalameta/metals/releases/tags/`
- extract `2022/06/03/aluminium` part
- switch it to `2022-06-03-aluminium`
- make request to the `https://raw.githubusercontent.com/scalameta/metals/main/website/blog/2022-06-03-aluminium.md`

Doing request to the `https://api.github.com/repos/scalameta/metals/releases/tags/v0.11.6` needs having user-agent set. I'll explore that possibility tommorow.

tanishiking:
@kpodsiad Thanks for kicking off, this is a great idea!

I remember [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) extension shows a welcome/what's the new view if there're updates, so I checked how it is implemented.

- [Sync the previous extension version](https://github.com/gitkraken/vscode-gitlens/blob/bf8cd8f8f63d310d5d4e5fc2a49a88e705bca102/src/extension.ts#L72-L95) using [globalState](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage).
- [compare previous and current ones, if it's downgrade don't show what's new](https://github.com/gitkraken/vscode-gitlens/blob/bf8cd8f8f63d310d5d4e5fc2a49a88e705bca102/src/extension.ts#L250-L256)
- Otherwise, run [ShowWelcomePage command](https://github.com/gitkraken/vscode-gitlens/blob/bf8cd8f8f63d310d5d4e5fc2a49a88e705bca102/src/messages.ts#L158-L171) to open a [welcome.html](https://github.com/gitkraken/vscode-gitlens/blob/b9b39a7407c475a486a2d1b2459196e4a8d58003/src/webviews/apps/welcome/welcome.html) that has a link to release notes
  - But it would be nicer if we can show the release notes directly, instead of showing the same welcome page every time we update the plugin.

Maybe we can store the server version in `globalState` and compare it with the previous ones.

src/releaseNotesProvider.ts
kpodsiad: I should add comment with this

src/releaseNotesProvider.ts
kpodsiad: actually it doesn't work because after version is set execution won't get here :c

src/releaseNotesProvider.ts
tgodzik: Could we fail the build if the map is no updated for the default server version?

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,180 @@
...
+      state.update(versionKey, newVersion);
+      state.setKeysForSync([versionKey]);
+
+      context.subscriptions.push(panel);
+    }
+  }
+
+  /**
+   * Don't show panel for remote environment because it installs extension on every time.
+   * TODO: what about wsl?
```
tgodzik, line 77: What if we instead only show the release notes if the version was set previously. If the version was not set then we ignore it, but we could start from the next release?

So for now empty means user used the previous version. After the next release we treat empty as new installers, WSL etc. where we don't need to show anything (or we could show a more global welcome view)

tgodzik:
> ### Current state
> sbt-playground-1654627142733 [ ](https://user-images.githubusercontent.com/37124721/172457795-a3115d4a-e72e-43b6-9d9f-dcb1c8f2860f.gif) sbt-playground-1654627142733 [ ](https://user-images.githubusercontent.com/37124721/172457795-a3115d4a-e72e-43b6-9d9f-dcb1c8f2860f.gif)

Could we improve the spacing maybe somehow? The text seems kind of squashed.

src/releaseNotesProvider.ts
kpodsiad: I'm afraid that I don't understand 😅

src/releaseProvider.ts
kpodsiad: `getMarkdownLink` retrieves link to markdown file with release notes "automagically", there no need to specify it manually! Code is very ugly, but work, I'll refactor this later if approach is ok. @tgodzik

kpodsiad:
@tgodzik I've increased margin for headers, it should be better now.

sbt-playground-1654933445764

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,243 @@
...
+  serverVersion: string,
+  outputChannel: vscode.OutputChannel
+) {
+  try {
+    // context.globalState.update(versionKey, "0.11.5");
+    const result = await showReleaseNotes(context, serverVersion);
+    if (result.kind === "left") {
+      const msg = `Release notes was not shown: ${result.value}`;
+      outputChannel.appendLine(msg);
+    }
```
kpodsiad, line 25: I've decided to introduce Either like datatype to handle errors/logging. I think that logging one line for every vscode startup won't be pain for users (although some of them are very curious and may ask what are release notes :D). But in case of any errors or problems this may be helpful. Probably testing on wsl/remote container will be easier either.

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,243 @@
...
+ * Translate server version to link to the markdown file with release notes.
+ * @param version clean version in major.minor.patch form
+ * If version has release notes return link to them, if not return nothing.
+ * Sample link to which we're doing request https://api.github.com/repos/scalameta/metals/releases/tags/v0.11.6.
+ * From such JSON obtain body property which contains link to the blogpost, but what's more important,
+ * contains can be converted to name of markdown file with release notes.
+ */
+async function getMarkdownLink(
+  version: string
+): Promise<Either<string, string>> {
```
kpodsiad, line 140: This function might be seems magic but as long as Github Release contains link to blogpost this will work and there is no need to specify release notes manually.

tanishiking:
The change looks good to me! Let me playing around this change a bit :)

kpodsiad:
I think `author` part needs some overhaul. Maybe simple prefix like `Authored by: <author name with link to twitter (current content)>`. What do you think?

src/releaseNotesProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Don't show panel for remote environment because it installs extension on every time.
+   * TODO: what about wsl?
+   */
+  function isRemote(): Either<string, void> {
+    // const isWsl = env.remoteName === "wsl";
+    return env.remoteName == null
+      ? makeRight(undefined)
+      : makeLeft(`is a remote environment ${env.remoteName}`);
+  }
```
kpodsiad, line 95: @Arthurm1 could you try this PR out on Windows & wsl? I'm not sure how wsl is being treated and I don't have windows myself

Arthurm1: I've run
```
yarn install
yarn build
```
and then installed `metals-1.17.2.vsix` but when I try to start Metals I get...

image

What step have I missed out?  Why would `remarkable` be missing as `yarn build` works

kpodsiad: I've run this only in debug extension mode so I'm not sure if this error happens also on other systems.
Try to remove node_moduels and run yarn install again

tgodzik: I can also test that out. You can usually test it out by running inside VS Code, but I wonder if the issue you are hitting on is soemthing to do with wrongly specifying the dependencies? @kpodsiad did you test it out this was on mac?

src/releaseNotesProvider.ts
tgodzik: I was just thinking that maybe the first time user install Metals and no version was set up, we would run only the welcome view. (or nothing at all). This is however a problem for when we want to show the release notes first time for even existing users. So we could at first assume that no one is the first user and always show release notes if no metals version was saved globally. And after the next release we can safely revert the bahaviour to only show release notes if not server version was specified previously.

This is however just a vague idea, so not a blocker at all for me.

src/releaseNotesProvider.ts
Arthurm1: In debug extension mode I can start Metals and get...
```
2022.06.11 13:27:59 INFO  Started: Metals version 0.11.5-SNAPSHOT in workspace '/root/metals' for client Visual Studio Code 1.68.0.
2022.06.11 13:28:03 INFO  time: initialize in 3.66s
Initializing Scala Debugger
Release notes was not shown: is a remote environment wsl
2022.06.11 13:28:03 INFO  Flyway Community Edition 8.5.8 by Redgate
```

Is that what's wanted?

Arthurm1:
Is there a command to bring up the release notes?  Maybe add `show release notes` in `Help and Feedback`?

image

kpodsiad:
> Is there a command to bring up the release notes?  Maybe add `show release notes` in `Help and Feedback`?
> 
> image
> 

Great idea @Arthurm1. I can both command and button in 'Help and Feedback'.

src/releaseNotesProvider.ts
kpodsiad: > I was just thinking that maybe the first time user install Metals and no version was set up, we would run only the welcome view.

For welcome view, if I understand correctly, there is [special contribution point](https://code.visualstudio.com/api/references/contribution-points#contributes.walkthroughs) so I think we can show both welcome and release notes.

IMO there is no harm in showing release notes for new people, this might even encourage them to visit scalemta.org/metals so if you're OK, like you said, I'll leave this as it is.

src/releaseNotesProvider.ts
kpodsiad: Ok, I found what was wrong. I accidentally put `remarkable` at `devDependencies` and they're are present only at dev development but they aren't included in bundle create via `yarn build`. Now it should work!

@Arthurm1 in which output tab did you have such logs? I couldn't find anything useful and without your's scrrenshot it would take much longer to find issue. Thanks for it!

@Arthurm1 I changed logic to show release notes on `wsl`. Could you try to see if it works and then, comment `state.update(versionKey, "0.11.5");` (line 39) and see if release notes are showed only once and subsequent windows don't display them.

Arthurm1: I can build and install it now.

No release notes are shown and I get this...
image


Same thing after commenting out `state.update(versionKey, "0.11.5");`

This looks correct to me because the code is...
image
So it will never get far enough in `showReleaseNotes` to show the release notes

Is it supposed to show the release notes for WSL?

kpodsiad: > Is it supposed to show the release notes for WSL?

Yes, we should show them. 

I realized that condition in `isRemote` was broken, now it should work correctly.

Arthurm1: Now I get release notes and if I comment out line 39 then I get `Release notes was not shown: do not show release notes for an older version`

So looking good.

kpodsiad:
I've added a command `Metals: show release notes` which will allow to open webview with current release notes.

src/releaseNotesProvider.ts
tgodzik: Ok sure, let's see if it causes any issues for people (it shouldn't really)

tgodzik (approved):
Looks good! Let's just rebase on the main branch

kpodsiad:
I'm merging, pre-release it's time to do your job :D

PR #1013, opened 2022-06-07
feature: Add scalafix run command and scalafix dependencies setting

tgodzik:
For running scalafix I decided on using `ctrl+alt+shift+o` shortcut since it's one step from organize imports and should be pretty intuitive (do more than organize import so I add shift)

As for the dependency settings I was thinking about using an object first, but that would require each rule to have it's dependency (but there might be multiple rules in a single dep) and it's easier to represent an Array in the settings UI.

Related to https://github.com/scalameta/metals/pull/3996

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
       "properties": {
         "metals.serverVersion": {
           "type": "string",
-          "default": "0.11.6",
+          "default": "0.11.6+67-926ec9a3-SNAPSHOT",
           "markdownDescription": "The version of the Metals server artifact. Requires reloading the window.  \n\n**VS Code extension version is guaranteed to work only with the default version, change if you know what you're doing**"
         },
         "metals.serverProperties": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -243,6 +243,14 @@
...
           "markdownDescription": "Optional custom path to the .scalafix.conf file.\n\nShould be an absolute path and use forward slashes `/` for file separators (even on Windows)."
         },
+        "metals.scalafixRulesDependencies": {
+          "type": "array",
+          "scope": "machine",
+          "items": {
+            "type": "string"
+          },
+          "markdownDescription": "List of Scalafix rules dependencies in case they are not available by default in Metals. For example: `com.github.liancheng::organize-imports:0.6.0`, `::` indicates Scala dependency"
+        },
```
kpodsiad, line 253: I haven't looked at metals' PR yet so I have a few questions arisen.

> they are not available by default in Metals

How one can know if rule is available by default?

> For example: `com.github.liancheng::organize-imports:0.6.0`, `::` indicates Scala dependency

not sure how to understand the second part with `::`

tgodzik: > How one can know if rule is available by default?
We should probably list them somewhere :thinking: 

> not sure how to understand the second part with ::

Maybe I tried exaplaining instead of just linking to the coursier convention.

dos65 (approved):

package.json
kpodsiad: Listing known rules is a good idea.



> > not sure how to understand the second part with ::
> 
> Maybe I tried exaplaining instead of just linking to the coursier convention.

OMG, now I realized what you meant 😂  I read `Scala dependency` and then interpreted `::` a List member haha. It's probably better to provide link which explains what `::` in the context of dependencies is.

tgodzik: I modified the explanation a bit. I will add the list of built in rules later on, since I haven't yet added and community ones,

kpodsiad (approved):

PR #1015, opened 2022-06-10
Update metals-languageclient and vscode-languageclient

tanishiking:
## Update metals-languageclient
To upgrade vscode-languageclient to 8.0.1, it was needed to remove
unused interface from metals-languageclient
see: https://github.com/scalameta/metals-languageclient/pull/459

Also, this version introduces some changes.

## Update vscode-languageclient
see: https://github.com/microsoft/vscode-languageserver-node/blob/95e411de3d10c0fa1e0371b84e88f482682a9170/README.md#3170-protocol-800-json-rpc-800-client-and-800-server

> cleanup of client start and stop methods. Both methods now return a promise since these methods are async. This is a breaking change since start returned a disposable before. Extensions should now implement a deactivate function in their extension main file and correctly return the stop promise from the deactivate call. As a consequence the onReady() got removed since extensions can await the start()

metals-vscode already implement `deactivate` method that returns promise from `stop()`, so it's all good

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -507,9 +507,7 @@ function launchMetals(
     )
   );
 
-  context.subscriptions.push(client.start());
-
-  return client.onReady().then(
+  return client.start().then(
     () => {
       const doctorProvider = new DoctorProvider(client);
       let stacktrace: WebviewPanel | undefined;
--- a/package.json
+++ b/package.json
@@ -909,9 +909,9 @@
   },
   "dependencies": {
     "ansicolor": "^1.1.100",
-    "metals-languageclient": "0.5.15",
+    "metals-languageclient": "0.5.16",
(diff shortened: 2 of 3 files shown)


src/extension.ts
```diff
@@ -507,9 +507,7 @@ function launchMetals(
     )
   );
 
-  context.subscriptions.push(client.start());
-
-  return client.onReady().then(
```
tanishiking, line 512: https://github.com/microsoft/vscode-languageserver-node/blob/95e411de3d10c0fa1e0371b84e88f482682a9170/README.md#3170-protocol-800-json-rpc-800-client-and-800-server

kpodsiad (approved):
LGTM

tanishiking:
Thanks !

PR #1035, opened 2022-06-21
feature: Add reset notifications command

tgodzik:
Related to https://github.com/scalameta/metals/pull/4049

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -872,6 +872,13 @@ function launchMetals(
         });
       });
 
+      registerCommand(`metals.reset-notifications`, (args = []) => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -872,6 +872,13 @@ function launchMetals(
         });
       });
 
+      registerCommand(`metals.reset-notifications`, (args = []) => {
```
kpodsiad, line 875: is `= []` needed?

kpodsiad (approved):

src/extension.ts
tgodzik: Probably not :sweat_smile: but didn't want to angry the typescript gods just in case

PR #1039, opened 2022-06-25
chore: show release notes on user demand on remote env

kpodsiad:


src/releaseNotesProvider.ts
```diff
@@ -43,16 +43,16 @@ async function showReleaseNotesImpl(
...
-  const version = getVersion(calledOn);
+  const version = getVersion();
   if (version.kind === "left") {
     return version;
   }
 
+  const remote = isRemote();
+  if (calledOn === "onExtensionStart" && remote.kind === "left") {
+    return remote;
+  }
```
kpodsiad, line 54: Makes more sense to check version first, only after remote. It yields better message: `Release notes was not shown: do not show release notes for an older version/already seen version`

src/releaseNotesProvider.ts
```diff
@@ -128,7 +132,9 @@ async function showReleaseNotesImpl(
 
     return isNewerVersion
       ? makeRight(cleanVersion)
-      : makeLeft("do not show release notes for an older version");
+      : makeLeft(
+          "do not show release notes for an older version/already seen version"
```
tanishiking, line 136:
```suggestion
          `do not show release notes for switching to the older or same version, previous version: ${previousVersion}, current version: ${currentVersion}`
```

What do you think about this? I personally think "same version" makes more sense than "already seen version" (if I understand correctly). Also, it would be helpful if there's version information in the message.

src/releaseNotesProvider.ts
```diff
@@ -43,16 +43,16 @@ async function showReleaseNotesImpl(
...
-  }
-
-  const version = getVersion(calledOn);
+  const version = getVersion();
   if (version.kind === "left") {
     return version;
   }
 
+  const remote = isRemote();
+  if (calledOn === "onExtensionStart" && remote.kind === "left") {
```
tanishiking, line 52:
```suggestion
  if ((calledOn === "onExtensionStart" && remote.kind === "left") || calledOn === "onUserDemand") {
```

I prefer to check `calledOn === "onUserDemand"` outside of `isRemote` function, because `"onUserDemand"` is irrelevant to the remote environment.

src/releaseNotesProvider.ts
```diff
@@ -89,19 +89,23 @@ async function showReleaseNotesImpl(
...
+   * Show panel:
+   * - on user demand
+   * - for wsl
+   * - when it's not a remote env
    */
   function isRemote(): Either<string, void> {
-    return env.remoteName == null || env.remoteName === "wsl"
+    return calledOn === "onUserDemand" ||
+      env.remoteName == null ||
+      env.remoteName === "wsl"
```
tanishiking, line 100:
```suggestion
    return
      env.remoteName == null ||
      env.remoteName === "wsl"
```

dit

tanishiking (commented):
I left some minor comments, but otherwise LGTM :)

src/releaseNotesProvider.ts
kpodsiad: Release notes should be shown only once for particular version (vs code keeps track of the newest one) and not shown for an older version than was already seen.
This message (it's more a debug one) will be logged to the output tab when any of aforementioned conditions are false:
- release notes were showed for current version
- user is using version older than the last one for which notes were showed

ckipp01: Ah ok, so something like

```
not showing release notes since they've already been seen for your current version
```

???

tanishiking: 👍 to @ckipp01 's comment, my suggestion was too much related to the internal behavior that shouldn't be exposed to  end users.

tanishiking (approved):
Except the log message, LGTM :+1:

Issue #1055, opened 2022-07-11
"Starting Metals Server" keeps on appearing


dzx-dzx:
**Describe the bug**
The same as #1028 .

**Screenshots**
![图片](https://user-images.githubusercontent.com/31068367/178250604-7d9c37a3-1107-4926-8b20-c95c50104b48.png)

**Installation:**

- Operating system: WSL2: Ubuntu 20.2 on Windows 11
- VSCode version: 1.69.0
- VSCode extension version: v1.18.0
- Metals version: 0.11.7

**Additional context**
Logs:
```
Java home: /usr/lib/jvm/java-17-openjdk-amd64
Metals version: 0.11.7
2022.07.11 18:53:12 INFO  logging to file ***/***/.metals/metals.log
2022.07.11 18:53:12 INFO  Started: Metals version 0.11.7 in workspace '***/***' for client Visual Studio Code 1.69.0.
2022.07.11 18:53:12 INFO  time: initialize in 0.88s
Initializing Scala Debugger
2022.07.11 18:53:13 INFO  Flyway Community Edition 8.5.13 by Redgate
2022.07.11 18:53:13 INFO  See what's new here: https://flywaydb.org/documentation/learnmore/releaseNotes#8.5.13
2022.07.11 18:53:13 INFO  
2022.07.11 18:53:13 INFO  Database: jdbc:h2:file:***/***/.metals/metals (H2 2.1)
2022.07.11 18:53:13 WARN  Flyway upgrade recommended: H2 2.1.214 is newer than this version of Flyway and support has not been tested. The latest supported version of H2 is 2.1.210.
2022.07.11 18:53:13 INFO  Successfully validated 3 migrations (execution time 00:00.043s)
2022.07.11 18:53:13 INFO  Current version of schema "PUBLIC": 3
2022.07.11 18:53:13 INFO  Schema "PUBLIC" is up to date. No migration necessary.
2022.07.11 18:53:13 INFO  Attempting to connect to the build server...
2022.07.11 18:53:13 INFO  Bloop uses /usr/lib/jvm/java-17-openjdk-amd64 defined at ***/.bloop/bloop.json
2022.07.11 18:53:12 INFO  skipping build import with status 'Requested'
2022.07.11 18:53:18 INFO  tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty file at ***/***/.metals/bsp.trace.json or ***/.cache/metals/bsp.trace.json
[...]

PR #1056, opened 2022-07-11
fix: add timeouts to http request

kpodsiad:
closes https://github.com/scalameta/metals-vscode/issues/1055, at least should ;)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -121,7 +121,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
   detectLaunchConfigurationChanges();
   configureSettingsDefaults();
 
-  return window.withProgress(
+  await window.withProgress(
     {
       location: ProgressLocation.Window,
       title: `Starting Metals server...`,
@@ -132,17 +132,18 @@ export async function activate(context: ExtensionContext): Promise<void> {
       try {
         const javaHome = await getJavaHome(getJavaHomeFromConfig());
         await fetchAndLaunchMetals(context, javaHome, serverVersion);
-        await showReleaseNotes(
-          "onExtensionStart",
-          context,
-          serverVersion,
-          outputChannel
-        );
       } catch (err) {
         outputChannel.appendLine(`${err}`);
         showMissingJavaMessage();
       }
     }
+  );
+
+  await showReleaseNotes(
+    "onExtensionStart",
+    context,
+    serverVersion,
+    outputChannel
   );
 }
 
--- a/src/util.ts
+++ b/src/util.ts
@@ -111,12 +111,17 @@ export async function fetchFrom(
 ): Promise<string> {
   const requestOptions: http.RequestOptions = { timeout: 5000, ...options };
   const promise = new Promise<string>((resolve, reject) => {
-    http.get(url, requestOptions, (resp) => {
-      let body = "";
-      resp.on("data", (chunk) => (body += chunk));
-      resp.on("end", () => resolve(body));
-      resp.on("error", (e) => reject(e));
-    });
+    http
+      .get(url, requestOptions, (resp) => {
+        let body = "";
+        resp.on("data", (chunk) => (body += chunk));
+        resp.on("end", () => resolve(body));
+        resp.on("error", (e) => reject(e));
+      })
+      .on("error", (e) => reject(e))
+      .on("timeout", () =>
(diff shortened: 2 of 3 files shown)


kpodsiad:
@dzx-dzx are you able to install this version locally? Is as simple as:
- clone repo
- `yarn install`
- `yarn build`
- there will appear something like `metals-1.18.2.vsix` 
- install this as extension
Screenshot 2022-07-11 at 18 32 14

It contains fix + should tell us more through logs.

src/util.ts
```diff
@@ -109,8 +109,9 @@ export async function fetchFrom(
   url: string,
   options?: http.RequestOptions
 ): Promise<string> {
+  const requestOptions: http.RequestOptions = { ...options, timeout: 5000 };
```
tanishiking, line 112: [nit]
```suggestion
  const requestOptions: http.RequestOptions = { timeout: 5000, ...options };
```

so we can specify the timeout millis from callee of `fetchFrom`

src/util.ts
```diff
@@ -109,8 +109,9 @@ export async function fetchFrom(
...
   const promise = new Promise<string>((resolve, reject) => {
-    http.get(url, options || {}, (resp) => {
+    http.get(url, requestOptions, (resp) => {
```
tanishiking, line 114: This is not related to this PR, but we have to put `on("error", (e) => {  ... } )` outside of `http.get` like

```ts
    http
      .get(url, requestOptions, (resp) => {
        let body = "";
        resp.on("data", (chunk) => (body += chunk));
        resp.on("end", () => resolve(body));
      })
      .on("error", (e) => reject(e))
```

`resp.on("error", (e) => reject(e))` seems never reject on error.

tanishiking: Also, it would be a good idea to show something on timeout 👍 

```ts
    http
      .get(url, requestOptions, (resp) => {
         // ...
      })
      .on("error", (e) => reject(e))
      .on("timeout", () => reject("Request timed out"));
```

tanishiking (changes requested):
I left some comments around http requests.

---off-topic below---
We might want to use `node-fetch` or `axios` (maybe `node-fetch` as we already use it in `metals-languageclient`) instead of nodejs's `https` module (in another PR), as its a plain API and we have to handle many things quite manually (like `(chunk) => body += chunk`) 😆

tanishiking:
FYI, open the project with VSCode, and click `F5` will build the extension and start it in debug mode :)

src/util.ts
kpodsiad: oh yeah, I forgot that order matters

src/util.ts
kpodsiad: good points, I left `resp.on("error", (e) => reject(e));` though. I guess it feels more safe to react on this kind of event, although I'm not sure what will cause it to happen

src/extension.ts
```diff
@@ -132,18 +132,19 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
       }
     }
   );
+
+  await showReleaseNotes(
+    "onExtensionStart",
+    context,
+    serverVersion,
+    outputChannel
+  );
```
kpodsiad, line 147: Ok, moving this out of `window.withProgress` should solve "Starting Metals server..." forever issue, but it's unrelated to the timeout thing.

dzx-dzx:
@kpodsiad The message did disappear but there wasn't release notes either. Should I share the logs?

kpodsiad:
@dzx-dzx is there sth along lines `Release notes was not shown: <reason>` after startup? Usually it's at the beginning of the output tab. You can also execute command `Metals: show release notes` and see what outcome it will yield (release notes, logs).

dzx-dzx:
> @dzx-dzx is there sth along lines `Release notes was not shown: <reason>` after startup? Usually it's at the beginning of the output tab. You can also execute command `Metals: show release notes` and see what outcome it will yield (release notes, logs).

Oh yes, it tells of my misconfigured proxy setting. Now the release notes show up. Thank you!

kpodsiad:
> Oh yes, it tells of my misconfigured proxy setting. Now the release notes show up. Thank you!

Glad to hear that, thanks for reporting issue.

tanishiking (approved):
LGTM

tgodzik (approved):
LGTM! Thanks for looking into this!

PR #1060, opened 2022-07-12
chore: refactor install java out of extension.ts

kpodsiad:
It's always less in `extension.ts`


tgodzik (approved):
You just need to rebase, but LGTM otherwise!

PR #1018, opened 2022-06-13
Revert "Merge pull request #1015 from tanishiking/bump-metals-languag…

tgodzik:
…e-client"

This reverts commit a9dd72cc30d8383e27f31eb1b92d08a54223adba, reversing
changes made to c7bec8d869daa3742ea3a30c73bbd54391de392b.

There is an issue in the language client discussed here https://github.com/scalameta/metals-vscode/issues/1016

Issue #1016, opened 2022-06-11
"Format on save" deletes recently written code


kubukoz:
**Describe the bug**

In the latest release, I started noticing some formatting issues that I never saw before: When used with format-on-save, saving a file will remove some code I just wrote.

In case it's not possible to demonstrate with a code example, here's a recording:

https://user-images.githubusercontent.com/894884/173199758-9acb55e4-9adb-4faf-a405-993ce618c7e7.mov

Steps (doesn't always work):

1. Enable format on save
2. Have some working code like this

```scala
object Main extends App {

  val hello = 42

  identity()
}
```

3. Save the file. Now put the cursor before `identity` and add an indent with `<tab>`. Go to the paranthesis and write something, even `???`. You should have something like this:

```scala
object Main extends App {

  val hello = 42

  	identity(???)
}
```

4. Save the file. This will (sometimes) delete the entirety of `???` or part of it.

Versioning note: the server version seems to be `0.11.6+67-926ec9a3-SNAPSHOT`. It seems to be set automatically by the most recent vscode extension version (1.17.3). When I set `0.11.6` explicitly, it still doesn't work. The extension worked fine with its defaults on 1.17.1, so it seems like an extension issue. 1.17.2 also seems to be affected.

**Expected behavior**

Code is formatted but no non-whitespace is removed.

**Installation:**

- Operating system: macOS
- VSCode version: 1.67.2
- VSCode extension version: 1.17.3
- Metals version: (found in VSCode settings, under `metals.serverVersion`): 0.11.6 / 0.11.6+67-926ec9a3-SNAPSHOT (doesn't seem to matter)

**Additional context**

**Search terms**

scalafmt

tanishiking:
- `FormattingProvider` receive the input from buffer https://github.com/scalameta/metals/blob/e354275ffc5351562593aaf8cb9c095374d6d273/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala#L107
- buffers will be updated on `didChange` and `didSave`
  - https://github.com/scalameta/metals/blob/e354275ffc5351562593aaf8cb9c095374d6d273/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1228
  - https://github.com/scalameta/metals/blob/e354275ffc5351562593aaf8cb9c095374d6d273/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1203

[...]

PR #1063, opened 2022-07-14
Bump metals-languageclient to 0.5.17 and vscode-languageclient to 8.0.2

tanishiking:
Previously, we reverted the updates to those libraries in https://github.com/scalameta/metals-vscode/pull/1018 because of the [issue](https://github.com/scalameta/metals-vscode/issues/1016).

Now, the fix for the cause is in vscode-languageclient 8.0.2 and we should be able to safe to update those libraries.

- PR: https://github.com/microsoft/vscode-languageserver-node/pull/972
- release note: https://github.com/microsoft/vscode-languageserver-node#3172-protocol-802-json-rpc-802-client-and-802-server

I'm going to play around with this build for a while before merging this PR :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -956,11 +956,11 @@
   },
   "dependencies": {
     "ansicolor": "^1.1.100",
-    "metals-languageclient": "0.5.15",
+    "metals-languageclient": "0.5.17",
     "promisify-child-process": "4.1.1",
     "remarkable": "^2.0.1",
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM! Thanks for taking care of it!

tanishiking:
I played around with this build for a while and I didn't see any issue with it :tada:

Issue #1071, opened 2022-07-19
update Metals version to the newest one through dialogue only updates in User tab.


vzmerr:
### Describe the bug

To reproduce, put `0.11.8-SNAPSHOT` in Metals Server Version in both Workspace and User tabs of VSCode. Then, totally close VSCode. (Do not `Relead Window` instead.) And reopen it. 
A dialogue would appear asking if you want to update the Metals Version to the newest one. And you should confirm. Then reload and check the Metals Version setting:
It is only updated in the User tab. But remains as `0.11.8-SNAPSHOT` in the Workspace tab, which overrides the User one.

### Expected behavior

choosing to update the Metals version to the newest one, through the dialogue in VSCode, should change the setting also in the Workspace tab for the result to be effective.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.8-SNAPSHOT

tanishiking:
I transferred it to `metals-vscode`, since this is a problem only with the VS Code extension :)

tanishiking:
Indeed, metals-vscode updates only one configuration (of User and Workspace configurations)

- Get the configuration value and target from `User` **or** `Workspace`
  - https://github.com/scalameta/metals-vscode/blob/1822de60091af184725383b0476c7f883cbab53e/src/getServerVersion.ts#L33-L36
- Ask the user and update the config
  - https://github.com/scalameta/metals-vscode/blob/1822de60091af184725383b0476c7f883cbab53e/src/getServerVersion.ts#L57-L61

we might want to update both `User` and `Workspace` of the server version configuration if both of them has outdated version 🤔

PR #1076, opened 2022-07-25
refactor: Do not use .metals for checking the metals version

tanishiking:
This PR replaces `$PROJECT_DIR/.metals/versions-meta.json` and `$HOME/.metals/versions-meta.json"` to `workspaceState` and `globalState` respectively.

I remember someone has trouble because Metals creates `.metals` in their home directory (on Discord, but I couldn't find the link to the discussion).
Anyway, it's best practice to use `context.globalState` (or `workspaceState`) for managing the state of extension, and it makes easy to test (in the future) 👍 

Let me test locally for a while before merging.


dos65:
@tanishiking I initially wanted to store it in inside settings but vscode doesn't allow writing into unknown fields. Every setting must be declared in `package.json`. There is an option to rename global directory from `.metals` to smth else

tanishiking:
> I initially wanted to store it in inside settings but vscode doesn't allow writing into unknown fields. Every setting must be declared in package.json.

Are you talking about the coursier mirror settings? https://github.com/scalameta/metals-vscode/blob/bfc4de113c688338d6776a1e330d4ddedd742c09/src/mirrors.ts#L28-L31
https://scalameta.org/metals/docs/troubleshooting/proxy/

I don't have a context why metals-vscode has settings to edit the coursier mirror properties, but I'm wondering why users want to configure it via metals-vscode instead of setting global coursier mirrors properties by themselves.

tgodzik:
> I don't have a context why metals-vscode has settings to edit the coursier mirror properties, but I'm wondering why users want to configure it via metals-vscode instead of setting global coursier mirrors properties by themselves.

It's just a bit more convenient for the users to have it metals instead of manually searching how to do it.

src/getServerVersion.ts
```diff
@@ -37,7 +42,11 @@ async function validateCurrentVersion(
 
   const checkForUpdate = async () => {
     if (suggestUpgradeSetting?.value) {
-      return needCheckForUpdates(serverVersion, suggestUpgradeSetting.target);
+      return needCheckForUpdates(
+        serverVersion,
+        suggestUpgradeSetting.target,
```
kpodsiad, line 47: If I understand correctly, depending on where user set `checkForUpdates` switch then version will be stored in that scope (global vs workspace), I'm right? 
Here, because it's a simple boolean there are no such cases like with server version itself (where empty string in workspace is treated as a valid value and global setting is discarded)

tanishiking: > If I understand correctly, depending on where user set checkForUpdates switch then version will be stored in that scope (global vs workspace), I'm right?

Yes, you're right!

> Here, because it's a simple boolean there are no such cases like with server version itself

What do you mean?

Do you mean, we should check for updating both Global and Workspace settings, even if `suggetUpgradeSetting` is `true` only in `Global` or `Workspace`?
In that case, I agree, but I think we can work on that in another issue / PR https://github.com/scalameta/metals-vscode/issues/1071

tanishiking:
> It's just a bit more convenient for the users to have it metals instead of manually searching how to do it.

Ok, it would be awesome if we could completely remove the usage of `~/.metals` directory for `metals-vscode`, but we can work on that in another issue / PR :+1:
(I personally think, vscode settings shouldn't change the global behavior (coursier mirror setting in this case). I think people won't expect a change in the vscode setting create/modify the global coursier setting though...)

tanishiking:
Also added unit test (while refactoring the project structure) for `needCheckForUpdate` https://github.com/tanishiking/metals-vscode/pull/82 
I'll open a PR against this repository after this PR has merged.

tgodzik:
> (I personally think, vscode settings shouldn't change the global behavior (coursier mirror setting in this case). I think people won't expect a change in the vscode setting create/modify the global coursier setting though...)

I understand what you mean, but it was really complicated for people to find the right place for where to put the file and it is far from obvious. Maybe it would be better to have mirror as a parameter in that case, but I think last time I checked that was not possible. And this was hard to get working right :sweat_smile:

dos65 (approved):
Looks good!

tgodzik (approved):

tanishiking:
> it was really complicated for people to find the right place for where to put the file and it is far from obvious. Maybe it would be better to have mirror as a parameter in that case, but I think last time I checked that was not possible. And this was hard to get working right 😅

I see, thank you for sharing the background!
Maybe we can keep those settings without placing proxy files in `~/.metals` directory, I'll take a look!

src/getServerVersion.ts
kpodsiad: @tanishiking I honestly don't remember what I was thinking and what I meant 😅

tanishiking: haha, no worries! sorry for my delayed reply

PR #1092, opened 2022-08-03
extend allowed file extesions for triggering custom symbol search

dos65:


kpodsiad (approved):
👍

kpodsiad:
Hmm, I wonder if we can extend it to `javap` and `semanticDB` files 🤔

PR #1105, opened 2022-08-15
bugfix: Fix issue with * not being inserted automatically

tgodzik:
Previously, the rule for automatically adjusting whitespace for Scala 3 (due to issues with optional braces) was catching almost all scaladoc rules. Now, I moved it last (only first rule is used) and also added $ at the end of the regex.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1138,17 +1138,6 @@ function enableScaladocIndentation() {
         action: { indentAction: IndentAction.Indent },
       },
       {
-        // stop vscode from indenting automatically to last known indentation
-        beforeText: /^\s*/,
-        /* we still want {} to be nicely split with two new lines into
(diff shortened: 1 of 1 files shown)


tanishiking (commented):
Could you elaborate on the change? I couldn't catch what was the problem and what this change fixed.

tgodzik:
> Could you elaborate on the change? I couldn't catch what was the problem and what this change fixed.

If you try to press enter in a scaladoc before the end of line `*` would not be added, because the other rule would be invoked. Now, adding `*` should work correctly in all cases.

tanishiking (approved):
> If you try to press enter in a scaladoc before the end of line * would not be added

Ahh, thank you for the explanation! Nice catch, LGTM!

PR #1113, opened 2022-08-16
refactor: delete unnecessary file

yoshinorin:
The `protocol.ts` file is empty (when https://github.com/scalameta/metals-vscode/commit/5b2644eaf2df396754d6b397adc453e01ebec7a5) and it is never referred from other files.
So, I think it is unnecessary.

Thank you :)


dos65 (approved):
Thanks!

PR #1121, opened 2022-08-22
chore: Add various icons for Metals provided extensions and web views

tgodzik:
image

I realised while investigating how to add extension support for a single file that we can actually provide much better range of icons.

Java/Scala icons are the default vs code ones. The doctor and the stack trace icons are downloaded, but according to the website they are free to use for commercial/personal use.


src/doctor.ts
```diff
@@ -39,6 +48,10 @@ export class DoctorProvider implements Disposable {
         ViewColumn.Active,
         { enableCommandUris: true }
       );
+      if (this.extensionContext)
+        this.doctor.iconPath = Uri.file(
+          path.join(this.extensionContext.extensionPath, "icons", "doctor.svg")
```
tgodzik, line 53: I was thinking of having different doctors depending on the theme, but that seemed to be too much on the nose :sweat_smile:

package.json
```diff
@@ -82,13 +94,34 @@
...
+        }
       },
       {
         "id": "buildtarget",
         "extensions": [
           ".metals-buildtarget"
-        ]
+        ],
+        "icon": {
+          "dark": "icons/book-dark.svg",
```
dos65, line 109: Is it a default one for vscode?
I don't see it in pr

dos65 (approved):
Nice idea!

src/doctor.ts
```diff
@@ -24,8 +31,10 @@ export class DoctorProvider implements Disposable {
    * We should take focus into account when doctor caching mechanism will be implemented on the server.
    */
   isOpened = false;
-
-  constructor(private client: LanguageClient) {}
+  private extensionContext: ExtensionContext | undefined;
+  constructor(private client: LanguageClient, context: ExtensionContext) {
+    this.extensionContext = context;
+  }
```
kpodsiad, line 37: Why union type with undefined? Also, if you add modifier to the constructor parameter, it'll be converted to class field automatically. 
so
```typescript
constructor(private client: LanguageClient) {}
```
is more or less equal to
```typescript
  client: LanguageClient
  constructor(client: LanguageClient) {
    this.client = client;
  }
```

```suggestion
  doctorIconPath: Uri;

  constructor(
    private client: LanguageClient,
    extensionContext: ExtensionContext
  ) {
    this.doctorIconPath = Uri.file(
      path.join(extensionContext.extensionPath, "icons", "doctor.svg")
    );
  }
```

src/doctor.ts
```diff
@@ -39,6 +48,11 @@ export class DoctorProvider implements Disposable {
         ViewColumn.Active,
         { enableCommandUris: true }
       );
+      if (this.extensionContext) {
+        this.doctor.iconPath = Uri.file(
+          path.join(this.extensionContext.extensionPath, "icons", "doctor.svg")
```
kpodsiad, line 53: And then just `this.doctor.iconPath = this.doctorIconPath;`

kpodsiad (approved):

package.json
tgodzik: It was actually already added previously, I reused an existing icon.

src/doctor.ts
tgodzik: > Why union type with undefined? Also, if you add modifier to the constructor parameter, it'll be converted to class field automatically.

it's just because my typescript knowledge is super limited :sweat_smile:

tgodzik: Fixed, thanks!

Issue #1144, opened 2022-09-01
Add more unit tests


tanishiking:
**Describe the problem**

Even though `metals-vscode` has many advanced features (such as checking for updates, installing java, showing release notes, etc.), there are quite a small number of unit/integration tests.

Therefore, when we update the features of `metals-vscode`, we usually "test" the extension by hand. However, testing by hand leaves a lot of anxiety for release. (What if some features are broken without notice?)

I propose adding some unit tests to the extension to remove the anxieties. (It would be best if we could test all the features with integration tests, but it requires too much, let's start with unit testing).

**Which features to unit test?**
In unit tests, we should focus on the logic that can be independent of vscode APIs.

- [ ] check for updates on the Metals server
  - [ ] `needCheckForUpdates`
  - [ ] `validateCurrentVersion`
- [ ] `releaseNoteProvider`
- [ ] `installJavaAction`

I guess other features can be tested by integration tests.

**For writing unit tests**
As described in https://github.com/microsoft/vscode/issues/82471 

> The vscode-test package allows you to run integration tests where you launch the tests through vscode and a instance of vscode opens up. But if I want to do normal unit tests against code that has even a single reference to vscode I am not able to do so due to https://github.com/microsoft/vscode/issues/6586.

to run unit tests, we should extract the logic that doesn't depend on vscode API by abstracting out the dependencies.

In order to do this, I would propose adding some layers to the project structures.

- `application layer`: this layer is basically referred from `extension.ts`
  - can access vscode APIs (executing command, setting up `repository`s for service layer using VSCode APIs.
- `service layer`: this layer contains the **business logic that doesn't rely on vscode APIs**, we will unit test against this layer.
- `repository layer`: this layer contains the logic that
  - access to vscode APIs (such as reading and updating configurations)
  - access to external APIs (HTTP)

```
src
|-- installJavaAction.ts (application layer)
|-- service
|   `-- CheckForUpdateService.ts
|-- repository
|   `-- CheckForUpdateRepo.ts
```

Here's a POC implementation for unit testing `needCheckForUpdate` https://github.com/tanishiking/metals-vscode/pull/82

**Additional context**

**Search terms**

PR #1145, opened 2022-09-02
test: Add test for "needCheckForUpdates"

tanishiking:
https://github.com/scalameta/metals-vscode/issues/1144

This commit adds a test for `needCheckForUpdates`.

For adding unit tests, this commit abstract away the logic that needs
access to vscode APIs, becuase we need to run an integration test for
accessing vscode APIs.
see: https://github.com/microsoft/vscode-wordcount/issues/5#issuecomment-171050759
https://github.com/microsoft/vscode/issues/82471#issuecomment-564731132

### DIFF
--- a/src/repository/CheckForUpdateRepo.ts
+++ b/src/repository/CheckForUpdateRepo.ts
@@ -1,0 +1,50 @@
+import { ExtensionContext, Memento } from "vscode";
+import { ConfigurationTarget } from "../ConfigurationTarget";
+
+type LastUpdated = {
+  prevVersion?: string;
+  lastUpdatedAt?: string;
+};
+
+export interface CheckForUpdateRepo {
+  getLastUpdated(target: ConfigurationTarget): LastUpdated;
+  saveLastUpdated(
+    serverVersion: string,
+    lastUpdatedAt: string,
+    target: ConfigurationTarget
+  ): void;
+}
+
+export class DefaultCheckForUpdateRepo implements CheckForUpdateRepo {
+  constructor(private context: ExtensionContext) {}
+
+  private CurrentVersionKey = "currentVersion";
+  private LastUpdatedAtKey = "lastUpdatedAt";
+
+  getLastUpdated(target: ConfigurationTarget): LastUpdated {
+    const state = this.storage(target);
+    const prevVersion = state.get<string>(this.CurrentVersionKey);
+    const lastUpdatedAt = state.get<string>(this.LastUpdatedAtKey);
+    return {
+      prevVersion,
+      lastUpdatedAt,
+    };
+  }
+
+  saveLastUpdated(
+    serverVersion: string,
+    lastUpdatedAt: string,
+    target: ConfigurationTarget
+  ): void {
+    const state = this.storage(target);
+    state.update(this.CurrentVersionKey, serverVersion);
+    state.update(this.LastUpdatedAtKey, lastUpdatedAt);
+    return;
+  }
+
+  private storage(target: ConfigurationTarget): Memento {
+    return target === ConfigurationTarget.Global
+      ? this.context.globalState
+      : this.context.workspaceState;
+  }
+}
--- a/src/test/unit/checkForUpdate.test.ts
+++ b/src/test/unit/checkForUpdate.test.ts
@@ -1,0 +1,89 @@
+import assert from "assert";
+import { ConfigurationTarget } from "../../ConfigurationTarget";
+import { needCheckForUpdates } from "../../service/checkForUpdate";
+import { CheckForUpdateRepo } from "../../repository/CheckForUpdateRepo";
+import sinon from "sinon";
+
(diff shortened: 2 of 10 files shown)


tanishiking:
Copied from https://github.com/tanishiking/metals-vscode/pull/82

> [kpodsiad](https://github.com/kpodsiad) commented [7 hours ago](https://github.com/tanishiking/metals-vscode/pull/82#issuecomment-1234763293)
> One problem with assert is that it doesn't provide meaningful error messages. I was thinking about using [earl](https://earljs.dev/docs/), wdyt?

Yeah, adding an assertion library is good 👍 
However, in that case, I would choose [chai](https://www.chaijs.com/) as its tried and tested solution (or I would pick `jest`  instead of `mocha + earljs` as all-in-one test framework, but I personally prefer `mocha + chai + sinon` for testing node project).

src/test/unit/checkForUpdate.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    const repo = new MockRepo(prevVersion, today);
+    const spy = sinon.spy(repo, "saveLastUpdated");
+    const actual = await needCheckForUpdates(
+      currentVersion,
+      today,
+      ConfigurationTarget.Global,
+      repo
+    );
+    expect(actual).false;
+    expect(spy.getCall(0).args).to.eql([
```
tanishiking, line 60: we can use `spy.calledWith(...)`, but this way will show better error message if test fail (while `calledWith` shows true or false)

kpodsiad:
> Copied from [tanishiking#82](https://github.com/tanishiking/metals-vscode/pull/82)
> 
> > [kpodsiad](https://github.com/kpodsiad) commented [7 hours ago](https://github.com/tanishiking/metals-vscode/pull/82#issuecomment-1234763293)
> > One problem with assert is that it doesn't provide meaningful error messages. I was thinking about using [earl](https://earljs.dev/docs/), wdyt?
> 
> Yeah, adding an assertion library is good 👍 However, in that case, I would choose [chai](https://www.chaijs.com/) as its tried and tested solution (or I would pick `jest` instead of `mocha + earljs` as all-in-one test framework, but I personally prefer `mocha + chai + sinon` for testing node project).

Actually I was wrong. `assert` is good enough, it has meaningful errors and it doesn't require additional dependency. I intentionally switched jest to mocha in this repository simple because it's [much smaller](https://bundlephobia.com/package/mocha@10.0.0) than [jest](https://bundlephobia.com/package/jest@29.0.1) and it doesn't bring so many dependencies (I recall some problems with them)

tanishiking:
> I intentionally switched jest to mocha in this repository simple because it's much smaller than jest and it doesn't bring so many dependencies

Yeah, I also think mocha chai sinon would be noice in this repository (than jest) :+1:
You don't want to add chai as a dependency? I prefer to use chai because I like BDD style testing and it's small library.

src/repository/CheckForUpdateRepo.ts
```diff
@@ -0,0 +1,50 @@
+import { ExtensionContext, Memento } from "vscode";
+import { ConfigurationTarget } from "../ConfigurationTarget";
+
+type LastUpdated = {
```
kpodsiad, line 4: I think it's more common to use interface for such data types

src/repository/CheckForUpdateRepo.ts
```diff
@@ -0,0 +1,50 @@
...
+export interface CheckForUpdateRepo {
+  getLastUpdated(target: ConfigurationTarget): LastUpdated;
+  saveLastUpdated(
+    serverVersion: string,
+    lastUpdatedAt: string,
+    target: ConfigurationTarget
+  ): void;
+}
+
+export class DefaultCheckForUpdateRepo implements CheckForUpdateRepo {
```
kpodsiad, line 18: SO it's not a problem that interface is located in the same file which has imports from `vscode` package as long as it doesn't use them, nice. Maybe I'll rewrite/write some unit tests for test explorer then :D

kpodsiad (approved):
> You don't want to add chai as a dependency? I prefer to use chai because I like BDD style testing and it's small library.

It's ok, I just meant that I forgot that node's assert also provides good errors ;) 

I think we're missing one important think (correct me if I' worng) - do we have some docs how to run single test file, single test from suite? That would be nice to have.
TBH, even `yarn test` might be useful for people like don't do js day to day.

src/repository/CheckForUpdateRepo.ts
tanishiking: I don't have a strong opinion, but I prefer to use `type` over `interface` for the data types that shouldn't be extended (by inheritance or declaration merging).

tanishiking:
Thanks! Added docs on how to run tests https://github.com/scalameta/metals-vscode/pull/1145/commits/b70d5b3101f37be5e1724974d436beb2a9af78cf
merging after CIs pass

PR #1184, opened 2022-10-12
improvement: Switch to marked to enable links in release notes

tgodzik:
Previously, if links were clicked that corresponded to header on the same site `#title-of-h2` it woukld not work due to missing auto generated id. Now, that works correctly with marked and this seems to be used by VS Code also though it appears to be inlined there:

https://github.com/microsoft/vscode/blob/4affcdb07c275abdf9944180e7c6e3e82ad3edb2/src/vs/base/common/marked/marked.d.ts

@kpodsiad was there any reason for using Remarkable?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1006,7 +1006,7 @@
     "@types/glob": "^8.0.0",
     "@types/mocha": "^10.0.0",
     "@types/node": "18.8.4",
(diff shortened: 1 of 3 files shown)


kpodsiad (approved):
I think it was the first npm package that I grabbed, no particular reason

PR #1208, opened 2022-10-31
remove any in scalaDebugger

kpodsiad:


src/scalaDebugger.ts
```diff
@@ -33,40 +59,37 @@ export function initialize(outputChannel: vscode.OutputChannel): Disposable[] {
...
+  return vscode.debug.startDebugging(undefined, configuration);
 }
 
 class ScalaMainConfigProvider implements vscode.DebugConfigurationProvider {
-  resolveDebugConfiguration(
+  async resolveDebugConfiguration(
     _folder: WorkspaceFolder | undefined,
     debugConfiguration: DebugConfiguration
-  ): ProviderResult<DebugConfiguration> {
+  ): Promise<DebugConfiguration | null> {
```
tgodzik, line 92: Wasn't `ProviderResult` already correct?

tgodzik (approved):
Looks good, just one question from my side.

src/scalaDebugger.ts
kpodsiad: `async` function has to return `Promise<T>`, `ProviderResult` is a bit weaker than Promise as it's `null | Thenable<T>`

PR #1222, opened 2022-11-09
bugfix: Default to cmd.exe on Windows

tgodzik:
Previously, we would use any default shell that the user had set up, but that wouldn't work properly with powershell (and I wasn't able to make it work with any combinations.) Now, since it really doesn't matter where the command is run, we use cmd.exe, which works good enough.

### DIFF
--- a/src/debugger/scalaDebugger.ts
+++ b/src/debugger/scalaDebugger.ts
@@ -17,6 +17,7 @@ import {
   ServerCommands,
 } from "metals-languageclient";
 import { ExtendedScalaRunMain, ScalaCodeLensesParams } from "./types";
+import { platform } from "os";
 
 const configurationType = "scala";
 
@@ -49,6 +50,16 @@ function isExtendedScalaRunMain(
   );
 }
 
+function shellOptions(
+  env: Record<string, string>
+): vscode.ShellExecutionOptions {
+  if (platform() == "win32") {
+    return { executable: "cmd.exe", shellArgs: ["/c"], env };
+  } else {
+    return { env };
+  }
+}
+
 async function runMain(main: ExtendedScalaRunMain): Promise<boolean> {
   if (workspace.workspaceFolders) {
     const env = main.data.environmentVariables.reduce<Record<string, string>>(
@@ -64,7 +75,7 @@ async function runMain(main: ExtendedScalaRunMain): Promise<boolean> {
       workspace.workspaceFolders[0],
       "Scala run",
       "Metals",
-      new ShellExecution(main.data.shellCommand, { env })
+      new ShellExecution(main.data.shellCommand, shellOptions(env))
     );
 
     await tasks.executeTask(task);
--- a/src/debugger/scalaDebugger.ts
+++ b/src/debugger/scalaDebugger.ts
@@ -50,19 +50,25 @@ function isExtendedScalaRunMain(
   );
 }
 
-function shellOptions(
-  env: Record<string, string>
-): vscode.ShellExecutionOptions {
+/**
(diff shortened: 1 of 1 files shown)


src/debugger/scalaDebugger.ts
```diff
@@ -49,6 +50,14 @@ function isExtendedScalaRunMain(
   );
 }
 
+function shellOptions(env: Record<string, string>): vscode.ShellExecutionOptions {
+  if (platform() == "win32") {
+    return { executable: "cmd.exe", shellArgs: ["/c"], env }
```
kpodsiad, line 55: does `powershell.exe` works here?

tgodzik: powershell doesn't seem to work at all when run as a task and it needs some additional escaping when run manually. And I am also not sure how the executable is called. Anyway, there is no reason to run powershell as we would not be using any shell specific things at all.

kpodsiad: `powershell.exe` is the thing which I found in the internet when searching for vs code task api examples, but cmd is fine 👍

src/debugger/scalaDebugger.ts
```diff
@@ -49,6 +50,16 @@ function isExtendedScalaRunMain(
...
 
+function shellOptions(
+  env: Record<string, string>
+): vscode.ShellExecutionOptions {
+  if (platform() == "win32") {
+    return { executable: "cmd.exe", shellArgs: ["/c"], env };
+  } else {
+    return { env };
+  }
+}
```
kpodsiad, line 61:
```suggestion
/**
 * Return platform specific options for task.
 *
 * By default, task will use whatever shell user has defined as a default one.
 * However, for Windows tasks seems to not work properly with Powershell,
 * that's why we want to explicitly specify executable as plain old cmd
 */
function platformSpecificOptions(): vscode.ShellExecutionOptions {
  if (platform() == "win32") {
    return { executable: "cmd.exe", shellArgs: ["/c"] };
  } else {
    return {};
  }
}
```

src/debugger/scalaDebugger.ts
```diff
@@ -64,7 +75,7 @@ async function runMain(main: ExtendedScalaRunMain): Promise<boolean> {
       workspace.workspaceFolders[0],
       "Scala run",
       "Metals",
-      new ShellExecution(main.data.shellCommand, { env })
+      new ShellExecution(main.data.shellCommand, shellOptions(env))
```
kpodsiad, line 78: some sugar with `const { environmentVariables, shellCommand } = main.data;` ;)
```ts
  const { environmentVariables, shellCommand } = main.data;
  if (workspace.workspaceFolders) {
    const env = environmentVariables.reduce<Record<string, string>>(
      (acc, envKeyValue) => {
        const [key, value] = envKeyValue.split("=");
        return { ...acc, [key]: value };
      },
      {}
    );

    const shellOptions = { ...platformSpecificOptions(), env };

    const task = new Task(
      { type: "scala", task: "run" },
      workspace.workspaceFolders[0],
      "Scala run",
      "Metals",
      new ShellExecution(shellCommand, shellOptions)
    );
```

kpodsiad (changes requested):

src/debugger/scalaDebugger.ts
tgodzik: Done!

src/debugger/scalaDebugger.ts
tgodzik: this is some crazy magic, changed!

kpodsiad (approved):
👍

PR #1245, opened 2022-12-02
bugfix: Check if tests namespace is enabled

tgodzik:
Eclipse Theia does not support test explorer currently.

Related to https://github.com/scalameta/metals-vscode/discussions/1244

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,6 +28,7 @@ import {
   ProviderResult,
   Hover,
   TextDocument,
+  tests,
 } from "vscode";
 import {
   LanguageClient,
@@ -600,7 +601,7 @@ function launchMetals(
           "Test Explorer"
         );
 
-      const istTestManagerDisabled = getTestUI() === "Code Lenses";
+      const istTestManagerDisabled = !tests || getTestUI() === "Code Lenses";
       const testManager = createTestManager(client, istTestManagerDisabled);
 
       const disableTestExplorer = workspace.onDidChangeConfiguration(() => {


src/extension.ts
```diff
@@ -600,7 +601,7 @@ function launchMetals(
           "Test Explorer"
         );
 
-      const istTestManagerDisabled = getTestUI() === "Code Lenses";
+      const istTestManagerDisabled = !tests || getTestUI() === "Code Lenses";
```
kpodsiad, line 604: I don't fully get it, how is it supposed to help with issue?

tgodzik: Seems that the namespace is undefined when the code is used in Theia

kpodsiad: ok, now I see it

kpodsiad, line 604:
```suggestion
      // vscodeTextExplorer can be undefined e.g. for eclipse theia, 
      // see https://github.com/scalameta/metals-vscode/discussions/1244
      const istTestManagerDisabled =
        !vscodeTextExplorer || getTestUI() === "Code Lenses";
```

src/extension.ts
```diff
@@ -28,6 +28,7 @@ import {
   ProviderResult,
   Hover,
   TextDocument,
+  tests,
```
kpodsiad, line 31:
```suggestion
  tests as vscodeTextExplorer,
```

kpodsiad (approved):
nitpick, it's worth to write what this is important

PR #1278, opened 2022-12-27
fix: workaround for empty completion list + `isIncomplete = true`

dos65:
If VScode receives empty completion list it ignores `isIncomplete` field. 
Fixes: https://github.com/scalameta/metals/issues/4756

VScode issue: https://github.com/microsoft/vscode/issues/155738


tgodzik (approved):
LGTM!

PR #1285, opened 2023-01-06
feature: Add option to enable semantic highlighting

tgodzik:
And also bump server to the version that supports it


kpodsiad:
approve is 2 clicks, merge is one 😂

PR #1298, opened 2023-01-13
feature: When using run discovery use shell command to run in task terminal

tgodzik:
Previously, only when run from code lenses we would use the task terminal for running. Now, also when running the discovery commands (including when using F5 with no configurations) we will by default run within that terminal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -220,7 +220,7 @@
       "properties": {
         "metals.serverVersion": {
           "type": "string",
-          "default": "0.11.10+31-181ba661-SNAPSHOT",
+          "default": "0.11.10+138-254db453-SNAPSHOT",
           "markdownDescription": "The version of the Metals server artifact. Requires reloading the window.  \n\n**VS Code extension version is guaranteed to work only with the default version, change if you know what you're doing**"
         },
         "metals.serverProperties": {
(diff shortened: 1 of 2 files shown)


src/debugger/RunDebugAdapter.ts
```diff
@@ -0,0 +1,121 @@
...
+    this.request_seq = request.seq;
+    this.command = request.command;
+    if (message) {
+      this.success = false;
+      (<any>this).message = message;
+    } else {
+      this.success = true;
+    }
+  }
+}
```
tgodzik, line 70: Everything until here is copied over from https://github.com/microsoft/vscode-debugadapter-node/blob/f33a09029b2d1212f07ba087042a315e8a958c68/adapter/src/messages.ts but I have no idea if we can import it :thinking:

src/debugger/RunDebugAdapter.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export class RunDebugAdapter implements vscode.DebugAdapter {
```
tgodzik, line 72: So this is a fake server so that we can run in terminal from config. Unfortunately, we need to do it via DAP, so the easiest way seemed just faking it. One minor issue is that the icons for debugging show up for a second.

PR #1346, opened 2023-03-01
improvement: Add 'Open issue' to command palette

tgodzik:


src/extension.ts
```diff
@@ -490,6 +490,7 @@ function launchMetals(
         ServerCommands.AmmoniteStop,
         ServerCommands.ScalaCliStart,
         ServerCommands.ScalaCliStop,
+        "open-new-github-issue",
```
tgodzik, line 493: languageclient library is not really used anymore by anything aside from the Metals extension, so probably it doesn't make much sense to add it there.

We should instead migrate most of the code here, but I would like to include this command before the release.

kpodsiad: We can embed that library in metals-vscode but in separate dir, so in the future it might be extracted easily, if needed. But having it locally would make things easier 👍

kpodsiad (approved):

PR #1381, opened 2023-03-21
Integrate metals-languageclient with the build

gabro:
This is a followup of #1373.

Not that I've already force-pushed `main` to import `metals-languageclient` (along with its git history) as a subdir under `packages`


gabro:
Self-merging to fix main

PR #1385, opened 2023-04-11
feature: Run local coursier if available to avoid using bootstraped one

tgodzik:
This might help out in corporate environements since it will require installing coursier only once and Metals will be able to use it.

In later steps I will try to automatically download coursier or native image via metals if not available.

Continuation of https://github.com/scalameta/metals-languageclient/pull/524

### DIFF
--- a/packages/metals-languageclient/src/fetchMetals.ts
+++ b/packages/metals-languageclient/src/fetchMetals.ts
@@ -1,6 +1,9 @@
 import * as semver from "semver";
+import path from "path";
+import fs from "fs";
 import { ChildProcessPromise, spawn } from "promisify-child-process";
 import { JavaConfig } from "./getJavaConfig";
+import { OutputChannel } from "./interfaces/OutputChannel";
 
 interface FetchMetalsOptions {
   serverVersion: string;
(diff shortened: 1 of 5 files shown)


packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -1,56 +1,114 @@
...
+      }
+    })
+    .find(
+      (p) =>
+        p.endsWith(path.sep + "cs") ||
+        p.endsWith(path.sep + "coursier") ||
+        (process.platform === "win32" && p.endsWith(path.sep + "cs.bat"))
+    );
+  if (possibleCoursier) {
+    const coursierVersion = spawnSync(possibleCoursier, ["version"]);
```
tgodzik, line 100: Not sure how to work around this, I can't find a way to chain ChildProcess since then is only available on Promise. Alternatively, we could make a spawn with a wait to make sure this doesn't take too long. Ideas?

tgodzik: I think I did? Doing `.then` on ChildProcessPromise was only returning `Promise<Output>` and not sure of there is a way around it really.

packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -1,56 +1,111 @@
...
+    output.appendLine(`Using coursier located at ${coursier}`);
+    return spawn(coursier, coursierArgs);
+  } else {
+    return spawnDefault();
+  }
+}
+
+export function validateCoursier(pathEnv: string): string | undefined {
+  const isWindows = process.platform === "win32";
+  const possibleCoursier = pathEnv
```
tanishiking, line 83: Maybe we want to use https://github.com/otiai10/lookpath to find commands from PATH, this library deals well with the platform differences (and it takes care of `PATHEXT` to checking `cs.bat` and `cs.exe`)

packages/metals-languageclient/src/fetchMetals.ts
tanishiking: Maybe https://github.com/sindresorhus/execa help?
As long as we use raw child_process stuffs (without writing a wrapper) there's no way to do that (get stdout wrapped by Promise if success, and Promise failed if return code isn't 0).

tanishiking (commented):
I left a few comments, but overall it looks good :+1:

packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -1,56 +1,111 @@
...
+    const coursier: string = possibleCoursier;
+    output.appendLine(`Using coursier located at ${coursier}`);
+    return spawn(coursier, coursierArgs);
+  } else {
+    return spawnDefault();
+  }
+}
+
+export function validateCoursier(pathEnv: string): string | undefined {
+  const isWindows = process.platform === "win32";
```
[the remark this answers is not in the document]
tgodzik: yep :P I checked the docs

packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -1,56 +1,111 @@
...
+    "-r",
+    "bintray:scalacenter/releases",
+    "-r",
+    "sonatype:public",
+    "-r",
+    "sonatype:snapshots",
+    "-p",
+  ];
+
+  const path = process.env["PATH"];
```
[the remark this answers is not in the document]
tgodzik: `PATH` works, the CI actually checks this correctly. https://superuser.com/questions/441611/what-is-the-linux-equivalent-of-windows-path

packages/metals-languageclient/src/fetchMetals.ts
tgodzik: Doesn't look like it :/ We want to pipe the output from coursier to VS Code output, but it seems I can do it only via ChildProcessPromise. I tried returning a Promise<ChildProcessPromise>, but that seems to have a wrong type in the actual javascript underneath:
https://github.com/scalameta/metals-vscode/commit/906323d33891b9318e95266618627b4445e85cdd

tgodzik: Och actually if I have an intermediate object it works, not 100% pretty, but feels ok.

tanishiking (approved):
LGTM:+1:

PR #1394, opened 2023-05-10
bugfix: display tests in explorer per workspace folder

kasiaMarek:
Previously: test suites were displayed per build target.
Now: we group them also per workspace folder

connected to: https://github.com/scalameta/metals/pull/5212
resolves: https://github.com/scalameta/metals/issues/5189

### DIFF
--- a/packages/metals-vscode/src/testExplorer/addTestSuites.ts
+++ b/packages/metals-vscode/src/testExplorer/addTestSuites.ts
@@ -1,10 +1,11 @@
 import * as vscode from "vscode";
 import { TargetUri } from "../types";
-import { AddTestSuite, TargetName } from "./types";
+import { AddTestSuite, TargetName, FolderName, FolderUri } from "./types";
 import {
   prefixesOf,
-  refineTestItem,
   TestItemPath,
+  refineRunnableTestItem,
+  refineTestItem,
   toVscodeRange,
 } from "./util";
 
@@ -21,12 +22,20 @@ export function addTestSuite(
   testController: vscode.TestController,
   targetName: TargetName,
   targetUri: TargetUri,
+  folderName: FolderName,
+  folderUri: FolderUri,
   event: AddTestSuite
 ): void {
+  const workspaceFolderItem = getOrCreateWorkspaceFolderItem(
+    testController,
+    folderName,
+    folderUri
+  );
+
   const buildTargetItem = getOrCreateBuildTargetItem(
     testController,
-    targetName,
-    targetUri
+    workspaceFolderItem,
+    targetName
   );
 
   function addTestSuiteLoop(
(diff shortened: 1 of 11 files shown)


packages/metals-vscode/src/testExplorer/types.ts
```diff
@@ -135,34 +139,50 @@ export interface TestSuiteRun {
   testCases: vscode.TestItem[];
 }
 
-export type MetalsTestItemKind = "project" | "package" | "suite" | "testcase";
+export type RunnableMetalsTestItemKind = "suite" | "testcase";
+
+export type MetalsTestItemKind =
+  | "workspaceFolder"
+  | "project"
```
kpodsiad, line 146: I wonder if `module` wouldn't be a better name instead of project

packages/metals-vscode/src/testExplorer/testRunHandler.ts
```diff
@@ -73,7 +73,11 @@ export async function runHandler(
     for (const test of tests) {
       if (!excludes.has(test)) {
         const kind = test._metalsKind;
-        if (kind === "project" || kind === "package") {
+        if (
+          kind === "project" ||
+          kind === "package" ||
+          kind === "workspaceFolder"
+        ) {
```
kpodsiad, line 80: I should have write some documentation about this :( Consider adding my loud thoughts as comments/proper doc somewhere.
there is no need to change more bits here because `createRunQueue` call itself recursively, so almost nothing changed in terms of running - still kind `suite` is the operating atom here and in result analyzer.

packages/metals-vscode/src/testExplorer/addTestSuites.ts
```diff
@@ -72,16 +81,37 @@ export function addTestSuite(
...
+  workspaceFolderItem: vscode.TestItem,
+  targetName: TargetName
 ): vscode.TestItem {
-  const buildTarget = testController.items.get(targetName);
+  const buildTarget = workspaceFolderItem.children.get(targetName);
   if (buildTarget) {
     return buildTarget;
   }
 
   const createdNode = testController.createTestItem(targetName, targetName);
```
kpodsiad, line 92: Should prefix target's test item id with workspace? Or it isn't necessary because id has to be unique only in a given subtree and if we're child of workspace then we're already guaranteed to be unique?

kpodsiad (approved):
I haven't run this branch myself, but changes looks reasonable. LGTM

packages/metals-vscode/src/testExplorer/addTestSuites.ts
kasiaMarek:
```
@param id Identifier for the TestItem. The test item's ID must be in the {@link TestItemCollection} it's added to
where
TestItemCollection is a collection of test items, found in TestItem.children and TestController.items.
```
So it only has to be unique among its siblings and we always access it by first accessing the parent (workspace folder).

packages/metals-vscode/src/testExplorer/types.ts
kasiaMarek: Agreed, "project" is confusing in this context. If anything a workspace folder should be named "project".

PR #1395, opened 2023-05-17
Add options for import java.compiler module

VankaTaganai:
This changes related to [Java Hover](https://github.com/scalameta/metals/pull/5024) for Metals. It's required to export `java.compiler` module manually for Java version greater than 17. I've handled version of Java like it's made in [Metals](https://github.com/scalameta/metals/blob/45df705da236edfddcea7e358592fd3202bb9460/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala#L144).

### DIFF
--- a/packages/metals-languageclient/src/getJavaConfig.ts
+++ b/packages/metals-languageclient/src/getJavaConfig.ts
@@ -3,6 +3,7 @@ import * as path from "path";
 
 export interface JavaConfig {
   javaOptions: string[];
+  javaHome: string;
   javaPath: string;
   coursierPath: string;
   coursierMirrorFilePath: string | undefined;
@@ -37,6 +38,7 @@ export function getJavaConfig({
 
   return {
     javaOptions,
+    javaHome,
     javaPath,
     coursierPath,
     coursierMirrorFilePath,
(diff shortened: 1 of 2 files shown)


dos65 (approved):
Works well! 

I just submitted a minor fix onto your branch - I realized that we actually have an actual javaHome and there is no need to extract it back from env.

dos65:
@ckipp01 @ayoub-benali - Hey! Could you say if you will be able to port changeset from this PR for nvim and sublime?
They are needed for better java support in case if jvm is greater or eq of 17.

ckipp01:
> @ckipp01 @ayoub-benali - Hey! Could you say if you will be able to port changeset from this PR for nvim and sublime? They are needed for better java support in case if jvm is greater or eq of 17.

Thanks for the ping @dos65! I haven't been following the work on the Java hover pr, but how come this is actually necessary? I don't mean why are the extra flags necessary, but how come they need to be used to start Metals? I was under the assumption that when Metals started the java compiler or whatever was being used it could pass in those flags so the client doesn't need to worry about them. Is that at all possible, or is absolutely necessary that Metals is _started_ with those flags?

To answer your original question, I _think_ this should be portable, I just need to rewrite it in lua for nvim-metals.

ayoub-benali:
Hey @dos65, from my understanding a set of flags are passed to the client if java_home is 17 or above. It shouldn't be a problem to do that for sublime

PR #1396, opened 2023-06-16
bugfix: Run task in the correct directory

tgodzik:
Also fixed two places where we used the first folder, though this shouldn't be an issue.

### DIFF
--- a/packages/metals-vscode/src/util.ts
+++ b/packages/metals-vscode/src/util.ts
@@ -1,10 +1,12 @@
 import * as path from "path";
 import os from "os";
 import {
-  workspace,
   TextEditor,
   WorkspaceConfiguration,
   ConfigurationTarget,
+  window,
+  workspace,
+  WorkspaceFolder,
 } from "vscode";
 import {
   ExecuteCommandRequest,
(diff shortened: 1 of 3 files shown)


packages/metals-vscode/src/util.ts
```diff
@@ -48,14 +50,29 @@ export function getConfigValue<A>(
...
+  if (activeEditor) {
+    const workspaceFolder = workspace.getWorkspaceFolder(
+      activeEditor.document.uri
+    );
+    if (workspaceFolder) {
+      workspaceIndex = workspaceFolder.index;
+    }
+  }
+  return workspace.workspaceFolders?.[workspaceIndex];
+}
```
gabro, line 74: Possibly a bit easier to read?

```suggestion
	const activeEditor = window.activeTextEditor;
  if (activeEditor) {
  	return workspace.getWorkspaceFolder(activeEditor.document.uri)
	}
	return workspace.workspaceFolders?.[0];
}
```

There's a minor difference in logic but I don't know if it's practically a difference: if `getWorkspaceFolder` returns `undefined` we return `undefined` instead of `workspaceFolders?.[0]`.

gabro (approved):
One suggestion, but it's already 👍

packages/metals-vscode/src/util.ts
kpodsiad: ++ for @gabro snippet. My 2c:
```suggestion
  const activeEditorUri = window.activeTextEditor?.document.uri;
  if (activeEditorUri) {
    return workspace.getWorkspaceFolder(activeEditorUri);
  }
  return workspace.workspaceFolders?.[0];
}
```

kpodsiad (approved):

packages/metals-vscode/src/util.ts
tgodzik: Thanks! Fixed!

PR #1401, opened 2023-07-11
Handle `usernotification` event from debug session

adpi2:
Copied from [microsoft/vscode-java-debug](https://github.com/microsoft/vscode-java-debug)

The scala-debug-adapter can send `usernotification` events that should be shown to the user. For instance, it can contain the compilation errors of a conditional breakpoint.

Tested locally:
usernotification

Fixes https://github.com/scalacenter/scala-debug-adapter/issues/455

### DIFF
--- a/packages/metals-vscode/src/extension.ts
+++ b/packages/metals-vscode/src/extension.ts
@@ -1310,11 +1310,10 @@ function toggleBooleanWorkspaceSetting(setting: string) {
 function registerDebugEventListener(context: ExtensionContext) {
   context.subscriptions.push(
     debug.onDidReceiveDebugSessionCustomEvent((customEvent) => {
-      const t = customEvent.session ? customEvent.session.type : undefined;
-      if (t !== SCALA_LANGID) {
-        return;
-      }
-      if (customEvent.event === USER_NOTIFICATION_EVENT) {
+      if (
+        customEvent.session.type !== SCALA_LANGID &&
+        customEvent.event === USER_NOTIFICATION_EVENT
+      ) {
         handleUserNotification(customEvent);
       }
     })
--- a/packages/metals-vscode/src/extension.ts
+++ b/packages/metals-vscode/src/extension.ts
@@ -29,6 +29,8 @@ import {
   Hover,
   TextDocument,
   tests as vscodeTextExplorer,
+  debug,
+  DebugSessionCustomEvent,
 } from "vscode";
 import {
   LanguageClient,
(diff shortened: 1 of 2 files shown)


packages/metals-vscode/src/extension.ts
```diff
@@ -1299,3 +1306,27 @@ function toggleBooleanWorkspaceSetting(setting: string) {
...
+function registerDebugEventListener(context: ExtensionContext) {
+  context.subscriptions.push(
+    debug.onDidReceiveDebugSessionCustomEvent((customEvent) => {
+      const t = customEvent.session ? customEvent.session.type : undefined;
+      if (t !== SCALA_LANGID) {
+        return;
+      }
+      if (customEvent.event === USER_NOTIFICATION_EVENT) {
+        handleUserNotification(customEvent);
+      }
```
kpodsiad, line 1319: can `session` be null/undefined? Type suggests it can't. Anyway, I think following looks a bit clearer:
```suggestion
      const t = customEvent.session?.type;
      if (t === SCALA_LANGID && customEvent.event === USER_NOTIFICATION_EVENT) {
        handleUserNotification(customEvent);
      }
```

ckipp01 (commented):
Ah nice, this is something I'll need to add to nvim-metals. I don't really use VS Code, so I'll leave the review for the others.

kpodsiad (approved):

packages/metals-vscode/src/extension.ts
```diff
@@ -1299,3 +1306,26 @@ function toggleBooleanWorkspaceSetting(setting: string) {
   const currentValues = configProperty?.workspaceValue ?? false;
   config.update(setting, !currentValues, ConfigurationTarget.Workspace);
 }
+
+function registerDebugEventListener(context: ExtensionContext) {
+  context.subscriptions.push(
+    debug.onDidReceiveDebugSessionCustomEvent((customEvent) => {
+      if (
+        customEvent.session.type !== SCALA_LANGID &&
```
tgodzik, line 1314: Why `!==` ? Shouldn't it be equal to "scala" ? Or does the debug adapter send in different type than "scala" ?

tgodzik (approved):
just one question, otherwise LGTM

packages/metals-vscode/src/extension.ts
adpi2: It's a mistake, thanks for catching it :)

PR #1411, opened 2023-07-26
bugfix: accommodate junit encoded test names

kasiaMarek:
use display name for test cases

connected to:
https://github.com/scalameta/metals/pull/5485
https://github.com/scalameta/metals/issues/5378

### DIFF
--- a/packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
+++ b/packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
@@ -37,7 +37,7 @@ export function analyzeTestRun(
     if (result != null) {
       // if suite run contains test cases (run was started for (single) test case)
       if (kind === "testcase") {
-        analyzeTestCases(run, result, [test], test.parent);
+        analyzeTestCases(run, result, [test], test._metalsParent);
       }
       // if test suite has children (test cases) do a more fine-grained analyze of results.
       // run was started for whole suite which has children (e.g. junit one)
@@ -94,8 +94,9 @@ function analyzeTestCases(
 
   if (parent) {
     const tests = new Set(testCases.map((t) => t.id as TestName));
-    const failed = Array.from(testCasesResults.values())
-      .filter((result) => !tests.has(result.testName))
+    const failed = Array.from(testCasesResults.entries())
+      .filter(([name, _]) => !tests.has(name))
+      .map(([_, result]) => result)
       .filter(isFailed);
     if (failed.length > 0) {
       const msg = extractErrorMessages(failed);
@@ -117,10 +118,10 @@ function createTestCasesMap(
 ): Map<TestName, SingleTestResult> {
   const tuples: [TestName, SingleTestResult][] = testSuiteResult.tests.map(
     (test) => {
-      const name = test.testName.startsWith(parentName)
-        ? test.testName
-        : `${parentName}.${test.testName}`;
-
+      const name =
+        parentName && test.testName.startsWith(parentName)
+          ? test.testName.slice(parentName.length + 1)
+          : test.testName;
       return [name as TestName, test];
     }
   );
--- a/packages/metals-vscode/src/testExplorer/testRunHandler.ts
+++ b/packages/metals-vscode/src/testExplorer/testRunHandler.ts
@@ -110,9 +110,11 @@ export async function runHandler(
         tests: [],
       };
     } else {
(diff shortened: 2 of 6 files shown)


packages/metals-vscode/src/testExplorer/testRunHandler.ts
```diff
@@ -110,9 +110,11 @@ export async function runHandler(
         tests: [],
       };
     } else {
+      const { id, parent, _metalsParent } = test;
+      const idStartsWithParent = parent && id.startsWith(`${parent.id}.`);
       return {
-        className: test._metalsParent.id,
-        tests: [test.label],
+        className: _metalsParent.id,
+        tests: [idStartsWithParent ? id.slice(parent.id.length + 1) : id],
```
tgodzik, line 117: When do we encode the test with name of the parent? Could we fix that on the server side?

tgodzik: Do we still need the change here?

kasiaMarek: Yeah, I asked you again for review kind of by accident (I revoked  the request but I guess it already alarmed you). I forgot for a sec I was supposed to fixed that. :D

jkciesluk (approved):

packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
```diff
@@ -37,7 +37,7 @@ export function analyzeTestRun(
     if (result != null) {
       // if suite run contains test cases (run was started for (single) test case)
       if (kind === "testcase") {
-        analyzeTestCases(run, result, [test], test.parent);
+        analyzeTestCases(run, result, [test], test._metalsParent);
```
tgodzik, line 40: Is this change needed? The type hasn't changes, so this seems unnecessary

packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
```diff
@@ -94,8 +94,9 @@ function analyzeTestCases(
 
   if (parent) {
     const tests = new Set(testCases.map((t) => t.id as TestName));
-    const failed = Array.from(testCasesResults.values())
-      .filter((result) => !tests.has(result.testName))
+    const failed = Array.from(testCasesResults.entries())
```
tgodzik, line 97: Could explain a bit about why these chanegs are needed? Also why do we need to drop the parent in the test case id? This doesn't seem related to the display name

tgodzik (changes requested):

packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
kasiaMarek: Oh, the changes in tests are needed per se. We can go either way: test cases can have the parent's name encoded in the id or not. You suggested that maybe it makes more sense not to do it: https://github.com/scalameta/metals-vscode/pull/1411#discussion_r1284598727, so I changed that. One way or another for the names that come from the dap session we need to do the mapping, since depending on the build server the names might have the parent's id or not. Here the names returned by DAP were used (without mapping), which I believe to be an error anyway.

I'm concerned with the id because previously label (which now corresponds to the display name) was used when a test case was run.

packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
kasiaMarek: Yeah, but this did not work. The tests did not cover this, since the `id` and DAP results were both encoded in the same way (both had parent name in id) in tests, which according to a comment in the code doesn't have to be true in practice. But when I changed the encoding this part started to matter.

packages/metals-vscode/src/testExplorer/testRunHandler.ts
```diff
@@ -110,9 +110,10 @@ export async function runHandler(
         tests: [],
       };
     } else {
+      const { id, _metalsParent } = test;
       return {
-        className: test._metalsParent.id,
-        tests: [test.label],
+        className: _metalsParent.id,
+        tests: [id],
```
tgodzik, line 116:
```suggestion
        className: test._metalsParent.id,
        tests: [test.id],
```
that would be a bit less magical.

tgodzik (approved):
Sorry for taking so long! One minor comment, otherwise LGTM!

PR #1412, opened 2023-07-31
improvement: Set toggle for showInferredType to switch between true,false and minimal

tgodzik:
Previously, showInferredType could only be true or false. Now it is a string value that can be also of value "minimal". Because of that I switched the setting to string enum, added an automigration part and switch toggle to go through all the options

Follow up from https://github.com/scalameta/metals/pull/5284#issuecomment-1634836066

### DIFF
--- a/packages/metals-vscode/package.json
+++ b/packages/metals-vscode/package.json
@@ -220,7 +220,7 @@
       "properties": {
         "metals.serverVersion": {
           "type": "string",
-          "default": "1.0.0",
+          "default": "1.0.0+20-46905735-SNAPSHOT",
           "markdownDescription": "The version of the Metals server artifact. Requires reloading the window.  \n\n**VS Code extension version is guaranteed to work only with the default version, change if you know what you're doing**"
         },
         "metals.serverProperties": {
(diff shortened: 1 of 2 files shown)


packages/metals-vscode/src/extension.ts
```diff
@@ -905,7 +916,23 @@ function launchMetals(
...
+          case "true":
+            newValue = "minimal";
+            break;
+          case "minimal":
+            newValue = "false";
+            break;
+          case "false":
+            newValue = "true";
+            break;
+        }
```
dos65, line 934: Honestly, I'm a bit lost with this logic. 
Why `false` turns into `true` and `minimal` into `false`? 😸 
 
Do I understand it correctly:
- before launching server we do boolean to string migration in `migrateOldSetting`
- then we register a commend and perform some additional migration  into enum?

tgodzik: > then we register a command and perform some additional migration into enum?

That's only for toggle. Previously it was only `true-> false` and `false -> true`, but now we have 3 possible options and it's still useful to have a quick option to switch things on and off. So I decided on `false -> true -> minimal -> false` order.

dos65: @tgodzik oh - got it, thanks for explanation!

dos65 (approved):

packages/metals-vscode/src/extension.ts
tanishiking: > false -> true -> minimal -> false

That makes sense 👍

tanishiking (approved):

packages/metals-vscode/package.json
```diff
@@ -250,8 +250,14 @@
...
         },
         "metals.showInferredType": {
-          "type": "boolean",
-          "markdownDescription": "When this option is enabled, for each method that have inferred types that are not explicitely specified they are displayed as additional decorations."
+          "type": "string",
+          "default": "false",
+          "enum": [
+            "false",
+            "true",
+            "minimal"
```
kpodsiad, line 258: aren't `false` and `true` a bit unclear if value is an enum? Maybe `off`, `minimal`, `all` would be better? Although I guess this will be slightly incompatible change on the server side. I'm not insisting, but I'd like to hear some opinions before merging this PR.

kpodsiad (changes requested):

packages/metals-vscode/package.json
tgodzik: Sure this would make sense, but we would need to change the configuration on the Metals side first and make sure we take into account the old values. This change is more in line with the current state.

kpodsiad: IMO it'll be better to set up proper enum on server side with fallback to old boolean behavior. I'm not even talking about changing names "false, "true", "minimal" but to have proper ADT on server side to avoid code like `userConfig().showInferredType.contains("true")` bur rather write `userConfig().showInferredType.contains(ShowInferredTypes.True)`

kpodsiad (approved):

PR #1414, opened 2023-08-04
bugfix: Don't run command without clicking and fix infinite recursion

tgodzik:
Previously, we would run a command without clicking on the status bar, which could be problematic if a message was shown multiple times. We would also invoke the RunDoctor command, which invokes itself. Now, RunDoctor invokes the correct part of the extension and status bar command is not invoked right away.

I am not sure why this never happened before, but might be a change in VS Code that prompted it.

### DIFF
--- a/packages/metals-vscode/src/extension.ts
+++ b/packages/metals-vscode/src/extension.ts
@@ -569,9 +569,9 @@ function launchMetals(
         commands.executeCommand(workbenchCommands.focusDiagnostics)
       );
 
-      registerCommand(ClientCommands.RunDoctor, () =>
-        commands.executeCommand(ClientCommands.RunDoctor)
-      );
+      registerCommand(ClientCommands.RunDoctor, async () => {
+        await doctorProvider.runDoctor();
+      });
 
       registerCommand(ClientCommands.ToggleLogs, () => {
         if (channelOpen) {
@@ -722,13 +722,7 @@ function launchMetals(
             item.tooltip = params.tooltip;
           }
           if (params.command) {
-            const command = params.command;
             item.command = params.command;
-            commands.getCommands().then((values) => {
-              if (values.includes(command)) {
-                commands.executeCommand(command);
-              }
-            });
           } else {
             item.command = undefined;
           }


packages/metals-vscode/src/extension.ts
```diff
@@ -722,13 +722,7 @@ function launchMetals(
             item.tooltip = params.tooltip;
           }
           if (params.command) {
-            const command = params.command;
             item.command = params.command;
-            commands.getCommands().then((values) => {
-              if (values.includes(command)) {
-                commands.executeCommand(command);
-              }
-            });
```
kasiaMarek, line 731: I think we use this callback, e.g. `DebugProvider.reportErrors` for `DebugErrorsPresent` in metals. (I haven't tested this)

tgodzik: So, this was supposed to only invoke the command if anyone clicks on the status, so I am not sure why this was done and why it only started to pop up.

tgodzik: But, I think we set it to status if we don't care to show it right away, otherwise we should invoke a command instead

tgodzik: scala-debug-adapter-1691161142589

This started to happen :D

kpodsiad (approved):
Looks good, but I'm also a bit intrested how things was working before :D

kasiaMarek (approved):

packages/metals-vscode/src/extension.ts
```diff
@@ -569,9 +569,9 @@ function launchMetals(
         commands.executeCommand(workbenchCommands.focusDiagnostics)
       );
 
-      registerCommand(ClientCommands.RunDoctor, () =>
-        commands.executeCommand(ClientCommands.RunDoctor)
-      );
+      registerCommand(ClientCommands.RunDoctor, async () => {
+        await doctorProvider.runDoctor();
+      });
```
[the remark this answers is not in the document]
tgodzik: It didn't work before, but it seems no once clicked the status bar :sweat_smile:

PR #1419, opened 2023-08-14
Add hot code reload

iusildra:
Address this [feature request](https://github.com/scalameta/metals-feature-requests/issues/336), but for now, it should only works with sbt, I'll look to make it works with bloop too later. I adapted the implementation from [vscode-java-debug](https://github.com/microsoft/vscode-java-debug) to work with metals

The implementation on the debugger side is in [this pr](https://github.com/scalacenter/scala-debug-adapter/pull/553) (not published yet)

There are a few inconsistencies left in my implementation on the VS Code extension:

- [x] If the setting `scalaHotReloadOn` is not ticked, the button still appear (we need to tick and untick to make it disappear)
- [x] If HCR is not supported by the debugger, it show a warning message "no changed classes found", which is not relevant

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,25 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "files.watcherExclude": {
     "**/target": true
-  }
+  },
+  "workbench.colorCustomizations": {
+    "activityBar.activeBackground": "#0083dc",
+    "activityBar.background": "#0083dc",
+    "activityBar.foreground": "#e7e7e7",
+    "activityBar.inactiveForeground": "#e7e7e799",
+    "activityBarBadge.background": "#9f005f",
+    "activityBarBadge.foreground": "#e7e7e7",
+    "commandCenter.border": "#e7e7e799",
+    "sash.hoverBorder": "#0083dc",
+    "statusBar.background": "#0065a9",
+    "statusBar.foreground": "#e7e7e7",
+    "statusBarItem.hoverBackground": "#0083dc",
+    "statusBarItem.remoteBackground": "#0065a9",
+    "statusBarItem.remoteForeground": "#e7e7e7",
+    "titleBar.activeBackground": "#0065a9",
+    "titleBar.activeForeground": "#e7e7e7",
+    "titleBar.inactiveBackground": "#0065a999",
+    "titleBar.inactiveForeground": "#e7e7e799"
+  },
+  "peacock.color": "#0065A9"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,25 +10,5 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "files.watcherExclude": {
     "**/target": true
-  },
-  "workbench.colorCustomizations": {
-    "activityBar.activeBackground": "#0083dc",
-    "activityBar.background": "#0083dc",
-    "activityBar.foreground": "#e7e7e7",
-    "activityBar.inactiveForeground": "#e7e7e799",
-    "activityBarBadge.background": "#9f005f",
-    "activityBarBadge.foreground": "#e7e7e7",
-    "commandCenter.border": "#e7e7e799",
-    "sash.hoverBorder": "#0083dc",
-    "statusBar.background": "#0065a9",
-    "statusBar.foreground": "#e7e7e7",
-    "statusBarItem.hoverBackground": "#0083dc",
-    "statusBarItem.remoteBackground": "#0065a9",
-    "statusBarItem.remoteForeground": "#e7e7e7",
-    "titleBar.activeBackground": "#0065a9",
-    "titleBar.activeForeground": "#e7e7e7",
-    "titleBar.inactiveBackground": "#0065a999",
-    "titleBar.inactiveForeground": "#e7e7e799"
-  },
-  "peacock.color": "#0065A9"
+  }
 }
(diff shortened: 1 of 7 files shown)


iusildra:
It will need both upcoming versions of the debugger (for the implementation) and metals (for the sync with compilation) to be fully functional

Also, before merging I'll need to change the metals server's version before which HCR would not work to give proper information

packages/metals-vscode/src/hotCodeReplace.ts
```diff
@@ -0,0 +1,112 @@
...
+  const debugSession: DebugSession | undefined =
+    vscode.debug.activeDebugSession;
+  if (!debugSession) {
+    return;
+  }
+
+  const serverVersion =
+    vscode.workspace.getConfiguration("metals").get<string>("serverVersion") ??
+    "";
+  if (serverVersion < "1.0.1") {
```
tgodzik, line 44: I think we could just make sure we update DAP in Metals and in Bloop first and once that is available we can merge this PR and update the version to the most recent snapshot at the time. If anyone switches to an older version of Metals it's fine if we fail.

tgodzik (changes requested):
Thanks for the great work! It will be amazing for the users.

Let's wait until we update scala debug adapter properly in Bloop and sbt

packages/metals-vscode/package.json
```diff
@@ -218,6 +218,11 @@
     "configuration": {
       "title": "Metals",
       "properties": {
+        "metals.debug.settings.hotCodeReplace": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Allow Hot Code Replace (HCR) while debugging. Only available since Metals 1.0.X"
```
tgodzik, line 224:
```suggestion
          "markdownDescription": "Allow Hot Code Replace (HCR) while debugging."
```

packages/metals-vscode/package.json
```diff
@@ -1085,7 +1105,8 @@
     "metals-languageclient": "file:../metals-languageclient",
     "promisify-child-process": "4.1.1",
     "semver": "^7.5.2",
-    "vscode-languageclient": "8.1.0"
+    "vscode-languageclient": "8.1.0",
+    "vscode-extension-telemetry-wrapper": "^0.13.3"
```
tgodzik, line 1109: Is this needed?

packages/metals-vscode/package.json
```diff
@@ -218,6 +218,11 @@
     "configuration": {
       "title": "Metals",
       "properties": {
+        "metals.debug.settings.hotCodeReplace": {
```
tgodzik, line 221:
```suggestion
        "metals.hotCodeReplace": {
```
maybe is enough?

packages/metals-vscode/src/hotCodeReplace.ts
```diff
@@ -0,0 +1,100 @@
...
+  const response = await redefineRequest;
+
+  if (response?.errorMessage) {
+    vscode.window.showErrorMessage(response.errorMessage);
+    return;
+  }
+
+  const NO_HCR = "Disable HCR";
+
+  if (!response?.changedClasses?.length) {
```
tgodzik, line 71: Is this needed, why would we want to disable HCR here?

packages/metals-vscode/src/hotCodeReplace.ts
```diff
@@ -0,0 +1,100 @@
...
+    const res = await vscode.window.showWarningMessage(
+      "No classes were reloaded",
+      "Ok",
+      NO_HCR
+    );
+
+    if (res === NO_HCR) {
+      vscode.workspace
+        .getConfiguration("metals.debug.settings")
+        .update("hotCodeReplace", false, vscode.ConfigurationTarget.Workspace);
```
tgodzik, line 81: Probably would be good to put `"hotCodeReplace"` and `"scalaHotReloadOn"` into const values

.vscode/settings.json
```diff
@@ -10,5 +10,25 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "files.watcherExclude": {
     "**/target": true
-  }
+  },
+  "workbench.colorCustomizations": {
```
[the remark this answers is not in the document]
iusildra: oups, yeah will do

packages/metals-vscode/src/hotCodeReplace.ts
iusildra: Forgot to remove it. It was when I wasn't sure of how to handle the case where HCR is not supported

packages/metals-vscode/package.json
iusildra: Good point, `registerCommand` would do the same

adpi2:
@tgodzik  this PR is ready to be reviewed and merged.

It adds the lightning button in the debug tool bar to trigger a hot code reload.

packages/metals-vscode/package.json
```diff
@@ -343,6 +343,11 @@
           "default": false,
           "markdownDescription": "When this option is enabled, when user pastes any snippet into a Scala file, Metals will try to adjust the indentation to that of the current cursor."
         },
+        "metals.debug.hotCodeReplace": {
```
tgodzik, line 346: MAybe we don't need a setting actually? If it doesn't work then the button would not work and that's fine as long it doesn't break anything else

adpi2: Okay, I'll remove the setting

adpi2: I don't think we need

tgodzik (approved):
LGTM!

PR #1422, opened 2023-08-18
chore: Set runProvider as true

tgodzik:
Previously, we would not set runProvider and rely entirely on debugProvider, however in some cases where debug is not supported by the build server, we might actually want to use just run. Now, with this set it's possible to see that VS Code actually supports running on the client side and that can be used.


tgodzik:
I will go ahead and merge this since this is a minor change.

Issue #1436, opened 2023-11-09
"Internal error" launch config error message


harpocrates:
**Describe the bug**

When using "Run and Debug" with a launch config, the cryptic "Internal error." dialog shows up in a variety of situations where there's some sort of error.

At least two situations that cause this:

  1. when trying to run non-compiling code, it would be vastly preferable to tell the user their code is not compiling and maybe even focus the "Problems" view so those errors are brought to the front.
  2. when trying to run after specifying a main class that doesn't exist. Here there is a "Main class not found: example.Hello2" error notification bubble, but it would be much better to include that information in the "Internal error" dialog that pops up.

**To Reproduce** Steps to reproduce the behavior:

1. Create some small project (I just followed defaults in `sbt new scala/scala-seed.g8`)
3. Import the project into metals
4. Create a launch configuration to call the default entry point (confirm it works)
5. Edit the project code to make it non-compiling
6. Try to run again, see the "Internal error." message

Here was my launch config:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "scala",
            "request": "launch",
            "name": "Hello World",
            "mainClass": "example.Hello",
            "buildTarget": "root",
            "args": [],
            "jvmOptions": [],
            "env": {}
        }

    ]
}
```

**Expected behavior**

The dialog that pops up should include some information about what went wrong.

**Screenshots**

![Screenshot 2023-11-09 at 12 38 10 PM](https://github.com/scalameta/metals-vscode/assets/10766081/75bf5f0a-99ee-4158-be32-82dc8ec03694)

**Installation:**

- Operating system: macOS and Windows
- VSCode version: 1.84.0
- VSCode extension version: 1.26.0 and 1.26.1
- Metals version: (found in VSCode settings, under `metals.serverVersion`)

tgodzik:
Thanks for reporting! It would be great to improve that behaviour. Last time I checked this was not possible to do sensibly, but I might have missed soemthing.

tgodzik:
I found a way around that, we can just throw an error with a custom message.

tgodzik:
Should be fixed in the newest metals snapshots

PR #1438, opened 2023-11-17
improvement: Throw an error with a correct message

tgodzik:
Thanks to this the proper message will be displayed.

Fixes https://github.com/scalameta/metals-vscode/issues/1436

Screenshot from 2023-11-17 19-15-49

### DIFF
--- a/packages/metals-vscode/src/debugger/scalaDebugger.ts
+++ b/packages/metals-vscode/src/debugger/scalaDebugger.ts
@@ -22,8 +22,9 @@ import { currentWorkspaceFolder } from "../util";
 const configurationType = "scala";
 
 export interface DebugSession {
-  name: string;
-  uri: string;
+  name: string | undefined;
+  uri: string | undefined;
+  error: string | undefined;
 }
 
 export function initialize(outputChannel: vscode.OutputChannel): Disposable[] {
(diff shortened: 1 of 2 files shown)


packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -22,8 +22,9 @@ import { currentWorkspaceFolder } from "../util";
 const configurationType = "scala";
 
 export interface DebugSession {
-  name: string;
-  uri: string;
+  name: string | undefined;
+  uri: string | undefined;
+  error: string | undefined;
 }
```
kpodsiad, line 28: are `{name: string, uri: string}` and `{error: string}` mutually exclusive?

packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -191,13 +199,18 @@ class ScalaDebugServerFactory implements vscode.DebugAdapterDescriptorFactory {
         session.configuration
       );
 
-      if (debugSession === undefined) {
-        return null;
-      } else {
+      if (debugSession !== undefined && debugSession.uri !== undefined) {
```
kpodsiad, line 202:
```suggestion
      if (debugSession?.uri) {
```

packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -191,13 +199,18 @@ class ScalaDebugServerFactory implements vscode.DebugAdapterDescriptorFactory {
...
         return debugServerFromUri(debugSession.uri);
+      } else if (
+        debugSession !== undefined &&
+        debugSession.error !== undefined
+      ) {
```
kpodsiad, line 207:
```suggestion
      } else if (debugSession?.error) {
```

packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -191,13 +199,18 @@ class ScalaDebugServerFactory implements vscode.DebugAdapterDescriptorFactory {
...
+        throw new Error(debugSession.error);
+      } else {
+        throw new Error("Could not start debug adapter");
       }
```
kpodsiad, line 211: `else { throw new Error("Could not start debug adapter") }` seems redundant, as there's fallback return at the end already

tgodzik:
I realized, we can actually do it all on the server side by using lsp4j's exception:

https://github.com/scalameta/metals/pull/5864

PR #1449, opened 2024-01-04
bugfix: Show error properly if lenses cannot be run

tgodzik:


packages/metals-vscode/src/extension.ts
```diff
@@ -589,22 +589,32 @@ function launchMetals(
...
-            if (!wasStarted) {
-              window.showErrorMessage("Debug session not started");
+          scalaDebugger.start(false, param).then(
+            (wasStarted) => {
+              if (!wasStarted) {
+                window.showErrorMessage("Debug session not started");
+              }
+            },
+            (reason: Error) => {
+              window.showErrorMessage(reason.message);
```
kpodsiad, line 599: did you check that typeof reason in runtime is `Error`? Where does it come from?

kpodsiad (approved):

packages/metals-vscode/src/extension.ts
tgodzik: That's what I was getting in runtime, should I do a check to make sure it's an Error?

tgodzik: I added a check just in case.

Issue #1450, opened 2024-01-04
Use coursier for downloading Java


tgodzik:
Currently, we use a combination of `locate-java-home` and Jabba to get Metals with Java setup. However, both of those options are not very well maintained.

Instead, I think we should work out a much more stable approach with Metals using its' own JVM.

This can be done using coursier, which would make us reuse any Java that was previously downloaded using it or via Scala CLI.

We would need to follow these steps:
1. Check if coursier is on PATH (this is already implemented) and if not try to download the native image (as explained https://get-coursier.io/docs/cli-installation#native-launcher). We should make sure that it's not downloaded every time. We could possibly have a version inside the repo and update it when releasing. The cs executable can be kept in the metals extension directory.
2. Download Java `cs java --jvm 17 --setup` 
3. Check Java Home `cs java-home --jvm 17`
4. Start metals with that Java Home
5. Bloop should use the exact same Java Home.
6. We should add a simple setting that allows user to choose JVM version for Metals itself. (8/11/17/21) currently and ideally (17/21) later on when we migrate to new sockets in Metals server.
7. javaHome setting would only be used for running bloopInstall or bsp launchers.

Questions:
- should we still try to find JAVA_HOME? This might be a bit complex as it would require us to check the version of the particular java manually.

PR #1452, opened 2024-01-11
improvement: download native coursier and use it to download java

kasiaMarek:
resolves: https://github.com/scalameta/metals-vscode/issues/1450


packages/metals-vscode/package.json
```diff
@@ -270,7 +270,12 @@
         "metals.javaHome": {
           "type": "string",
           "scope": "machine-overridable",
-          "markdownDescription": "Optional path to the Java home directory. Requires reloading the window.\n\nDefaults to the most recent Java version between 8 and 17 (inclusive) computed by the `locate-java-home` npm package."
+          "markdownDescription": "Optional path to the Java home directory. Requires reloading the window.\n\nDefaults to Java 17 (or another version specified in javaVersion) downloaded by coursier."
+        },
+        "metals.javaVersion": {
+          "type": "string",
+          "scope": "machine-overridable",
+          "markdownDescription": "Optional Java version to start metals with, can be one of \"8\", \"11\", \"17\", \"21\"."
```
tgodzik, line 278: We could make it an enum so that it's easier for the users to choose one.

packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -19,17 +25,67 @@ interface PackedChildPromise {
...
+  coursierFetchPath: string,
   output: OutputChannel
-): Promise<PackedChildPromise> {
-  const fetchProperties = serverProperties.filter(
-    (p) => !p.startsWith("-agentlib")
-  );
+): Promise<{ coursier: string; javaHome: string }> {
+  const handleOutput = (out: Buffer) => {
+    const msg = out.toString().trim();
+    output.appendLine("CS: " + msg);
```
tgodzik, line 35:
```suggestion
    output.appendLine("Coursier: " + msg);
```
?

packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -19,17 +25,67 @@ interface PackedChildPromise {
...
+          "Failed to fetch coursier. You may want to try installing coursier manually and adding it to PATH."
+        );
+        throw err;
+      });
+  };
+
+  const resolveJavaHome = async (coursier: string) => {
+    await run(
+      coursier,
+      ["java", "--jvm", javaVersion, "--setup"],
```
tgodzik, line 63: Did you manage to check if https://get-coursier.io/docs/cli-java#system-jvm-detection is true? Looks like it doesn't work for me and even if JAVA_HOME is set to JDK 17 it will still try to download :thinking: 

Would be great to make sure we don't only rely on coursier, because would not manage to get Metals working when Java cannot be downloaded.

If coursier is at fault we may have to do a fallback or chekc manually beforehand

tgodzik (commented):
Great work! I only would make sure that we can detect Java if it exists on the system.

packages/metals-languageclient/src/fetchMetals.ts
kasiaMarek: I'll check manually. It seems to be the same for me, it downloads JDK 17  even if Java home is 17. I'm a bit confused by the docs, those don't really explain how it should behave when java version is set with respect to JAVA_HOME.

tgodzik: This might be broken then :thinking: We might need to still use the find java home npm package based on the selected JDK version.

packages/metals-vscode/src/extension.ts
```diff
@@ -195,22 +187,24 @@ function fetchAndLaunchMetals(
   /* eslint-enable @typescript-eslint/no-non-null-assertion */
 
   const coursierMirror = getCoursierMirrorPath(config);
+  const { coursier, javaHome } = await metalsLanguageClient.setupCoursier(
+    javaVersion,
+    context.extensionPath,
```
tgodzik, line 192: Btw. I think this changes with every version, we should most likely put coursier somewhere where we would not need to redownload with every extension version change

packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -135,3 +172,64 @@ export function calcServerDependency(serverVersion: string): string {
...
+      };
+      return acc ? acc.then(() => res()) : res();
+    }, initValue);
+    return (await out)?.code;
+  }
+
+  if (process.platform == "win32") {
+    // Windows
+    return runChainedCommands([
+      "curl -fLo cs-x86_64-pc-win32.zi https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-win32.zip",
```
tgodzik, line 199: We should specify a version of coursier instead of latests, this will be safer if coursier breaks at some version

packages/metals-languageclient/src/getJavaHome.ts
```diff
@@ -9,64 +9,89 @@ import {
   IJavaHomeInfo,
 } from "@viperproject/locate-java-home/js/es5/lib/interfaces";
 import { toPromise } from "./util";
-//import os from "os";
 import fs from "fs";
 import path from "path";
+import { spawn } from "promisify-child-process";
+
+export type JavaVersion = "8" | "11" | "17" | "21";
```
tgodzik, line 16: After https://github.com/scalameta/metals/pull/6092 we will be able to leave 17 and 21

packages/metals-vscode/src/extension.ts
```diff
@@ -195,22 +188,31 @@ function fetchAndLaunchMetals(
   /* eslint-enable @typescript-eslint/no-non-null-assertion */
 
   const coursierMirror = getCoursierMirrorPath(config);
+
+  const metalsDirPath = path.join(os.homedir(), ".metals");
```
tgodzik, line 192:
```suggestion
  const metalsDirPath = metalsDir(ConfigurationTarget.Global);
```
maybe? we already have it in utils

packages/metals-languageclient/src/setupCoursier.ts
```diff
@@ -0,0 +1,197 @@
...
+  PromisifySpawnOptions,
+  spawn,
+} from "promisify-child-process";
+import { JavaVersion, getJavaHome } from "./getJavaHome";
+import { OutputChannel } from "./interfaces/OutputChannel";
+import path from "path";
+import fs from "fs";
+
+const coursierVersion = "v2.1.8";
+const coursierCommit = "11b428f35ca84a598ca30cce1c35ae4f375e5ee3";
```
tgodzik, line 13: Could you add a comment why it's needed to specify commit?

tgodzik (commented):
Let's try it after the release is out

tgodzik (approved):
LGTM, let's try it!

PR #1472, opened 2024-03-13
don't override `JAVA_HOME` env variable

kasiaMarek:
connected to: https://github.com/scalameta/metals/pull/6204

### DIFF
--- a/packages/metals-languageclient/src/getJavaConfig.ts
+++ b/packages/metals-languageclient/src/getJavaConfig.ts
@@ -40,7 +40,6 @@ export function getJavaConfig({
   const extraEnv = {
     ...coursierRepositories,
     ...coursierMirrors,
-    JAVA_HOME: javaHome,
   };
 
   return {


packages/metals-languageclient/src/getJavaConfig.ts
```diff
@@ -40,7 +40,6 @@ export function getJavaConfig({
   const extraEnv = {
     ...coursierRepositories,
     ...coursierMirrors,
-    JAVA_HOME: javaHome,
```
tgodzik, line 43: Maybe we should still set it if not available?

kasiaMarek: It shouldn't be needed, since we always fallback to checking system properties java home.

tgodzik: We can change that if it turns out to be a problem sure

tgodzik (approved):

PR #1476, opened 2024-03-20
chore(deps): remove `@types/glob`

yoshinorin:
## Description 

Remove `@types/glob` dependency.

## Why

The `glob 9+` (metals-vscode project are using `9.3.1`) supports TypeScript, there is no need to install TypeDefinition dependencies.

Please see: [glob v9 CHANGELOGS](https://github.com/isaacs/node-glob/blob/v9/changelog.md#90)

-----

Thank you :)


tanishiking (approved):
Thanks!

PR #1479, opened 2024-04-07
allow to run main of different files in parallel

kaplan-shaked:


packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -76,10 +76,10 @@ async function runMain(main: ExtendedScalaRunMain): Promise<boolean> {
       },
       {}
     );
-
+    const file = currentFile();
```
tgodzik, line 79: This might prove problematic if we run but the file is not focused :thinking: 

Actually, we can already make it better via `main.class` which will be the name of the class being run, which should be unique.

kaplan-shaked: Good idea. testing it now.

tgodzik (approved):
Thanks!

packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -79,7 +79,7 @@ async function runMain(main: ExtendedScalaRunMain): Promise<boolean> {
 
     const shellOptions = { ...platformSpecificOptions(), env };
     const task = new Task(
-      { type: "scala", task: "run" },
+      { type: "scala " + main.data.class, task: "run" },
```
tgodzik, line 82:
```suggestion
      { type: "scala", task: "run", class: main.data.class },
```
I read the API for Task and it seems `type` should stay as "scala" and we can just add another property which looks to be working. If we ever need to use https://code.visualstudio.com/api/extension-guides/task-provider#task-definition then we would avoid weird issue.

I should have thought of it in the first place, thanks for working on this!

kaplan-shaked: Done.

tgodzik (approved):
Awesome, thanks!

PR #1484, opened 2024-04-18
bugfix: Remove locate-java-home

tgodzik:
It seems it uses promises underneath and if anything throws it just hangs, so Metals will never start. I manged to reproduce it locally, though hopefully doesn't reproduce in a lot of machines


tgodzik:
One thing we might need want to do is at least try to read java from PATH, not just from env.

packages/metals-languageclient/yarn.lock
```diff
@@ -652,13 +652,13 @@
   dependencies:
     "@types/yargs-parser" "*"
 
-"@viperproject/locate-java-home@^1.1.5":
-  version "1.1.11"
-  resolved "https://registry.yarnpkg.com/@viperproject/locate-java-home/-/locate-java-home-1.1.11.tgz#9ae7c281ab005bcbe8b730fee495def0440385f3"
-  integrity sha512-cnwxSsB1QB4eeerRd8XsBEDh/BHD1726g/UqA3RAvPBlbZVqzm75jyq+0p9hsVQdtHxUdLRzpjSktck8YEuaAQ==
+"@viperproject/locate-java-home@^1.1.15":
```
kasiaMarek, line 655: do we even need this package anymore?

tgodzik:
I will merge to let people test out since it seems that it fails for more people. I will add code for getting Java from PATH later on

PR #1486, opened 2024-04-21
improvement: Search PATH for Java to use

tgodzik:
Previously, we would rely on the separate library for detecting java, but we had no control on whether it worked and it doesn't seem actively used by a lot of people.

Now, we first check JAVA_HOME and then search PATH for the actual Java to use. If we can't find any of that we download ourselves. This seems simple enought logic.

### DIFF
--- a/packages/metals-languageclient/src/getJavaHome.ts
+++ b/packages/metals-languageclient/src/getJavaHome.ts
@@ -64,7 +64,7 @@ export async function fromPath(
     outputChannel.appendLine(
       `Found java executable under ${javaExecutable} that resolves to ${realJavaPath}`
     );
-    let possibleJavaHome = path.dirname(path.dirname(realJavaPath));
+    const possibleJavaHome = path.dirname(path.dirname(realJavaPath));
     const isValid = await validateJavaVersion(
       possibleJavaHome,
       javaVersion,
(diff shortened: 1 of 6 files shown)


packages/metals-languageclient/src/util.ts
```diff
@@ -1,10 +1,43 @@
...
       new Promise((resolve, reject) => pipe(res, E.fold(reject, resolve)))
   );
 }
+
+export function findOnPath(names: string[]) {
+  const envPath = process.env["PATH"] || process.env["Path"];
+  const isWindows = process.platform === "win32";
+
+  if (envPath) {
+    const possibleCoursier = envPath
```
kasiaMarek, line 19: fix name

packages/metals-languageclient/src/getJavaHome.ts
```diff
@@ -52,6 +54,26 @@ async function validateJavaVersion(
...
+  javaVersion: JavaVersion,
+  outputChannel: OutputChannel
+): Promise<string | undefined> {
+  let javaExecutable = findOnPath(["java"]);
+  if (javaExecutable) {
+    let realJavaPath = realpathSync(javaExecutable);
+    outputChannel.appendLine(
+      `Found java executable under ${javaExecutable} that resolves to ${realJavaPath}`
+    );
+    let possibleJavaHome = path.dirname(path.dirname(realJavaPath));
```
kasiaMarek, line 67:
```suggestion
    const possibleJavaHome = path.dirname(path.dirname(realJavaPath));
```
I don't think there is any reason for `let` here (also line 61 and 63).

kasiaMarek (approved):
lgtm

PR #1498, opened 2024-05-11
use properties and environment when fetching metals

### LINKED ISSUE #1497: Missing HTTP proxy settings for Coursier when fetching Metals
**Describe the bug**

Since [v1.30.0](https://github.com/scalameta/metals-vscode/releases/tag/v1.30.0), Coursier fails to fetch Metals when an HTTP proxy is needed.

**To Reproduce**

1. Install v1.30.0 of the extension
1. Configure the `serverProperties` setting with the HTTP proxy, e.g.:
   - `-Dhttps.proxyHost=...`
   - `-Dhttps.proxyPort=...`
1. Trigger the Metals extension install process
2. It fails to fetch Metals with an unknown host exception (due to the missing proxy configuration)

This happens both if Coursier is already present on the PATH and if fetched by the extension.

**Expected behavior**

During the install process Coursier uses the HTTP proxy configuration from the `serverProperties` setting to successfully fetch Metals.

**Screenshots**

None.

**Installation:**

- Operating system: Linux
- VSCode version: 1.87.2
- VSCode extension version: 1.30.3
- Metals version: 1.3.0

**Additional context**

In v1.29.0 and earlier the `fetchMetals` method [includes](https://github.com/scalameta/metals-vscode/blob/v1.29.0/packages/metals-languageclient/src/fetchMetals.ts#L64) the `serverProperties` settings as JVM arguments when invoking the default included Coursier. However, in v1.30.0 and newer [no external settings are included](https://github.com/scalameta/metals-vscode/blob/v1.30.0/packages/metals-languageclient/src/fetchMetals.ts#L24) in the Coursier call to fetch Metals.

I believe the same issue occurs when trying to specify an HTTP proxy using `$JAVA_OPTS` or the `.jvmopts` file. Those options are [included in v1.29.0](https://github.com/scalameta/metals-vscode/blob/v1.29.0/packages/metals-languageclient/src/fetchMetals.ts#L63), but [not in v1.30.0](https://github.com/scalameta/metals-vscode/blob/v1.30.0/packages/metals-languageclient/src/fetchMetals.ts#L24). This is just based on the code, I wasn't using it that way so I didn't test it.

This issue might also be present in v1.29.0 [when using an external copy](https://github.com/scalameta/metals-vscode/blob/v1.29.0/packages/metals-languageclient/src/fetchMetals.ts#L83) of Couriser, but I didn't test that either.

**Search terms**

coursier, proxy

mliarakos:
Fixes #1497

### DIFF
--- a/packages/metals-languageclient/src/fetchMetals.ts
+++ b/packages/metals-languageclient/src/fetchMetals.ts
@@ -4,6 +4,7 @@ import { JavaConfig } from "./getJavaConfig";
 
 interface FetchMetalsOptions {
   serverVersion: string;
+  serverProperties: string[];
   javaConfig: JavaConfig;
 }
 
@@ -17,15 +18,19 @@ interface PackedChildPromise {
 
 export async function fetchMetals({
   serverVersion,
-  javaConfig: { coursier },
+  serverProperties,
+  javaConfig: { javaOptions, coursier, extraEnv },
 }: FetchMetalsOptions): Promise<PackedChildPromise> {
   const serverDependency = calcServerDependency(serverVersion);
+
+  const fetchProperties = serverProperties.filter((p) => !p.startsWith("-agentlib"));
+  const javaArgs = javaOptions.concat(fetchProperties).map((p) => `-J${p}`);
 
   const coursierArgs = [
     "fetch",
     "-p",
     "--ttl",
-    // Use infinite ttl to avoid redunant "Checking..." logs when using SNAPSHOT
+    // Use infinite ttl to avoid redundant "Checking..." logs when using SNAPSHOT
     // versions. Metals SNAPSHOT releases are effectively immutable since we
     // never publish the same version twice.
     "Inf",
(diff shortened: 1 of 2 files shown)


packages/metals-languageclient/src/fetchMetals.ts
```diff
@@ -17,15 +18,21 @@ interface PackedChildPromise {
 
 export async function fetchMetals({
   serverVersion,
-  javaConfig: { coursier },
+  serverProperties,
+  javaConfig: { javaOptions, coursier, extraEnv },
 }: FetchMetalsOptions): Promise<PackedChildPromise> {
   const serverDependency = calcServerDependency(serverVersion);
 
+  const fetchProperties = serverProperties.filter(
```
tgodzik, line 26: I would just warn here, since we do debug metals sometimes and that requires setting debug settings.

Alternatively we could check if it's set to quiet=y.

tgodzik (changes requested):
Looks good! Just one small suggestion. Thanks for working on this!

packages/metals-languageclient/src/fetchMetals.ts
mliarakos: I restored that `filter` call [from v1.29.0](https://github.com/scalameta/metals-vscode/blob/v1.29.0/packages/metals-languageclient/src/fetchMetals.ts#L30), but I'm not familiar with why it was there to being with.

To make sure I understand, are you asking to emit a warning message if `-agentlib` is present indicating that it isn't being passed to Coursier?

tgodzik: Ach right, my bad. I forgot it was just for coursier, I assumed too much :sweat: It's fine as it is, thanks!

tgodzik (approved):
Thanks! LGTM

PR #1513, opened 2024-05-29
improvement: Print Java all at once

tgodzik:
image

It seems that when writing on date we can get one line first and then the other, which would duplicate the `(`${javaBin} -version:`)` message

### DIFF
--- a/packages/metals-languageclient/src/getJavaHome.ts
+++ b/packages/metals-languageclient/src/getJavaHome.ts
@@ -37,7 +37,6 @@ async function validateJavaVersion(
     });
 
     javaVersionOut.stderr?.on("data", (out: Buffer) => {
-      outputChannel.appendLine(`${javaBin} -version:`);
       const msg = "\t" + out.toString().trim().split("\n").join("\n\t");
       outputChannel.appendLine(msg);
     });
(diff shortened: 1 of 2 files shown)


packages/metals-languageclient/src/getJavaHome.ts
```diff
@@ -35,14 +35,12 @@ async function validateJavaVersion(
...
+    const javaInfoStr = (await javaVersionOut).stderr as string;
 
-    javaVersionOut.stderr?.on("data", (out: Buffer) => {
-      outputChannel.appendLine(`${javaBin} -version:`);
-      const msg = "\t" + out.toString().trim().split("\n").join("\n\t");
-      outputChannel.appendLine(msg);
-    });
+    const msg = "\t" + javaInfoStr.toString().trim().split("\n").join("\n\t");
+    outputChannel.appendLine(`${javaBin} -version:`);
+    outputChannel.appendLine(msg);
```
kasiaMarek, line 42: I think the reason we had `javaVersionOut.stderr?.on("data", ...)` is because, if running `java -version` fails `await javaVersionOut` will throw, so `(await javaVersionOut).stderr as string` won't get printed.

kasiaMarek (changes requested):

packages/metals-languageclient/src/getJavaHome.ts
tgodzik: Ok, good point. I will just remove the initial line with `-version` there. It's not really adding anything and we already print which java is being looked into before that

kasiaMarek (approved):

Issue #1509, opened 2024-05-19
Can't start Metals server with JDK with cyrillic characters in the path


divijan:
**Unable to start Metals extension with cyrillic characters in path under Windows**
Cannot find java executable in path with cyrillic characters in it.

Metals Extension output:

```
Metals version: 1.3.1
Using coursier located at C:\Users\до\.metals\cs.exe //here my two-character cyrillic username is displayed correctly, I just changed it for security reasons
C:\Program Files\Eclipse Adoptium\jdk-8.0.412.8-hotspot\bin\java -version:
openjdk version "1.8.0_412"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_412-b08)
OpenJDK 64-Bit Server VM (Temurin)(build 25.412-b08, mixed mode)
Found java executable under C:\Program Files\Eclipse Adoptium\jdk-8.0.412.8-hotspot\bin\java.exe that resolves to C:\Program Files\Eclipse Adoptium\jdk-8.0.412.8-hotspot\bin\java.exe
C:\Program Files\Eclipse Adoptium\jdk-8.0.412.8-hotspot\bin\java -version:
openjdk version "1.8.0_412"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_412-b08
C:\Program Files\Eclipse Adoptium\jdk-8.0.412.8-hotspot\bin\java -version:
)
OpenJDK 64-Bit Server VM (Temurin)(build 25.412-b08, mixed mode)
No installed java with version 17 found. Will fetch one using coursier.
Coursier: openjdk version "17" 2021-09-14
OpenJDK Runtime Environment Temurin-17+35 (build 17+35)
OpenJDK 64-Bit Server VM Temurin-17+35 (build 17+35, mixed mode, sharing)
Using Java Home: C:\Users\??\AppData\Local\Coursier\cache\arc\https\github.com\adoptium\temurin17-binaries\releases\download\jdk-17%252B35\OpenJDK17-jdk_x64_windows_hotspot_17_35.zip\jdk-17+35 //first time I see ?? instead of cyrillic characters
[Error - 5:47:06 PM] Metals client: couldn't create connection to server.
[...]

PR #1514, opened 2024-05-29
improvement: Retry with coursier jar if anything fails

tgodzik:
Previously, if anything failed with coursier native image, for example we wouldn't have proper encoding then user would not be able to do anything.

Now, if there is Java avaialable we can retry with the embedded coursier jar.

Might help with https://github.com/scalameta/metals-vscode/issues/1509

### DIFF
--- a/packages/metals-vscode/src/extension.ts
+++ b/packages/metals-vscode/src/extension.ts
@@ -207,7 +207,8 @@ function debugInformation(
 async function fetchAndLaunchMetals(
   context: ExtensionContext,
   serverVersion: string,
-  javaVersion: JavaVersion
+  javaVersion: JavaVersion,
+  forceCoursierJar = false
 ) {
   outputChannel.appendLine(`Metals version: ${serverVersion}`);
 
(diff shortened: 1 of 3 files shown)


packages/metals-vscode/src/extension.ts
```diff
@@ -228,9 +229,24 @@ async function fetchAndLaunchMetals(
...
 
+  const canRetryWithJar =
+    javaHome && !coursier.endsWith(".jar") && !forceCoursierJar;
+
+  function retry(error: any): Promise<any> {
+    if (canRetryWithJar) {
+      outputChannel.appendLine(
+        "Trying again with the embedded coursier. This might take longer."
+      );
+      return fetchAndLaunchMetals(context, serverVersion, javaVersion, true);
```
kasiaMarek, line 244: If you rerun the whole `fetchAndLaunchMetals` it will again try to download coursier, this seems redundant.

kasiaMarek (approved):

PR #1515, opened 2024-05-30
Prefer the metals VSCode extension to run in the "Workspace" as opposed to "UI"

Quafadas:
I've been experimenting with using metals in codespaces. 

Under certain conditions (e.g. firewalls), I believe the metals extension may struggle to communicate with it's build server if running in the on the UI side of the architecture. 

This change says "I prefer to run in the workspace (i..e next to the code, and where the build server also wants to be), but I can run on the UI side if you want me to". I think metals might also work in the UI under this architecture under permissive network conditions, so I have't chosen to prevent the UI side entirely.

I've had to manually set this flag a few times now, and I think it's more likely that metals will work better as a workspace extension universally, so I'm putting up for a PR as part of the extension manifest. 

Here is what I believe the relevant docs are;

https://code.visualstudio.com/api/advanced-topics/remote-extensions#architecture-and-extension-kinds

https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location

https://code.visualstudio.com/api/references/extension-manifest

### DIFF
--- a/packages/metals-vscode/package.json
+++ b/packages/metals-vscode/package.json
@@ -42,6 +42,7 @@
     "Programming Languages",
     "Debuggers"
   ],
+  "extensionKind": ["workspace", "ui"],
   "activationEvents": [
     "onCommand:metals.start-server",
     "onCommand:metals.new-scala-project",
--- a/packages/metals-vscode/package.json
+++ b/packages/metals-vscode/package.json
@@ -42,7 +42,10 @@
     "Programming Languages",
     "Debuggers"
   ],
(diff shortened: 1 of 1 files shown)


tgodzik (changes requested):
Looks good! Could you just run `yarn lint` on packages/metals-vscode ?

Quafadas:
Thanks for the tip ... I just spent a few minutes trying to figure out prettier! Wrong starting point :-)... 

change in bound.

Quafadas:
That got it 🙏

PR #1521, opened 2024-07-22
feature: Also run plain Java when noDebug from launch.json

tgodzik:
Previously, if user run code from code lenses they would get a java command running without DAP. However, if they wanted to add parameters and run via launch.json they would run it via DAP.

Now, even if running from configuration we will run with simple java command. One issue here is that we need to start debugger anyway of run via launch.json, but I worked around it by just running echo instead.

Needs https://github.com/scalameta/metals/pull/6306

### DIFF
--- a/packages/metals-languageclient/src/interfaces/DebugDiscoveryParams.ts
+++ b/packages/metals-languageclient/src/interfaces/DebugDiscoveryParams.ts
@@ -1,6 +1,12 @@
 export interface DebugDiscoveryParams {
-  path: string;
+  path: string | undefined;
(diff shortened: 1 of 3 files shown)


tgodzik:
This should be good to review now.

kasiaMarek:
But https://github.com/scalameta/metals/pull/6306 was merged after the release.

tgodzik:
I just meant if you could take a look at it, no pressure. Will include it in the next release.

kasiaMarek (approved):

Issue #1526, opened 2024-08-13
Cannot start metals server on macOS with remote SSH


KireinaHoro:
**Describe the bug**

I'm trying to import a Mill-based project through remote SSH with Metals.  Seems like the plugin locates JDK correctly and downloads all the required Maven dependencies, but couldn't start the server.  Log:

```
etals version: 1.3.5
Using coursier located at /Users/jsteward/.metals/cs
Searching for Java on PATH. Found java executable under /etc/profiles/per-user/jsteward/bin/java that resolves to /nix/store/hxskzr9a3m0prpy9l41bnw3rmmahvdb6-zulu-ca-jdk-21.0.2/zulu-21.jdk/Contents/Home/bin/java
	openjdk version "21.0.2" 2024-01-16 LTS
	OpenJDK Runtime Environment Zulu21.32+17-CA (build 21.0.2+13-LTS)
	OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing)
Using Java Home: /nix/store/hxskzr9a3m0prpy9l41bnw3rmmahvdb6-zulu-ca-jdk-21.0.2/zulu-21.jdk/Contents/Home
[Error - 11:53:03] Server initialization failed.
  Message: Internal error.
  Code: -32603 
java.nio.file.FileSystemNotFoundException: Provider "vscode-remote" not installed
	at java.base/java.nio.file.Path.of(Path.java:213)
	at java.base/java.nio.file.Paths.get(Paths.java:98)
	at scala.meta.internal.mtags.MtagsEnrichments$XtensionURIMtags.toAbsolutePath(MtagsEnrichments.scala:130)
	at scala.meta.internal.mtags.MtagsEnrichments$XtensionStringMtags.toAbsolutePath(MtagsEnrichments.scala:187)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionString.toAbsolutePath(MetalsEnrichments.scala:773)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionString.toAbsolutePath(MetalsEnrichments.scala:770)
	at scala.meta.internal.metals.Folder$.apply(WorkspaceLspService.scala:1403)
	at scala.meta.metals.MetalsLanguageServer.$anonfun$initialize$4(MetalsLanguageServer.scala:148)
	at scala.collection.immutable.List.map(List.scala:247)
[...]

PR #1527, opened 2024-08-13
bugfix: Remove ui extensionKind

tgodzik:
Metals must run in the same machine.


tgodzik:
Should fix https://github.com/scalameta/metals-vscode/issues/1526

PR #1530, opened 2024-09-02
Resolve java home and java version using binary found on PATH.

megri:


packages/metals-languageclient/src/getJavaHome.ts
```diff
@@ -53,27 +53,55 @@ async function validateJavaVersion(
   return false;
 }
 
+function propertyValueOf(
+  input: string,
+  propertyName: string
+): string | undefined {
```
tgodzik, line 59: Could we wrap it in a try just in case?

packages/metals-languageclient/src/getJavaHome.ts
```diff
@@ -53,27 +53,55 @@ async function validateJavaVersion(
...
     );
-    const possibleJavaHome = path.dirname(path.dirname(realJavaPath));
-    const isValid = await validateJavaVersion(
-      possibleJavaHome,
-      javaVersion,
-      outputChannel
-    );
-    if (isValid) return possibleJavaHome;
-    else {
+    const cmdArgs = ["-XshowSettings:properties", "-version"];
```
tgodzik, line 80: Och, this is super useful!

packages/metals-languageclient/src/getJavaHome.ts
```diff
@@ -53,27 +53,55 @@ async function validateJavaVersion(
...
+      const cmdOutput = cmd.stderr as string;
+      const discoveredJavaHome = propertyValueOf(cmdOutput, "java.home");
+      const discoveredJavaVersion = propertyValueOf(
+        cmdOutput,
+        "java.specification.version"
+      );
+      const isValid =
+        discoveredJavaHome &&
+        discoveredJavaVersion &&
+        discoveredJavaVersion >= javaVersion;
```
tgodzik, line 92:
```suggestion
        parseInt(discoveredJavaVersion) >= parseInt(javaVersion);
```
otherwise 8 will be larger than 21

tgodzik (changes requested):
Great work! I just left a few comments

packages/metals-languageclient/src/getJavaHome.ts
megri: Nice catch!

packages/metals-languageclient/src/getJavaHome.ts
megri: Which part specifically?

tgodzik: I would say the whole method. I think it's pretty safe, but every string manipulation bites us at some point however good it looks :sweat_smile:

megri: Hm, if the output of the java process would be broken somehow we could end up trying to grab the second part of a split on `=`, where there is no second part. This would raise an index out of bounds-exception.

I'll add a `try/catch() { log(…); return }` statement around the split-part then :)

tgodzik: Thanks!

tgodzik (approved):
Let's merge it, thanks for working on it!

PR #1543, opened 2024-10-04
Add support for `build.mill` and `build.mill.scala`

lolgab:
Companion PR of https://github.com/scalameta/metals/pull/6752

### DIFF
--- a/packages/metals-vscode/package.json
+++ b/packages/metals-vscode/package.json
@@ -52,6 +52,8 @@
     "onLanguage:scala",
     "workspaceContains:build.sbt",
     "workspaceContains:build.sc",
+    "workspaceContains:build.mill",
+    "workspaceContains:build.mill.scala",
     "workspaceContains:project/build.properties",
     "workspaceContains:project.scala",
     "workspaceContains:.scala-build",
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Looks good! Let's wait for the metals PR first.

ckipp01:
Just a note for clients that also have logic to detect the root based off build files, they should also account for this.

lolgab:
> Just a note for clients that also have logic to detect the root based off build files, they should also account for this.

I'm not sure I understand. Can you elaborate?

ckipp01:
> > Just a note for clients that also have logic to detect the root based off build files, they should also account for this.
> 
> I'm not sure I understand. Can you elaborate?

Sure, and this is more of a note just for client maintainers, not necessarily for you. But in some clients, like Neovim and other terminal based editors, the idea of a "workspace" is sort of artificial. This causes the client to have to have some sort of logic to determine where the workspace root is. For example in `nvim-metals` we have some logic where we take a pre-set set of [patterns](https://github.com/scalameta/nvim-metals/blob/f861db9fda55939797ac1b05238c49b0dcdc3bdb/lua/metals/config.lua#L322) that we look for and then we have some logic about [traversing down](https://github.com/scalameta/nvim-metals/blob/main/lua/metals/rootdir.lua) to account for nested builds in different tools, etc. So even though this changes here in Metals, it's good, but this doesn't 100% mean it will actually work in some clients unless they also account for these filetypes.

Issue #1558, opened 2024-12-11
Unable to use local java version different from system-wide default for Metals


matteobilardi:
**Describe the bug**

In an environment with no access to the internet, where I have both VSCode, the metals extension, java 11 and java 17 already installed, and java 11 is the default system-wide java version, it is not possible to run metals with a Scala project that requires exactly java 11 both for running and building.

**To Reproduce** 
On ubuntu jammy
```
sudo apt install openjdk-17-jdk openjdk-11-jdk 
sudo update-alternatives --config java // Pick java-11-openjdk as the default java
````
Install vscode and the metals extension and open a scala project that requires java 11.
Disconnect from the internet.

**Expected behavior**

Currently, it is possible to set distinct java versions for metals and the project, e.g. via
```
metals.javaVersion: 17,
metals.javaHome: "/usr/lib/jvm/java-11-openjdk-amd64"
```

However, this only works if either java 17 is the system-wide java binary available in PATH or if there is an internet connection for coursier to use to download java. I would like to be able to tell metals to use the java version at `/usr/lib/jvm/java-17-openjdk-amd64/bin/java` to run itself, and the java home at `/usr/lib/jvm/java-11-openjdk-amd64`  to run the project, while at the same time not having java 17 being the default version of java available in PATH, and without having a connection to the internet. That is, I think that something like the following should be possible

```
metals.javaPath: "/usr/lib/jvm/java-17-openjdk-amd64/bin/java", // use this for metals
metals.javaHome: "/usr/lib/jvm/java-11-openjdk-amd64" // use this for the scala project
```

**Installation:**

- Operating system: Linux
- VSCode version: 1.94
- VSCode extension version: 1.43.0
- Metals version: 1.4.1

**Additional context**

**Search terms**

JAVA_HOME, javaVersion, javaHome, coursier, offline

PR #1559, opened 2024-12-12
improvement: Allow to override metals java home

tgodzik:
Peviously, if use had JAVA_HOME and java on PATH pointing to JDK 11, we would download JDK 17 using coursier. This however might not work if there is no internet and there was no way to work around that. Now, we allow users to specify an override for those cases.

Fixes https://github.com/scalameta/metals-vscode/issues/1558

### DIFF
--- a/packages/metals-languageclient/src/setupCoursier.ts
+++ b/packages/metals-languageclient/src/setupCoursier.ts
@@ -23,6 +23,7 @@ const coursierCommit = "11b428f35ca84a598ca30cce1c35ae4f375e5ee3";
 
 export async function setupCoursier(
   javaVersion: JavaVersion,
+  javaHomeOverride: string | undefined,
   coursierFetchPath: string,
   extensionPath: string,
   output: OutputChannel,
@@ -106,7 +107,13 @@ export async function setupCoursier(
   }
   output.appendLine(`Using coursier located at ${coursier}`);
 
-  var javaHome = await getJavaHome(javaVersion, output);
+  var javaHome: JavaHome | undefined;
+
+  if (javaHomeOverride) {
+    javaHome = await validateJavaVersion(javaHomeOverride, javaVersion, output);
+  } else {
+    javaHome = await getJavaHome(javaVersion, output);
+  }
 
   if (!javaHome && coursier) {
     output.appendLine(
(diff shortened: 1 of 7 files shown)


packages/metals-vscode/package.json
```diff
@@ -279,6 +279,11 @@
           "scope": "machine-overridable",
           "markdownDescription": "Optional path to the Java home directory that will be used for compiling the project.\n\nDefaults to JDK used by Metals's server (look: Java Version).\n\nThis Java version should be lower or equal to JDK version used by the Metals's server."
         },
+        "metals.metalsJavaHome": {
+          "type": "string",
+          "scope": "machine",
+          "markdownDescription": "Optional path to the Java home directory that will be used for the running Metals server.\n\nBy default Metals will try to infer it using the version specified in metals.javaVersion.\n\nThis Java version should be higher than 17."
```
kasiaMarek, line 285:
```suggestion
          "markdownDescription": "Optional path to the Java home directory that will be used for the running Metals server.\n\nBy default Metals will try to infer it using the version specified in metals.javaVersion.\n\nThis Java version should be higher of equal 17."
```
or `at least`?

packages/metals-languageclient/src/setupCoursier.ts
```diff
@@ -136,7 +143,7 @@ export async function setupCoursier(
   else
     throw Error(
       `Cannot resolve Java home or coursier, JAVA_HOME should exist with a version of at least ${javaVersion}.` +
-        `Alternatively, you can reduce the requirement using "metals.javaVersion" setting.`
+        `Alternatively, you can reduce the requirement using "metals.javaVersion" setting override the path using metals.metalsJavaHome setting.`
```
kasiaMarek, line 146:
```suggestion
        `Alternatively, you can reduce the requirement using "metals.javaVersion" setting and override the path using "metals.metalsJavaHome" setting.`
```

packages/metals-vscode/src/util.ts
```diff
@@ -115,6 +115,13 @@ export function getJavaVersionFromConfig() {
   return undefined;
 }
 
+export function getJavaVersionOverride(): string | undefined {
+  return workspace
+    .getConfiguration("metals")
+    .get<string>(UserConfiguration.MetalsJavaHome)
+    ?.trim();
```
kasiaMarek, line 122: Maybe if empty string you want to return `undefined`?

tgodzik: Done!

kasiaMarek (approved):

PR #1565, opened 2024-12-18
bugfix: Don't show panel if unable to parse release notes

tgodzik:
It seems previously it would always become visible right away (might be recent VS Code change). Now, we delay showing until the very end.

### DIFF
--- a/packages/metals-vscode/src/releaseNotesProvider.ts
+++ b/packages/metals-vscode/src/releaseNotesProvider.ts
@@ -66,6 +66,8 @@ async function showReleaseNotesImpl(
   // below are helper functions
 
   async function showPanel(version: string, releaseNotesUrl: string) {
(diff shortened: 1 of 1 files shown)


tgodzik:
@kasiaMarek could you take a look?

packages/metals-vscode/src/releaseNotesProvider.ts
```diff
@@ -66,6 +66,8 @@ async function showReleaseNotesImpl(
   // below are helper functions
 
   async function showPanel(version: string, releaseNotesUrl: string) {
+    const releaseNotes = await getReleaseNotesMarkdown(releaseNotesUrl);
```
kasiaMarek, line 69: Should we add some sort of timeout on this?

tgodzik: Added!

kasiaMarek (approved):

PR #1567, opened 2025-01-14
Add item schema to `metals.excludedPackages` setting

harpocrates:
This just ensures that the non-JSON view of settings in VS code will provide a GUI way of visualizing and updating the list of strings.


tgodzik (approved):
Thanks!

PR #1571, opened 2025-01-23
improvement: Add location when error message contains it

tgodzik:
Fixes https://github.com/scalameta/metals-feature-requests/issues/417

### DIFF
--- a/packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
+++ b/packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
@@ -10,6 +10,7 @@ import {
   TestSuiteResult,
 } from "./types";
 import { gatherTestItems } from "./util";
+import { existsSync } from "fs";
 
 /**
  * Analyze results from TestRun and pass inform Test Controller about them.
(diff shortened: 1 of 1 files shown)


packages/metals-vscode/src/testExplorer/analyzeTestRun.ts
```diff
@@ -162,5 +163,21 @@ function extractErrorMessages(failed: Failed[]): vscode.TestMessage[] {
...
 
 function toTestMessage(error: string): vscode.TestMessage {
-  return { message: ansicolor.strip(error) };
+  const message = ansicolor.strip(error);
+  const regex = /[\\\/]([^\s]*):(\d+)/;
+  const match = regex.exec(message);
+
+  if (match) {
+    const [, file, line] = match;
+    const filePath = vscode.Uri.file(file).fsPath;
```
kasiaMarek, line 172: `file` will just match the file name, not the full path, right? So `vscode.Uri.file(file).fsPath` will create a full path out of it?

tgodzik: Uri.file creates an actual uri out if a path, fsPath will print a proper file system path with windows issues handled. We might be able to just use file, but it feels like this is safer

tgodzik:
I also tried adding the location to test results in https://github.com/scalacenter/scala-debug-adapter/compare/main...tgodzik:scala-debug-adapter:add-loc?expand=1

But the biggest issue with that approach is that we only get the file name, which we then need to search for on our own. This might also be duplicated in your code base. It's easier for now to just do the regex

tgodzik:
superceeded by https://github.com/scalameta/metals-vscode/pull/1574 but should work in most test frameworks now

PR #1591, opened 2025-02-27
refactor: Remove separate packages as we never really use them

tgodzik:
Using separate interfaces and packages doesn't really work great and we've had multiple issues with that. On top of it we had to have custom interfaces which can easily get out of sync with language server protocol library.

Removing them will be the next step.


src/checkServerVersion.ts
```diff
@@ -34,7 +34,11 @@ export async function checkServerVersion({
     serverVersionInfo(config);
   const isOutdated = (() => {
     try {
-      return semver.lt(serverVersion, latestServerVersion);
+      if (serverVersion && latestServerVersion) {
```
tgodzik, line 37: I had to rework some thing since language client didn't have linting on.

kasiaMarek (approved):

PR #1597, opened 2025-03-07
fix run main classes from dependnecies in non-debug mode

cvogt:
Trying to run the main class from one of the dependencies succeeds in debug mode, but fails in non-debug mode. This makes it work.

The reason is because discover-jvm-run-command fails for main classes in dependencies. (Also see related https://github.com/scalameta/metals/pull/7284 which improves the error message for that situation)


tgodzik (approved):
Makes sense! Thanks!

PR #1601, opened 2025-03-11
improvement: suppress unhelpful debug error

cvogt:
(Stacked on top of https://github.com/scalameta/metals-vscode/pull/1607)

Status Quo: When trying to run or debug while the workspace has compile errors, metals emits an error message "Cannot run class, since the workspace has errors.". This error is thrown in the lsp client, passes through metals-vscode and is caught by vscode itself, and turned into this popup:

Screenshot 2025-03-11 144447

The shown link to launch.json is not very helpful and actually misleading. I've repeatedly seen people click it in the moment even though logically it is not the right place to look.

This PR catches the error and replaces the error message with a more helpful one, that links to the vscode problem pane:

Screenshot 2025-03-11 144035

### DIFF
--- a/src/debugger/scalaDebugger.ts
+++ b/src/debugger/scalaDebugger.ts
@@ -1,4 +1,5 @@
 import * as vscode from "vscode";
+import { ResponseError } from "vscode-languageserver-protocol";
 import {
   commands,
   DebugConfiguration,
@@ -20,6 +21,8 @@ import {
 import { ServerCommands } from "../interfaces/ServerCommands";
 
 const configurationType = "scala";
+
+const workspaceHadErrors = 543; // custom code we picked to represent this
 
 export interface DebugSession {
   name: string;
@@ -134,15 +137,40 @@ export async function start(
   }
 }
 
+function handleCompileError(error: ResponseError) {
+  vscode.window
+    .showErrorMessage(error.message, { modal: true }, "View Problems")
+    .then((action) => {
+      if (action === "View Problems") {
+        vscode.commands.executeCommand("workbench.action.problems.focus");
+      }
+    });
+}
+
 async function debug(
   noDebug: boolean,
   debugParams: DebugDiscoveryParams | ScalaCodeLensesParams
 ): Promise<boolean> {
   await commands.executeCommand("workbench.action.files.save");
-  const response = await vscode.commands.executeCommand<DebugSession>(
-    ServerCommands.DebugAdapterStart,
-    debugParams
-  );
+
+  let response: DebugSession | undefined;
+  try {
+    response = await vscode.commands.executeCommand<DebugSession>(
+      ServerCommands.DebugAdapterStart,
+      debugParams
+    );
+  } catch (error) {
+    if (error instanceof ResponseError && error.code === workspaceHadErrors) {
+      handleCompileError(error);
+      // Returning false suppresses vscode's error popup wrongly suggesting to "Open launch.json".
+      // We put the responsibility to visibly show error messages in the hands
+      // of the server for compile errors.
+      // Also an error in the executeCommand still shows up in metals.log
+      return false;
+    }
+
+    throw error;
+  }
 
   if (response === undefined) {
     return false;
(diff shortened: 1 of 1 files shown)


packages/metals-vscode/src/debugger/scalaDebugger.ts
```diff
@@ -134,15 +134,43 @@ export async function start(
   }
 }
 
+function handleCompileError(error: unknown) {
+  const errorMessage = error instanceof Error ? error.message : String(error);
+  if (
+    errorMessage.includes("Cannot run class, since the workspace has errors.")
```
tgodzik, line 140: If we ever change the message this will stop working. Maybe instead we should use ShowMessageRequest and then we can use client.executeCommand method to show the error modal if supported. 

Most of the work would need to be done on the Metals side, specifically catch that exception and show the message. We could potentially have some option for each error such as 'show doctor' in some other cases or just showing the message if there is nothing to do.

On the VS Code side we would only need to define a client command that would invoke `vscode.commands.executeCommand("workbench.action.problems.focus");` later on.

What do you think?

tgodzik:
And obviously, thanks for the contributions!

packages/metals-vscode/src/debugger/scalaDebugger.ts
cvogt-bwater: Makes sense. Should we leave the ``return new vscode.DebugAdapterExecutable("", []);` in place to suppress vscode's unhelpful error message, which would otherwise race with or follow the one we're emitting from metals?

tgodzik: It might make sense, yeah. But we should do it after the fix in metals

tgodzik: You might need to rebase, but it should not cause any conflicts. I merged the two packages since it was more trouble than it's worth.

cvogt-bwater: sounds good. Will make the change in metals. Not sure what you are referring to with the two packages and more trouble.

cvogt:
Ok, I changed this PR and limited it to only suppressing the unhelpful error message. (Ignore the changes from the commit stemming from yet to be merged https://github.com/scalameta/metals-vscode/pull/1607)

Makes sense to release together with this Metals PR emitting a more helpful error message as you suggested: https://github.com/scalameta/metals/pull/7321

The client command you suggested already existed :).

tgodzik (commented):
The change looks good! Let's wait until the metals PR is merged.

cvogt2:
without this change, metals-vscode 1.48 now shows two popups. one emitted from metals and one less helpful one from vscode itself. This change fixes that.

Is it a huge effort to do another release of metals-vscode right away?

tgodzik:
> Is it a huge effort to do another release of metals-vscode right away?

Not at all, just wondering if now we don't suppress too much. We will only show the message if there are compile errors, but all other errors (such as not class found) will not show up, right?

cvogt2:
@tgodzik So vscode's default message will suggest to open launch.json is almost never the right suggestion. I like the idea of putting the responsibility to report the right error to the server side. But I am with you that a `catch(error)` is quite broad. Let me take a look if I can narrow that somehow and if it is not what we expect maybe show a different popup or log it?

tgodzik:
> Let me take a look if I can narrow that somehow and if it is not what we expect maybe show a different popup or log it?

Ideally, there is still a number of errors that actually would need opening launch.json (things that were wrongly defined there). But for anything not coming from launch.json it doesn't make sense to show it.

cvogt:
I narrowed the catching and otherwise rethrow the error. That limits the scope of where vscode's default error message is suppressed. Either way, the error does also show up in metal's log:

Screenshot 2025-04-03 123047

src/debugger/scalaDebugger.ts
```diff
@@ -139,10 +140,27 @@ async function debug(
...
+
+  let response: DebugSession | undefined;
+  try {
+    response = await vscode.commands.executeCommand<DebugSession>(
+      ServerCommands.DebugAdapterStart,
+      debugParams
+    );
+  } catch (error) {
+    if (
+      error instanceof ResponseError &&
```
tgodzik, line 152: We now do not get any messages about Class not found or wrong target used. We should probably add those first so that users still get all the relevant information. Basically, everything in https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L848C27-L848C50 should show a relevant message.

Maybe we could just catch those errors and show the message? Or show the message where they produced?

tgodzik (changes requested):
Sorry for being so late! I was meaning to test out the change to see some other cases, but I didn't have the time until now.

src/debugger/scalaDebugger.ts
cvogt2: Thanks for testing! Are you suggesting to catch those on the metals server side and send a showMessage to the client? Or do you have an idea how to catch and distinguish them on the client side without tying the logic to the wording of the error message, which we wanted to avoid.

tgodzik: Catching it on the metals side let's us actually control what is show to the user, so it's probably better since we already do that with the errors in workspace message. We can have custom actions for each (or default to opening launch.json when indeed that is broken)

cvogt2: Ok, agreed. In that case, how about we cut a new metals-vscode now without this change? https://github.com/scalameta/metals-vscode/pull/1607 is much more important as it fixes wrong behavior.

And I'll work on a new patch to metals server we can merge before #1601, which just is a UI issue. And then those can go in for the next metals server release happens.

tgodzik: > Ok, agreed. In that case, how about we cut a new metals-vscode now without this change? https://github.com/scalameta/metals-vscode/pull/1607 is much more important as it fixes wrong behavior.

Sure!

cvogt: Ok, so I looked futher into this. I think mostly for other messages than compile errors suggesting to open launch.json is actually good. Trying to rebuild this suggestion on the server side is finicky. So rather than catching these other exception and then just trying to mimick vscode's default we suppressed, how about we go a different route:

json rpc allows sending custom return codes. let's pick one to represent compile errors. We only suppress vscode's default message for those exactly, but show it in all other cases. I tried it and it works well. this also allows moving the emission of the error message back to the client as I had it originally, which makes it look the same as the default as a modal not a notification. I think this is good. I'll make a PR against metal and update this PR and we can discuss further.

cvogt: ok, force pushed an update. this will work together with https://github.com/scalameta/metals/pull/7423 . however either can be merged independently as https://github.com/scalameta/metals/pull/7423 just goes back to prior behavior which solves the double message people see right now. And #1601 doesn't do anything until #7423 is merged.

As a side note, this uses a custom return code for the jsonrpc response which is allowed according to spec as I understand.

tgodzik (approved):
LGTM

PR #1645, opened 2025-04-17
feat: setting for by-name parameter inlay hint

harpocrates:


.vscode/settings.json
```diff
@@ -11,4 +11,4 @@
   "files.watcherExclude": {
     "**/target": true
   }
-}
+}
```
tgodzik, line 14: The formatter is not happy about the missing ending newlines. You can run `yarn format` to fix it.

tgodzik (changes requested):
Thanks for raising this!

.vscode/settings.json
harpocrates: Yeah, I did but I have a git hook that undid the work. 😄

tgodzik (approved):
LGTM!

PR #1649, opened 2025-04-21
config metals.testEnvironmentVariables support

amitmiran137:
sometimes in your test suite you depend on specific environment variable to be passed

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -774,7 +774,11 @@ function launchMetals(
       // see https://github.com/scalameta/metals-vscode/discussions/1244
       const istTestManagerDisabled =
         !vscodeTextExplorer || getTestUI() === "Code Lenses";
-      const testManager = createTestManager(client, istTestManagerDisabled, getTestEnvVars());
+      const testManager = createTestManager(
+        client,
+        istTestManagerDisabled, 
+        getTestEnvVars()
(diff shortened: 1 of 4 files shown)


tgodzik:
Using env or envFile field in launch.json doesn't help? We could potentially pick up one from the workspace automatically.

amitmiran137:
> Using env or envFile field in launch.json doesn't help? We could potentially pick up one from the workspace automatically.

when you have hundrends of different Spec files you don't want to define a launch.json for each spec

src/extension.ts
```diff
@@ -761,11 +761,24 @@ function launchMetals(
...
+
       // vscodeTextExplorer can be undefined e.g. for eclipse theia,
       // see https://github.com/scalameta/metals-vscode/discussions/1244
       const istTestManagerDisabled =
         !vscodeTextExplorer || getTestUI() === "Code Lenses";
-      const testManager = createTestManager(client, istTestManagerDisabled);
+      const testManager = createTestManager(
+        client,
+        istTestManagerDisabled,
+        getTestEnvVars()
```
tgodzik, line 780: Looks like maybe we should make it lazy actually, since now you need to restart the plugin. Could you take a look? I haven't noticed the issue before

PR #1658, opened 2025-04-26
improvement : Add defaultShell as a User Option

ajafri2001:


package.json
```diff
@@ -231,6 +231,12 @@
           "scope": "window",
           "markdownDescription": "Enable best effort mode for Metals in Scala 3. If enabled, Metals will try to provide most up to date information from the codebase even if it's not compiling."
         },
+        "metals.defaultShell": {
+          "type": "string",
+          "default": "",
```
tgodzik, line 236: Can we remove the default? I don't think we need it.

tgodzik (approved):

PR #1688, opened 2025-05-25
feat: expose API to get language client

harpocrates:
This makes it possible to get a reference to the `LanguageClient` from other VS code plugins, which opens up possibilities for tighter programmatic integration (ex. confirm project is correctly imported, additional ad-hoc LSP requests).


tgodzik (approved):
LGTM! Looks like github had some issues, but rerunning the test helped

PR #1699, opened 2025-06-05
feat: auto add imports on paste

kasiaMarek:
requires: https://github.com/scalameta/metals/pull/7542

### DIFF
--- a/src/metalsCopyPaste.ts
+++ b/src/metalsCopyPaste.ts
@@ -1,8 +1,6 @@
 import { window, env, ExtensionContext, commands, Position } from "vscode";
 import { LanguageClient } from "vscode-languageclient/node";
-import {
-  ExecuteCommandRequest
-} from "vscode-languageserver-protocol";
+import { ExecuteCommandRequest } from "vscode-languageserver-protocol";
 
 /**
  * Copies the current selection to the clipboard.
@@ -54,7 +52,9 @@ export async function pasteSelection(
   }
 
   const clipboardText = await env.clipboard.readText();
-  if (!clipboardText) return;
+  if (!clipboardText) {
+    return;
+  }
 
   // Get current selection for the paste range
   const selection = editor.selection;
@@ -77,7 +77,9 @@ export async function pasteSelection(
     !originDocumentUri ||
     originStartLine === undefined ||
     originStartCharacter === undefined
-  ) return;
+  ) {
+    return;
+  }
 
   // Calculate the new end position after pasting
   const lines = clipboardText.split("\n");


src/metalsCopyPaste.ts
```diff
@@ -0,0 +1,155 @@
...
+  const editor = window.activeTextEditor;
+
+  if (!editor) {
+    window.showErrorMessage("No active editor found");
+    return;
+  }
+
+  const clipboardText = await env.clipboard.readText();
+  if (!clipboardText) {
+    window.showErrorMessage("Clipboard is empty");
```
tgodzik, line 59: I think we shouldn't show an error, the current paste, doesn't do that.

src/metalsCopyPaste.ts
```diff
@@ -0,0 +1,155 @@
...
+  const originStartCharacter = context.workspaceState.get(
+    "copyStartCharacter"
+  ) as number;
+
+  if (
+    !originDocumentUri ||
+    originStartLine === undefined ||
+    originStartCharacter === undefined
+  ) {
+    window.showErrorMessage("No previous copy operation found");
```
tgodzik, line 77: We should just do a normal paste in this case

src/metalsCopyPaste.ts
```diff
@@ -0,0 +1,155 @@
...
+  const lines = clipboardText.split("\n");
+  const newEndPosition = new Position(
+    selection.start.line + lines.length - 1,
+    lines.length === 1
+      ? selection.start.character + clipboardText.length
+      : lines.pop()?.length || 0
+  );
+
+  // It will suplicate with the other didChange notification,
+  // but I want to make sure the content or the server is updated.
```
tgodzik, line 99:
```suggestion
  // but I want to make sure the content on the server is updated.
```
is that really needed? We could potentially, just send the full string in Metals Paste

kasiaMarek:
> would like to take a look at the issue?

Yeah, I can take a look.

> I don't even see the did paste being sent 🤔 I hoped https://github.com/tgodzik/metals/pull/new/fix-did-paste would fix things.

It will send an empty result back. The pasted code is added to the buffer on the VSCode side. Metals only sends the imports that are missing, so any lag is only visible with imports, and if the server is down, copy-paste should still work.

tgodzik:
I meant that the metals-paste command is not being sent from what I noticed, I wanted to fix the issue by manually invoking the RangeFormattingProvider and joining the edits.

kasiaMarek:
>  I hoped https://github.com/tgodzik/metals/pull/new/fix-did-paste would fix things. 

That does fix the issue for me. If you cannot see the `metals-did-paste` command, maybe for some reason your keyboard shortcuts aren't bound the way they should be?

tgodzik:
Hmm... I was pasting though, and if I just used normal paste the onRangeFormatting should have worked 🤔 

I will double check after scala days and add the test to the PR.

PR #1773, opened 2025-11-04
feature: Automatically create config for main methods that were run

tgodzik:


tgodzik:
This needs a fix in Metals to also accept a full build identifier

PR #1811, opened 2026-01-03
Fix: Adjusting logging when reporting coursier location and expose forceCoursierJar to settings (#1810)

### LINKED ISSUE #1810: coursier location is not correctly reported in logs when fallback jar is used
**Describe the bug**
When the plugin uses the fallback jar, it incorrectly reports the location being used as undefined due to the location of this log line:
https://github.com/scalameta/metals-vscode/blob/e317f62f34be6adfcdc64838d46ef29fda62d5e7/src/setupCoursier.ts#L118C28-L118C37

**To Reproduce** 

1) Force fallback to jar (unclear if this is possible via setting, but in my corporate environment it happens automatically due to github and maven timing out)
2) Eventually it will reach the log line:
"Using coursier located at: undefined"

**Expected behavior**

If using downloaded coursier, should show the path to cs.
If using jar based coursier, should show the path to the fallback jar and java home

**Installation:**

- Operating system: Windows
- VSCode version: 1.107.1
- VSCode extension version: 1.60.0
- Metals version: 1.6.4

jbburns:
Fixes #1810

### DIFF
--- a/package.json
+++ b/package.json
@@ -228,7 +228,7 @@
           "type": "boolean",
           "default": false,
           "scope": "machine",
-          "markdownDescription": "Do not attempt to download/use Coursier executable and instead use fallback JAR (Offers capability to override mirrors & credentials in corporate envionrments.)"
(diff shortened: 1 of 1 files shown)


tgodzik (changes requested):

tgodzik (changes requested):

tgodzik (approved):
LGTM!

PR #1843, opened 2026-02-05
Allow to use Metals on all BSD systems without hacks.

alexdupre:
Changes:
- identify all BSD systems instead of just "openbsd"
- drop the recognized, but unsupported, option UseZGC in all direct java commands (coursier, metals server, bloop)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -293,9 +293,26 @@ async function fetchAndLaunchMetals(
     coursierMirrorFilePath: coursierMirror,
   });
 
+  /**
+   * GraalVM for JDK 17-20 prints additional warnings that breaks things.
+   * Looks like JDK on Window is also affected.
+   */
+  const skipZGC =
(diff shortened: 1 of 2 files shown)


tgodzik:
> Nevermind, the patch to metals-vscode is ok.
> 
> The other fix I have to do is on bloop, because when no jvm properties are passed, bloop adds UseZGC, too. I don't think this can be fixed directly in metals-vscode, either the user has to manually override the `metals.bloopJvmProperties` or a similar conditional needs to be implemented in bloop. The former is ok as a temporary workaround (I tested it and worked), but the latter is a better/definitive solution that will make metals-vscode working OOTB.

Yeah, the. assessment seems correct in this case. Thanks for contributing!

alexdupre:
Thanks @tgodzik

For reference: https://github.com/scalacenter/bloop/pull/2874

Issue #1882, opened 2026-03-10
chore: add syntax highlighting for Twirl templates


AsadHumayun:
With the recent merge of #1872 and #1871 it seems that this is a pending / in a to-do state. After doing some research, I was able to find this repository https://github.com/guillaumebort/play2-sublimetext2 made by the person who created Play framework that has TextMate grammars for Twirl, both in XML and JSON formats.

I have emailed the author of the repo but am awaiting a response - there is no licence attached to the repository so not sure how to go about this but it is possible to apply this grammar locally and it does work.

It would be really helpful if we could get this merged but I completely understand that it isn't exactly easy to write an entire grammar. 

Apologies if I created this in the wrong place - this did not feel like a full-blown new feature request as it seems already in WIP.

tgodzik:
Good idea! It should be quite simple to attach a new working syntax, while indeed a new one might quite complex. One could try generating it with LLMs, but setting that out with testing and making sure it all  works correctly is also a bit complex.

If you get a response from the author and we can embed the syntax then you can for sure open up a PR with it. Should be a quick approval from me

AsadHumayun:
Yeah good idea, it does sound like it is the best solution from what I've seen. While we'll probably still need a TextMate grammar for fallback colouring etc and if we for some reason can't compile the project, yeah this would probably be the best way to go about doing it. 

I have taken a quick look at the PR that initially introduced Twirl template support to Metals, and that at the moment skips sending semantic tokens entirely if it's a Twirl file. I will take a look over the weekend and see if there is something that could be done with `TwirlParser` to add this support...

PR #1909, opened 2026-04-04
feat: add TextMate grammar syntax for Twirl HTML files

AsadHumayun:
This is an initial PR asking for initial feedback on a TextMate grammar that I have been working on for Twirl's HTML files.

It is important to know that there are definitely some edge cases to this Twirl syntax. I have tried my absolute best with whatever knowledge I currently have about regex, but it seems that you can't get more-complex regexes working correctly.

Some of these edge cases can be:
- Note that the second paramter list is not coloured or picked up as scala source
Image

- There are also some issues with an example such as this below:
image

- ~~`<`, `>` chars always show up as red - not exactly sure why this is happening, my guess is that it it doesn't get picked up as scala source, and then goes through to the HTML grammar, which expects a tag and doesn't find that, so shows up as an illegal char.~~ (resolved)
 This is an example of what I have tried to remedy the issue:
```json
		"ignore-only-gtlt": {
			"name": "string.other",
			"match": "(.*)(<|>)(.*)",
			"captures": {
				"1": { "patterns": [{ "include": "source.scala" }] },
				"2": { "name": "string.other" },
				"3": { "patterns": [{ "include": "source.scala" }] }
			}
		},
```
The main idea here is that we catch anything before and after `<` or `>`, and then we manually catch `<`, `>` and put them as strings or something, but I have found that this generally causes more harm than good, see:
image

> There are also some concerns around the performance of the regex `.*(<|>).*` - see [[Details of the Cloudflare outage on July 2, 2019](https://blog.cloudflare.com/details-of-the-cloudflare-outage-on-july-2-2019/)] in which a similar regex, `.*=.*` led to catastrophic backtracking and heavy CPU utilisation.

It also seems to mess up HTML tags -
image

I think it's probably best to leave to just leave this and let semantic tokens pick up the slack here.

- As you can see in the below screenshot, any inline scala is not recognised inside HTML tags. If I wanted to correct this, I would have effectively had to fork the HTML grammar and add my own patterns in there which definitely isn't a good idea (adding lots of maintenance overhead, complexity, etc).

Image

Perhaps one of the worser-instances of this is as follows:
image

The idea here is that the implementation of semantic highlighting is going to address a lot of these issues, and that the textmate grammar added here is enough for _most_ cases and should serve fine as a temporary colouring until the language server (in this instance, Metals) can jump in and resolve the tokens and traverse the AST. I am currently working on an implementation of these in my own repository in https://github.com/AsadHumayun/twirl-semantic-tokens/blob/main/src/main/scala/traverser/Traverser.scala. I am working on it, don't expect it soon, but I am making some progress on adding these. 

This PR brings in https://github.com/AsadHumayun/twirl-grammars/commit/fb7f6108067d048cb6c8b4937fec5024e7635cab of the grammar. (not as a submodule or anything, just a note for myself to where this was taken from)

Partially addresses #1882.

## Summary by CodeRabbit

* **New Features**
  * Enhanced Twirl language support with broader file recognition and improved syntax highlighting
  * Added smart bracket/quote auto-closing, auto-indentation, and bracket colorization
  * Improved folding for Twirl block comments and template sections
  * Automatic comment formatting and on-enter behavior for Twirl block comments

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,19 +162,16 @@
           "light": "icons/scalameta-logo.png"
         }
       },
-      {
-        "id": "twirl-html",
-        "aliases": [
-          "Twirl HTML"
-        ],
-        "extensions": [
-          ".scala.html"
-        ],
+			{
+				"id": "twirl-html",
+				"extensions": ["*.scala.html"],
+				"aliases": ["Twirl HTML"],
+				"configuration": "./language-configuration.json",
         "icon": {
           "dark": "icons/file_type_play.svg",
           "light": "icons/file_type_play.svg"
         }
-      },
+			},
       {
         "id": "twirl-xml",
         "aliases": [
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
       },
       {
         "language": "twirl-html",
-        "scopeName": "text.html.twirl",
+        "scopeName": "source.twirl",
         "path": "./syntaxes/twirl-html.json",
         "embeddedLanguages": {
           "text.html.basic": "html"
--- a/package.json
+++ b/package.json
@@ -162,16 +162,20 @@
           "light": "icons/scalameta-logo.png"
         }
       },
-			{
-				"id": "twirl-html",
-				"extensions": ["*.scala.html"],
-				"aliases": ["Twirl HTML"],
-				"configuration": "./syntaxes/language-configurations/twirl-html.language-configuration.json",
+      {
+        "id": "twirl-html",
+        "extensions": [
+          "*.scala.html"
+        ],
+        "aliases": [
+          "Twirl HTML"
+        ],
+        "configuration": "./syntaxes/language-configurations/twirl-html.language-configuration.json",
         "icon": {
           "dark": "icons/file_type_play.svg",
           "light": "icons/file_type_play.svg"
         }
-			},
+      },
       {
         "id": "twirl-xml",
         "aliases": [
--- a/package.json
+++ b/package.json
@@ -251,7 +251,9 @@
         "language": "twirl-html",
         "scopeName": "source.twirl",
         "path": "./syntaxes/twirl-html.json",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -248,7 +249,7 @@
       },
       {
         "language": "twirl-html",
-        "scopeName": "text.html.twirl",
+        "scopeName": "source.twirl",
         "path": "./syntaxes/twirl-html.json",
```
RedCMD: @AsadHumayun
should use `.tmLanguage.json` file extension for all TextMate grammar files
```suggestion
        "path": "./syntaxes/twirl-html.tmLanguage.json",
```

tgodzik (commented):
I think any improvement in this case is useful even if we don't handle all he edge cases. Tested the solution and seems to work in some basic cases I have. 

Since it's really hard ro review regexes I am happy to merge it as soon as you yourself are happy with it. The code looks mostly clear aside from the more regexes

syntaxes/twirl-html.json
```diff
@@ -1,10 +1,192 @@
...
+      "patterns": [{ "include": "$self" }]
+    },
+    "scala-template-block": {
+      "name": "meta.embedded.twirl.source.templateBlock",
+      "begin": "(@)([\\w\\.]*)\\s*\\=\\s*\\{",
+      "end": "\\}",
+      "comment": "Attempts to match a template block in assignment, so @ something = { ... }",
+      "beginCaptures": {
+        "1": { "patterns": [{ "include": "#symbols-at" }] },
+        "2": { "patterns": [{ "name": "keyword.control" }] }
```
RedCMD: @AsadHumayun
`name` doesn't work when nested byitself in a `patterns` array
```suggestion
        "2": { "name": "keyword.control" }
```
image

tgodzik (approved):
Let's try it out!

Issue #1921, opened 2026-04-13
`JAVA_TOOL_OPTIONS` truncated at `=` when using "Run" code lens (via Bloop)


ChrisMuki:
**Describe the bug**

When running a main class via the Metals "Run" code lens (using Bloop as build server), environment variables whose values contain `=` characters are truncated at the first `=` in the value.

For example, `JAVA_TOOL_OPTIONS` set to:
```
-Dhttp.proxyHost=cias.example.com -Dhttp.proxyPort=8080
```
is received by the spawned JVM process as just:
```
-Dhttp.proxyHost
```

This causes proxy settings, truststore paths, and any other `-D` system properties passed via `JAVA_TOOL_OPTIONS` to be silently broken. Running the same class via `sbt run` in a terminal works correctly — the full value is preserved.

The root cause is in [`scalaDebugger.ts` (`runMain`)](https://github.com/scalameta/metals-vscode/blob/main/src/debugger/scalaDebugger.ts):

```typescript
const [key, value] = envKeyValue.split("=");
```

`split("=")` splits on **every** `=`. Destructuring only takes the first two elements, so everything after the second `=` is discarded.

Suggested fix — split only on the first `=`:

```typescript
const eqIdx = envKeyValue.indexOf("=");
const key = envKeyValue.substring(0, eqIdx);
const value = envKeyValue.substring(eqIdx + 1);
```

**To Reproduce**

1. Set `JAVA_TOOL_OPTIONS="-Dhttp.proxyHost=example.com -Dhttp.proxyPort=8080"` in the container environment (e.g. `/etc/environment`)
2. Create a Scala main class that prints `sys.env.get("JAVA_TOOL_OPTIONS")` and `sys.props.get("http.proxyHost")`
3. Click "Run" code lens in Metals (with Bloop as build server)
4. Observe the output shows truncated values (`-Dhttp.proxyHost` instead of the full value)

**Expected behavior**

The full value of `JAVA_TOOL_OPTIONS` (and any other env var containing `=`) should be passed to the spawned JVM process unchanged.

**Screenshots**

N/A

**Installation:**

- Operating system: Linux (Ubuntu 22.04, devcontainer)
- VSCode version: 1.99.0
- VSCode extension version: 1.65.0 (also reproduced on 1.59.0)
- Metals version: 1.5.1

**Additional context**

[...]

PR #1922, opened 2026-04-13
fix: preserve env var values containing '=' in runMain

ChrisMuki:
Addressing Issue #1921 

split("=") splits on every '=' character, causing values like JAVA_TOOL_OPTIONS="-Dhttp.proxyHost=example.com -Dhttp.proxyPort=8080" to be truncated at the first '=' in the value.

Use indexOf("=") to split only on the first '=' so the full value is preserved. Extract parseEnvVariables into its own module for testability.

## Summary by CodeRabbit

* **Refactor**
  * Improved environment variable parsing in the debugger by extracting parsing logic into a dedicated helper, enhancing maintainability and code reusability.

* **Tests**
  * Added comprehensive unit tests covering standard parsing, values with special characters, and edge cases for environment variable handling.

### DIFF
--- a/src/debugger/parseEnvVariables.ts
+++ b/src/debugger/parseEnvVariables.ts
@@ -3,15 +3,17 @@
  * Splits only on the first '=' so that values containing '=' are preserved.
  */
 export function parseEnvVariables(
-    environmentVariables: string[],
+    environmentVariables: string[]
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM! Thanks!

PR #1947, opened 2026-05-05
improvement: Add a new setting for performance specific options

tgodzik:
## Release Notes

* **New Features**
  * Added `metals.initializationSettings` configuration option in VS Code settings, allowing customization of Metals server initialization behavior. Changes to this setting will prompt a reload of the extension.


tgodzik:
We should also add a check that we are using v2 when those options are set. I will experiment further with it

PR #1953, opened 2026-05-10
fix(extension): remove isActive filter during extension initialization

cheleb:
Remove the filter that checks for `e.isActive` when identifying active client extensions. This prevents a race condition where extensions might not yet be marked as active during the initialization phase, ensuring they are correctly identified and managed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed extension selection to properly include all allowlisted extensions instead of only those currently marked as active.

[Review Change Stack](https://app.coderabbit.ai/change-stack/scalameta/metals-vscode/pull/1953)


tgodzik (approved):
Thanks!

PR #1963, opened 2026-05-18
Add Metals custom server launcher

mahib-stripe:
### Summary

Adds an external server command option `metals.ServerCommand` of type string.

When serverCommand is set, the Metals VSCode extension will use a custom launcher script that skips coursier/download and starts that command as the Metals-compatible stdio server, while keeping the normal VSCode language client wiring.

### Motivation

At Stripe, we want to point VSCode at a custom wrapper for launching metals so we can integrate with existing lsp-proxy tooling for free metrics and LSP consolidation (i.e. only pick one LSP for the task at hand).

Elsewhere, custom launchers are used but it would be nice to upstream custom launcher scripts as a solution to handle other cases of integration.

### Test plan
- yarn install passed.
- yarn compile passed.
- ./node_modules/.bin/ts-mocha src/test/unit/getServerOptions.test.ts passed.
- yarn format-check passed.
- yarn eslint . passed with existing warnings only.
- added a focused unit test for the new server-option helper 

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration, metals.customServerLauncher, allowing you to specify a custom command to launch the Metals language server.

* **Tests**
  * Added unit tests validating the custom server launcher configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/scalameta/metals-vscode/pull/1963?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/package.json
+++ b/package.json
@@ -377,6 +377,12 @@
           "default": "1.6.7",
           "markdownDescription": "The version of the Metals server artifact. Requires reloading the window.  \n\n**VS Code extension version is guaranteed to work only with the default version, change if you know what you're doing**"
         },
+        "metals.customServerLauncher": {
+          "type": "string",
+          "default": "",
+          "scope": "machine-overridable",
+          "markdownDescription": "Optional command to launch a Metals-compatible language server over standard input/output. When set, this completely overrides the default Metals server launch flow, including `metals.serverVersion`, `metals.serverProperties`, and `metals.customRepositories`."
+        },
         "metals.serverProperties": {
           "type": "array",
           "items": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -92,10 +92,11 @@ import { setupCoursier } from "./setupCoursier";
 import { getJavaOptions } from "./getJavaOptions";
 import { getJavaConfig, JavaConfig } from "./getJavaConfig";
 import { fetchMetals } from "./fetchMetals";
-import { getServerOptions } from "./getServerOptions";
+import { getCustomServerOptions, getServerOptions } from "./getServerOptions";
 import { isSupportedLanguage } from "./isSupportedLanguage";
 import { readRequiredVmOptions } from "./readRequiredVmOptions";
 import { MetalsInitializationOptions } from "./interfaces/MetalsInitializationOptions";
+import { ServerOptions } from "./interfaces/ServerOptions";
 import { restartServer } from "./commands/restartServer";
 import { ServerCommands } from "./interfaces/ServerCommands";
 import { ClientCommands } from "./interfaces/ClientCommands";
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -567,9 +592,6 @@ async function launchMetals(
   javaConfig: JavaConfig,
   serverVersion: string,
 ) {
-  // Make editing Scala docstrings slightly nicer.
-  enableScaladocIndentation();
```
olafurpg, line 571: Intentional?

package.json
```diff
@@ -377,6 +377,30 @@
           "default": "1.6.7",
           "markdownDescription": "The version of the Metals server artifact. Requires reloading the window.  \n\n**VS Code extension version is guaranteed to work only with the default version, change if you know what you're doing**"
         },
+        "metals.serverCommand": {
```
olafurpg: I would recommend we make this a single property of type object so the settings look like 

```
"metals.customServerLauncher": {
  "command": STRING,
  "args": ARRAY,
  "env": ...
}
```

It's unclear what `env` buys you, if it's hardcoded then you might as well hardcode it in your command. It would make more sense to forward the environment of the extension host. Similar thoughts on `args`, if you're hardcoding it via settings then you could simplify this down to

```
"metals.customServerLauncher": STRING
```

olafurpg (changes requested):
It's worth adding this setting but we can probably simplify it a bit. We already have `metals.server*` settings and it's worth making it clear that this totally overrides that.

src/extension.ts
mahib-stripe: yes, [`launchMetals()` ](https://github.com/scalameta/metals-vscode/blob/bb0d98a035f13c9f84ec09bb5231d00cf3b668ef/src/extension.ts#L571) already does this, so if we use a custom launch it should use the same editor behavior

package.json
mahib-stripe: Done!

mahib-stripe:
@olafurpg addressed your comments here

olafurpg (approved):
LGTM 👍🏻 I'll let @tgodzik merge since I have not maintained this repo for many years :)

src/extension.ts
```diff
@@ -405,12 +406,33 @@ function debugInformation(
...
 async function fetchAndLaunchMetals(
   context: ExtensionContext,
   serverVersion: string,
   javaVersion: JavaVersion,
   forceCoursierJar = false,
 ) {
+  const configuredServerOptions = getConfiguredServerOptions();
+  if (configuredServerOptions) {
+    outputChannel.appendLine("Using custom Metals server launcher.");
+    enableScaladocIndentation();
```
tgodzik: Is this needed here? We already have it in https://github.com/scalameta/metals-vscode/blob/bb0d98a035f13c9f84ec09bb5231d00cf3b668ef/src/extension.ts#L571

tgodzik: Or we could move it later into launchMetalsWithServerOptions, or earlier.

tgodzik (changes requested):
One minor comments, otherwise LGTM from me

src/extension.ts
mahib-stripe: I ended up moving it into `launchMetalsWithServerOptions()` so it's only called once, thanks for flagging!

tgodzik (approved):
LGTM!

PR #1998, opened 2026-06-09
Add proto document selector for protobuf LSP

mahib-stripe:
Committed-By-Agent: codex

### Summary
Per the tin, lets users configure the proto document selector for Metals LSP when the `protobufLsp` flag is set to true. I was able to connect [`peterj.protobuf`](https://github.com/peterj/vscode-protobuf/blob/main/package.json) to Metals and proceeded to jump-to-def between proto to proto, instead of just Java to proto.

(Recent changes): prototext support follows the same setting as protobuf here.

We use the peterj extension at Stripe which already contributes .proto as language id proto. 
We may consider switching to DrBlury.protobuf-vsc, though longer term I think we would need to evaluate the scalameta protobuf extension before shifting over all protobuf LS usages.

More context (thread with @olafurpg): https://bazelbuild.slack.com/archives/C0A8ZMZSHL0/p1781018942853479?thread_ts=1781013257.001399&cid=C0A8ZMZSHL0

Requesting a review @tgodzik (for some reason the UI on the right side doesn't let me add reviewers 😅)

## Summary by CodeRabbit

* **New Features**
  * Adds a new setting "metals.protobufLsp" (window, default: true) to enable Metals language server support for Protocol Buffers.
  * Prototext support is controlled by the same setting.
  * Enabling this requires a separate extension that contributes the proto language and a window reload.

* **Tests**
  * Updated tests to validate document selector behavior across protobuf/prototext flag combinations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -369,7 +369,7 @@
         },
         "metals.protobufLsp": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "scope": "window",
           "markdownDescription": "Enable Metals language server support for Protocol Buffers files. Requires another extension to contribute the `proto` language and reloading the window."
         },
--- a/package.json
+++ b/package.json
@@ -367,6 +367,12 @@
           "scope": "window",
           "markdownDescription": "Enable best effort mode for Metals in Scala 3. If enabled, Metals will try to provide most up to date information from the codebase even if it's not compiling."
         },
+        "metals.protobufLsp": {
+          "type": "boolean",
+          "default": false,
+          "scope": "window",
+          "markdownDescription": "Enable Metals language server support for Protocol Buffers files. Requires another extension to contribute the `proto` language and reloading the window."
+        },
         "metals.defaultShell": {
           "type": "string",
           "scope": "machine-overridable",
--- a/src/documentSelector.ts
+++ b/src/documentSelector.ts
@@ -1,0 +1,38 @@
+import { LanguageClientOptions } from "vscode-languageclient/node";
+
+export interface LanguageSupport {
+  protobuf: boolean;
+  prototext: boolean;
+}
+
+export function buildDocumentSelector({
+  protobuf,
+  prototext,
+}: LanguageSupport): LanguageClientOptions["documentSelector"] {
+  const documentSelector: LanguageClientOptions["documentSelector"] = [
+    { scheme: "file", language: "scala" },
+    { scheme: "file", language: "java" },
+    { scheme: "file", language: "twirl-html" },
+    { scheme: "file", language: "twirl-xml" },
+    { scheme: "file", language: "twirl-js" },
+    { scheme: "file", language: "twirl-txt" },
+    { scheme: "jar", language: "scala" },
+    { scheme: "jar", language: "java" },
+  ];
+
+  if (protobuf) {
+    documentSelector.push(
+      { scheme: "file", language: "proto" },
+      { scheme: "jar", language: "proto" },
+    );
+  }
+
+  if (prototext) {
+    documentSelector.push(
+      { scheme: "file", language: "prototext" },
+      { scheme: "jar", language: "prototext" },
+    );
+  }
+
+  return documentSelector;
+}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -678,7 +678,7 @@ async function launchMetalsWithServerOptions(
 
   const clientOptions: LanguageClientOptions = {
     documentSelector: buildDocumentSelector({
-      protobuf: config.get<boolean>("protobufLsp") ?? false,
+      protobuf: config.get<boolean>("protobufLsp") ?? true,
       prototext: false,
     }),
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -367,6 +368,12 @@
           "scope": "window",
           "markdownDescription": "Enable best effort mode for Metals in Scala 3. If enabled, Metals will try to provide most up to date information from the codebase even if it's not compiling."
         },
+        "metals.protobufLsp": {
+          "type": "boolean",
+          "default": false,
```
tgodzik:
```suggestion
          "default": true,
```
should we default to true? We already activate on protobuf.

src/documentSelector.ts
```diff
@@ -0,0 +1,25 @@
+import { LanguageClientOptions } from "vscode-languageclient/node";
+
+export function buildDocumentSelector(
```
olafurpg: minor: we can make this an object instead of boolean argument. In our internal extension, we recently merged this diff here to support prototext as well

```diff
-function buildDocumentSelectors(protobufEnabled: boolean): DocumentSelector {
-  const baseSelectors: DocumentSelector = [
+interface LanguageSupport {
+  protobuf: boolean;
+  prototext: boolean;
+}
+
+function buildDocumentSelectors(support: LanguageSupport): DocumentSelector {
+  const selectors: DocumentSelector = [
     { scheme: "file", language: "scala" },
     { scheme: "file", language: "java" },
     { scheme: "jar", language: "scala" },
     { scheme: "jar", language: "java" }
   ];
 
-  if (protobufEnabled) {
-    return [
-      ...baseSelectors,
+  if (support.protobuf) {
+    selectors.push(
       { scheme: "file", language: "proto" },
       { scheme: "jar", language: "proto" }
-    ];
+    );
   }
-
-  return baseSelectors;
+  if (support.prototext) {
+    selectors.push(
+      { scheme: "file", language: "prototext" },
+      { scheme: "jar", language: "prototext" }
+    );
+  }
+  return selectors;
 }
```

package.json
```diff
@@ -47,6 +47,7 @@
   ],
   "activationEvents": [
     "onDebugResolve:scala",
+    "onLanguage:proto",
```
olafurpg: Just so it's clear, this will auto-configure all LSP requests to be routed to Metals and may cause subtle regressions for users who rely on other protobuf extensions. I think VS Code aggregates requests from all clients so the latency will be determined by the slowest LSP, and you may get duplicate suggestions.

What I remember mostly is that the external protobuf extension we had enabled mostly timed out in our large monorepo so we had to uninstall it for everyone to roll our the Metals Protobuf support.

olafurpg: Internally, we've had this setting enabled since ~February, but it was feature flagged. You can achieve the same effect by having the document selector listen to a config property instead of the activation event.

olafurpg (commented):
I'll leave it to @tgodzik to make the call. I think it's the right decision to make Protobuf a first-class supported language, it's a really nice feature when you work in a large Bazel codebase.

package.json
mahib-stripe: I would probably suggest "no" since we can easily set `protobufLsp = true` and not everyone is going to be using protobuf anyways (i.e. maybe they use Thrift, or nothing at all). Though if it doesn't hurt those non-users, we can set it to true i suppose.

src/documentSelector.ts
mahib-stripe: Done!

package.json
mahib-stripe: ack - removed this

package.json
tgodzik: I think it might hard to discover for a lot of users, and even if you're not using protobuf this will not cause ny problem to you. Let's set it to true by default.

src/extension.ts
```diff
@@ -676,16 +677,10 @@ async function launchMetalsWithServerOptions(
...
-      { scheme: "file", language: "twirl-html" },
-      { scheme: "file", language: "twirl-xml" },
-      { scheme: "file", language: "twirl-js" },
-      { scheme: "file", language: "twirl-txt" },
-      { scheme: "jar", language: "scala" },
-      { scheme: "jar", language: "java" },
-    ],
+    documentSelector: buildDocumentSelector({
+      protobuf: config.get<boolean>("protobufLsp") ?? false,
+      prototext: false,
```
tgodzik: So this is never turned on then? Should we instead have it use the same setting from above?

tgodzik (changes requested):

package.json
mahib-stripe: Set to true

src/extension.ts
mahib-stripe: yeah good call out - `prototext` should be set in line with how `protobuf` is set

mahib-stripe: Fixed

tgodzik (approved):
LGTM!

PR #2007, opened 2026-06-12
Activate Metals for proto files

rmm-stripe:
## Summary

Adds proto and prototext language activation events so Metals can start when opening protobuf files directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed VS Code extension activation to properly trigger when opening Proto and ProtoText files.


tgodzik (approved):
LGTM!

PR #2017, opened 2026-06-18
Passthrough noDebug into the metals server

jchyb:
VS code gives us a noDebug parameter (set for TestExplorer's "run test" commands), but we lose it at some point instead of passing to the server. We need to pass them for the linked PR to work. This is a safe change - the parameter should be ignored by the server if that version does not support it.

## Release Notes

* **New Features**
  * Added support for running Scala test suites in “no debug” mode.
  * When enabled via run/debug actions, the intent is now passed through to the debug adapter so tests can execute without attaching a debugger while preserving the existing debug workflow.
* **Chores**
  * Updated request payload shapes to include an optional `noDebug` flag across the relevant Scala run/test debug requests.

### DIFF
--- a/src/debugger/scalaDebugger.ts
+++ b/src/debugger/scalaDebugger.ts
@@ -197,11 +197,21 @@ async function debug(
 ): Promise<boolean> {
   await commands.executeCommand("workbench.action.files.save");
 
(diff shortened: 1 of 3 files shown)


odisseus (approved):

tgodzik (approved):
LGTM

Issue #1737, opened 2025-08-11
New Quoting mechanism cause run code lens not working in Windows system


dragove:
**Describe the bug**

In windows, the `run` code lens not working.

**To Reproduce** Steps to reproduce the behavior:

create a simple main method in `main.scala`

```scala
@main def main(): Unit = println("hi")
```

and click the `run` interactive code lens

**Expected behavior**

Code runs successfully

**Screenshots**

<img width="1292" height="557" alt="Image" src="https://github.com/user-attachments/assets/9062a500-09ca-4912-9f7e-8e3e625d94ce" />

**Installation:**

- Operating system: Windows
- VSCode version: 1.0.103
- VSCode extension version: 1.54.0
- Metals version: 1.6.2

**Additional context**

Task Info:
Scala run

Process ID (PID): 5508

Command line: cmd.exe '/c """C:\Users\jingchun\scoop\apps\temurin-lts-jdk\21.0.8-9.0.LTS\bin\java.exe" "-classpath" "C:\Users\jingchun\Workspace\Scala\demo.metals.tmp\classpath_1E33B161E21A14179AFA92DF926FB776.jar" "Main"""'

Shell integration: Basic

**Search terms**

cmd, windows, powershell

tgodzik:
Thanks for reporting! We only change the main class to be surrounded by ", does it work for you without those added quotes?

dragove:
Following are some tests. I hope these can help.

```powershell
PS C:\Users\jingchun\Workspace\Scala\demo> cmd.exe '/c """C:\Users\jingchun\scoop\apps\temurin-lts-jdk\21.0.8-9.0.LTS\bin\java.exe" "-classpath" "C:\Users\jingchun\Workspace\Scala\demo.metals.tmp\classpath_1E33B161E21A14179AFA92DF926FB776.jar" "Main"""'
文件名、目录名或者卷标语法不正确

PS C:\Users\jingchun\Workspace\Scala\demo> cmd.exe /c """C:\Users\jingchun\scoop\apps\temurin-lts-jdk\21.0.8-9.0.LTS\bin\java.exe" "-classpath" "C:\Users\jingchun\Workspace\Scala\demo\.metals\.tmp\classpath_1E33B161E21A14179AFA92DF926FB776.jar" "Main"""
Hello World!!!

[...]

PR #2019, opened 2026-06-18
Fixes #2018: Force cmd.exe /c for Windows run task using classpath

### LINKED ISSUE #2018: Regression in Windows Run since v1.55.0
**Describe the bug**

On Windows, running a main class via **Run** (Ctrl+F5 / "Run without debugging" / the `run` code lens — anything `noDebug`) hangs forever. The program actually executes (its output appears), but the run never completes: VS Code shows the session spinning indefinitely and the task is never marked finished.

I traced this to the extension running the main as a VS Code **Task** with a `ShellExecution` (not a debug session). When the Metals server returns an `ExtendedScalaMainClass` containing `classpath` + `javaBinary` (which it does for scala-cli / any build target with a JVM run environment), `escapedShellCommand` takes the `classpath && javaBinary` branch, and that branch **no longer applies `platformSpecificOptions()`** — so the task runs in the user's *default terminal profile* instead of a forced `cmd.exe /c`.

My default Windows profile is `cmd.exe /k …`. Because of `/k`, the shell stays open after `java` exits, so the task never terminates and the run hangs. The fallback `shellCommand` branch still forces `cmd.exe /c` and works fine.

**Regression / root cause**

Introduced by commit [`4b062cd3` — *"bugfix: Running via cmd might no longer be needed"*](https://github.com/scalameta/metals-vscode/commit/4b062cd3) (2025-08-11), first released in **v1.55.0** (last good release: **v1.54.0**).

Before that commit, `runMain` built `shellOptions = { ...platformSpecificOptions(), env }` and passed it to *both* branches of `escapedShellCommand`. The commit moved `shellOptions` construction inside `escapedShellCommand` and dropped `platformSpecificOptions()` from the `classpath && javaBinary` branch:

```ts
if (classpath && javaBinary) {
  const shellOptions = { env };                               // ← no platformSpecificOptions(): runs in the user's default shell
  return taskFromArgs(classpath, javaBinary, jvmOptions, args, mainClass, shellOptions);
} else {
  const shellOptions = { ...platformSpecificOptions(), env }; // fallback still forces cmd.exe /c
  return new ShellExecution(main.data.shellCommand, shellOptions);
}
```

The original `platformSpecificOptions()` comment already documented *why* the forced cmd is needed on Windows ("tasks seem to not work properly… that's why we want to explicitly specify executable as plain old cmd"). That reasoning still holds for the `classpath` branch.

1. On Windows, with the default integrated terminal profile being Command Prompt (`cmd.exe`, especially one that keeps the shell open, e.g. `/k`).
2. Open a minimal scala-cli single-file project:
   ```scala
   //> using scala 3.8.4
   @main def main = println("hello")
   ```
3. Run the `main` via Ctrl+F5 / "Run without debugging" / the `run` code lens (any `noDebug` run).
4. `hello` prints in the terminal, but the run never finishes — it hangs / spins forever.

Notes:
- **F5 (Debug)** works fine — it uses the DAP `startDebugging` path, not the task path.
- Running the same command directly (`scala-cli run .`, or `java -cp <classpath-jar> main`) works and exits 0. The BSP `debugSession/start` + full DAP launch (driven manually) also completes cleanly. Only the `noDebug` → `ShellExecution` task path hangs.

**Expected behavior**

A `noDebug` run should complete (and the task should terminate) regardless of the user's default terminal profile, exactly as it did in ≤ v1.54.0.

**Screenshots**

N/A

**Installation:**

- Operating system: Windows 10 Pro (10.0.19045)
- VSCode version: 1.124.2
- VSCode extension version: 1.67.0
- Metals version: 1.6.7

**Additional context**

- Suggested fix: restore `platformSpecificOptions()` on the `classpath && javaBinary` branch (i.e. `const shellOptions = { ...platformSpecificOptions(), env }` for both branches), or otherwise force a self-terminating shell for the run task on Windows.
- Workarounds that confirm the diagnosis:
Update `settings.json`:
```
"Command Prompt": {
    "path": ["${env:windir}\\Sysnative\\cmd.exe", "${env:windir}\\System32\\cmd.exe"],
    "env": { "PROMPT": "$g" },   // was: "args": ["/k", "prompt $g"]
    "icon": "terminal-cmd"
}
```

**Search terms**
Stuck

soronpo:
The `classpath && javaBinary` branch of escapedShellCommand dropped platformSpecificOptions(), so on Windows the no-debug run task fell back to the user's default terminal profile instead of `cmd.exe /c`. With a profile that keeps the shell open (e.g. `cmd.exe /k ...`), the task never terminates and the run hangs forever (the program runs but the session spins indefinitely).

Restore platformSpecificOptions() on this branch, matching the shellCommand fallback, so the run task is launched via a self-terminating `cmd /c` on Windows.

Regression introduced in 4b062cd3 ("Running via cmd might no longer be needed").

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistent behavior in debugger task execution configuration to ensure consistent operation across different code paths.

### DIFF
--- a/src/debugger/scalaDebugger.ts
+++ b/src/debugger/scalaDebugger.ts
@@ -100,7 +100,7 @@ function escapedShellCommand(
   const mainClass = main.data.class;
   const args = main.data.arguments;
   if (classpath && javaBinary) {
-    const shellOptions = { env };
+    const shellOptions = { ...platformSpecificOptions(), env };
     return taskFromArgs(
       classpath,
       javaBinary,


tgodzik:
The change was done in order to possibly fix the problem the user had in the issue https://github.com/scalameta/metals-vscode/issues/1737

I don't remember however why I did that change 🤔 

Would be great to see if we can't reproduce the original issue, but it might be hard to do.

PR #2073, opened 2026-08-04
improvement: Read zip inside the plugin

tgodzik:
Instead of reading the files in Metals and sending heavy data using LSP it's much easier to just read the jar in VS Code itself. Virtual docs are anyway only supported in VS Code, so any other editors will just get the normal filesystem breadcrumbs.

This has an added benefit of being able to quickly switch between siblings.

https://github.com/user-attachments/assets/5940345a-cb16-4501-9643-9fc002419d1b

TODO: Metals focus command needs to be fixed.

## Summary by CodeRabbit

* **New Features**
  * Added experimental, read-only browsing for files and directories inside JAR archives.
  * Scala and Java language features now work with documents opened from JAR files.
  * Protobuf and prototext support includes JAR-based documents when enabled.
  * Navigation and code intelligence correctly resolve locations within JAR contents.
  * Added reliable handling for missing or invalid JAR resources.
  * Added a setting to enable experimental JAR filesystem support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: coursier/setup-action@v1
       - uses: actions/setup-node@v2
         with:
           node-version: "22"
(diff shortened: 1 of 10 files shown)


tgodzik:
I've put this feature behind `experimentalJarFileSystemEnabled` to be on the safe side, but I think it works pretty nice

